PR #15, opened 2025-08-08
feat(chat): allowing to select model

axel7083:
Mixted with https://github.com/kortex-hub/kortex/issues/13

https://github.com/user-attachments/assets/8de15949-25a4-4c30-84be-5e3e4f178d37


extensions/gemini/src/gemini.ts
```diff
@@ -138,6 +138,13 @@ export class Gemini implements Disposable {
       lifecycle: {
         delete: clean.bind(this),
       },
+      models: [{
```
axel7083, line 141: hardcoding because too lazy for now to fetch models

packages/renderer/src/lib/chat/hooks/selected-model.svelte.ts
```diff
@@ -1,5 +1,8 @@
 import { SynchronizedCookie } from '/@/lib/chat/utils/reactivity.svelte';
 
+/**
+ * @deprecated
+ */
 export class SelectedModel extends SynchronizedCookie {
```
benoitf, line 6: (remark) it can be removed it was an import from the template

fbricon: SelectedModel can be removed? It's marked as deprecated but it's unclear what would it be replaced with

PR #19, opened 2025-08-11
feat: support mcp selection for chat

axel7083:
Fixes https://github.com/kortex-hub/kortex/issues/16

https://github.com/user-attachments/assets/046e57c3-205c-4d15-9f7f-9b3c9da25a8f


packages/renderer/src/lib/chat/components/ipc-chat-transport.ts
```diff
@@ -1,8 +1,14 @@
...
 import type { ModelInfo } from '/@/lib/chat/components/model-info';
 
+interface Dependencies {
+  getModel: () => ModelInfo,
+  getMCP: () => Array<string>
+}
+
 export class IPCChatTransport<T extends UIMessage> implements ChatTransport<T> {
-  constructor(private readonly getModel: () => ModelInfo) {
+  constructor(private readonly dependencies: Dependencies) {
```
benoitf, line 11: looks a bit too generic name

PR #76, opened 2025-08-21
fix: add persistence to the openai inference provider

### LINKED ISSUE #34: openai provider is not persistent over reboot
Since I need to save both baseURL + token, I could not copy paste the code from the gemini & github mcp, so the openai provider has no persistent over reboot

jeffmaury:
Fixes #34

### DIFF
--- a/extensions/openai-compatible/src/openAI.ts
+++ b/extensions/openai-compatible/src/openAI.ts
@@ -18,13 +18,31 @@
 import { createHash } from 'node:crypto';
 
 import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
-import type { Disposable, Provider, provider as ProviderAPI, ProviderConnectionStatus } from '@kortex-app/api';
+import type {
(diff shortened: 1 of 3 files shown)


extensions/openai-compatible/src/openAI.ts
```diff
@@ -36,13 +54,71 @@ export class OpenAI implements Disposable {
...
+    let raw: string | undefined;
+    try {
+      raw = await this.secrets.get(TOKENS_KEY);
+    } catch (err: unknown) {
+      console.error('openai: something went wrong while trying to get tokens from secret storage', err);
+    }
+    // if undefined return empty array
+    if (!raw) return [];
+    // split raw string by token separator
+    return raw.split(TOKEN_SEPARATOR).map(str => {
```
axel7083, line 84: I used the `separator` technique in other extensions, but for long term we should probably just `JSON.parse / JSON.stringify`

Issue #68, opened 2025-08-21
Add types to MCP registry definition


simonrey1:
MCP registry located at https://github.com/benoitf/mcp-registry-online/blob/gh-pages/v0/servers/index.html

does not have type validation.

We could use `import type { MCPServerConfig } from "@mastra/core/mcp";`

PR #77, opened 2025-08-25
fix: use external lib to get mcp types

simonrey1:
I also had to change some code because git would not let me commit with the existing errors

fixes https://github.com/kortex-hub/kortex/issues/68


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -114,16 +115,17 @@ export class MCPRegistry {
 
       const { servers } = await this.listMCPServersFromRegistry(registry.serverUrl);
       for (const server of servers) {
+        if (!server.id) {
+          console.warn(`[MCPRegistry] no id for server`);
```
axel7083:
```suggestion
          console.warn(`[MCPRegistry] no id for server`, server);
```

might be nice to log the object

PR #81, opened 2025-08-27
feat(mcp): adding details page

axel7083:
Fixes https://github.com/kortex-hub/kortex/issues/80

image

### DIFF
--- a/packages/preload/src/index.ts
+++ b/packages/preload/src/index.ts
@@ -1120,6 +1120,30 @@ export function initExposure(): void {
     },
   );
 
+  contextBridge.exposeInMainWorld(
+    'createMCPProviderConnection',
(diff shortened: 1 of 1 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -2155,6 +2155,22 @@ export class PluginSystem {
       return mcpManager.listMCPRemoteServers();
     });
 
+    this.ipcHandle(
+      'mcp-manager:getTools',
+      async (_listener, mcpId: string): Promise<Record<string, { description: string }>> => {
+        const tools = await mcpManager.getToolSet([mcpId]);
```
axel7083, line 2161: tools is not serialisable

PR #83, opened 2025-08-27
feat: adding mcp-registry package

axel7083:
Following https://github.com/kortex-hub/kortex/pull/77 we depends on an external lib, but this lib is not up to date.

The types are changing almost everyday, and depending on a third party can be problematic. 

They provide an `openapi` spec on their repository https://github.com/modelcontextprotocol/registry/blob/main/docs/server-registry-api/openapi.yaml that seems to be up to date. We can use the [openapi-ts.dev](https://openapi-ts.dev/cli) to generate the types

---

Open to discussion honestly, we could have a cronjob on github running every day to updates the types, but since we don't have GitHub actions this may be tricky :/


packages/api/src/mcp/mcp-registry-server-entry.ts
```diff
@@ -19,9 +19,9 @@
 // definition of a MCP server as listed in a MCP registry
 // it should come frome types https://github.com/modelcontextprotocol/registry/issues/186
 
-import type { MCPServerConfig } from '@mastra/core/mcp';
+import type { components } from 'mcp-registry';
 
 export interface MCPRegistryServerList {
-  servers: MCPServerConfig[];
+  servers: components['schemas']['ServerDetail'][];
```
simonrey1, line 25: suggestion: export a MCPServerConfig from this file that contains `components['schemas']['ServerDetail']` to simplify other files

packages/renderer/src/lib/mcp/MCPRegistryCreateFromRegistryModal.svelte
```diff
@@ -1,17 +1,15 @@
...
 
 import Dialog from '/@/lib/dialogs/Dialog.svelte';
 
 import PasswordInput from '../ui/PasswordInput.svelte';
 
 interface Props {
-  mcpRegistryServerDetail: MCPServerConfig & {
-    remotes?: (RemoteInfo & { headers?: { name: string; is_secret: boolean; description: string }[] })[];
-  };
+  mcpRegistryServerDetail: components['schemas']['ServerDetail'];
```
simonrey1, line 12: Cool :) I had to improve the type of mastra lib here, and you have everything in the official openAPI

PR #90, opened 2025-08-28
fix: handle failures on openai extension start

jeffmaury:


extensions/openai-compatible/src/openAI.ts
```diff
@@ -62,7 +62,14 @@ export class OpenAI implements Disposable {
...
     const connectionInfos = await this.getConnectionInfos();
     for (const connectionInfo of connectionInfos) {
-      await this.registerInferenceProviderConnection({ token: connectionInfo.apiKey, baseURL: connectionInfo.baseURL });
+      try {
+        await this.registerInferenceProviderConnection({
+          token: connectionInfo.apiKey,
+          baseURL: connectionInfo.baseURL,
+        });
+      } catch (err: unknown) {
+        console.error(`openai: failed to restore connection for baseURL ${connectionInfo.baseURL}`);
```
benoitf: should we log the error ? (err)

jeffmaury: Fixed

Issue #85, opened 2025-08-27
Export an existing chat session as a flow


jeffmaury:
Once I am satisfied with a chat session, I want to export it as a flow. As flows are implemented by various runtimes, I would select the runtime (even if in the initial implementation, only a single runtime will be available)

PR #94, opened 2025-08-28
feat: export flow button

simonrey1:
This PR adds a button to the chat UI that redirects users to the flow creation page.

It passes a data object containing the model, MCP, and last user prompt. For now, only the prompt field is pre-populated, but the other information is available in the object for future use.

This is the first step toward a fully functional end-to-end flow creation.

Partial fix of https://github.com/kortex-hub/kortex/issues/85

### DIFF
--- a/packages/renderer/src/lib/flows/flowCreationStore.ts
+++ b/packages/renderer/src/lib/flows/flowCreationStore.ts
@@ -3,7 +3,7 @@ import { writable } from 'svelte/store';
 import type { ModelInfo } from '/@/lib/chat/components/model-info';
 
 export interface FlowCreationData {
-  lastUserMessage: string;
+  prompt: string;
   model: ModelInfo;
   mcp: Set<string>;
 }
--- a/packages/renderer/src/lib/chat/components/chat.svelte
+++ b/packages/renderer/src/lib/chat/components/chat.svelte
@@ -117,10 +117,12 @@ let attachments = $state<Attachment[]>([]);
 			{readonly}
 			loading={chatClient.status === 'streaming' || chatClient.status === 'submitted'}
 			messages={chatClient.messages}
+      {selectedModel}
+      {selectedMCP}
 		/>
      <form class="bg-background mx-auto flex w-full gap-2 px-4 pb-4 md:max-w-3xl md:pb-6">
        {#if !readonly}
-         <MultimodalInput {attachments} {user} {chatClient} selectedModel={selectedModel} selectedMCP={selectedMCP} class="flex-1" />
+         <MultimodalInput {attachments} {user} {chatClient} class="flex-1" />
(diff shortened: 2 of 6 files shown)


packages/renderer/src/lib/flows/flowCreationStore.ts
```diff
@@ -0,0 +1,16 @@
...
+  model: ModelInfo;
+  mcp: Set<string>;
+}
+
+/**
+ * A writable store to temporarily hold the data for a new flow
+ * when navigating from a chat session to the creation page.
+ * It's set to `null` after being read to prevent stale data.
+ */
+export const flowCreationStore = writable<FlowCreationData | null>(null);
```
benoitf: Hello, I don't see why we need the store ?
also null is forbidden

simonrey1: I am using a store instead of query params to pass objects easily (no dangerous JSON parse) and ensure type safety.

Ok I will change to undefined no problem

benoitf: I don't see any reason of 'dangerous JSON' or type safety ?
if the value are incorrect then you're just displaying a page saying that data are invalid

benoitf: message is a string, MCP you needs the ids and for the model its id

so all them are string

benoitf: > I am using a store instead of query params to pass objects easily (no dangerous JSON parse) and ensure type safety.
> 
> Ok I will change to undefined no problem

the main problem is that it won't follow any of the existing patterns currently being used

simonrey1: Using a Svelte store for the creation form is a more extensible solution than using query parameters. It allows us to pass the entire object without the extra steps of serialization and parsing.

With query parameters, we’d have to not only JSON.parse() the data but also validate that the result, like our MCP list, is actually an array. The store avoids this added complexity.

Regarding the existing patterns, I believe this is a better, more maintainable pattern for handling complex data in our application.

benoitf: > With query parameters, we’d have to not only JSON.parse() the data but also validate that the result, like our MCP list, is actually an array. The store avoids this added complexity.

There is no JSON.parse, it uses `meta.params`

simonrey1: > There is no JSON.parse, it uses `meta.params`

So meta.params accepts a Record of string.

But we can select multiple MCP.

So we have to do JSON.parse (or .split(',')) on the `meta.params.selectedMcps`

Screenshot 2025-08-28 at 11 36 03

Issue #86, opened 2025-08-27
Allow local execution of a flow


jeffmaury:
Once a flow has been created from a chat session, I want to execute it. As flows are implemented by various runtime, the proper runtime should be selected

PR #113, opened 2025-08-29
feat: allow to execute locally goose recipes

benoitf:
https://github.com/user-attachments/assets/82f4f9ea-0ac8-4b47-9170-d7ed2363e97a

fixes https://github.com/kortex-hub/kortex/issues/86

### DIFF
--- a/packages/main/src/plugin/flow/flow-manager.ts
+++ b/packages/main/src/plugin/flow/flow-manager.ts
@@ -15,24 +15,56 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import { Disposable, Flow, FlowProviderConnection } from '@kortex-app/api';
+import type { Disposable, Flow, FlowProviderConnection, Logger } from '@kortex-app/api';
 import { inject, injectable, preDestroy } from 'inversify';
 
 import { ApiSenderType } from '/@/plugin/api.js';
 import { ProviderRegistry } from '/@/plugin/provider-registry.js';
-import { FlowInfo } from '/@api/flow-info.js';
+import type { FlowExecuteInfo } from '/@api/flow-execute-info.js';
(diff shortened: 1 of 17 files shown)


packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -15,24 +15,56 @@
...
 
 @injectable()
 export class FlowManager implements Disposable {
   #installedProviders: Set<string> = new Set();
   #flows: Map<string, Array<Flow>> = new Map();
+  // key = execution id = task id
+  #flowsExecution: Map<string, FlowExecuteInfo> = new Map();
+  #flowsLogs: Map<string, BufferLogger> = new Map();
   #disposable: Map<string, Disposable> = new Map();
+  #currentLogger: BufferLogger | undefined = undefined;
```
axel7083, line 59: What happens if I try to start two Flow?

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -15,24 +15,56 @@
...
+  }
+}
 
 @injectable()
 export class FlowManager implements Disposable {
   #installedProviders: Set<string> = new Set();
   #flows: Map<string, Array<Flow>> = new Map();
+  // key = execution id = task id
+  #flowsExecution: Map<string, FlowExecuteInfo> = new Map();
+  #flowsLogs: Map<string, BufferLogger> = new Map();
```
axel7083, line 57: Can we dispose those maps in the dipose() method?

packages/main/src/plugin/flow/flow-manager.ts
benoitf: it's only displaying one logger content, but it'll still grab all logs of all flows

axel7083: I have some issue understanding the prupose of the `#currentLogger`

PR #135, opened 2025-08-29
fix: don't display flow runtime selector if there is a single one

### LINKED ISSUE #119: In flow creation, when only one flow provider, select automatically in combobox
In flow creation, when only one flow provider, select automatically in combobox

jeffmaury:
Fixes #119

### DIFF
--- a/packages/renderer/src/lib/flows/FlowCreate.svelte
+++ b/packages/renderer/src/lib/flows/FlowCreate.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 import { Button, ErrorMessage, Input } from '@podman-desktop/ui-svelte';
-import { onMount } from 'svelte';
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -28,6 +30,25 @@ let flowProviderConnectionKey: string | undefined = $state<string>();
 flowCreationStore.set(undefined);
 
 let hasInstalledFlowProviders = $state(window.hasInstalledFlowProviders());
+let showFlowConnectionSelector = $state(true);
+
+onMount(() => {
+  // If there is exactly one available flow connection across all providers,
```
benoitf: wondering why it's not using $derived as well

PR #137, opened 2025-08-29
feat(goose): cli install

axel7083:
Fixes https://github.com/kortex-hub/kortex/issues/148

https://github.com/user-attachments/assets/3a177714-cbc2-4bb6-8f97-44b5d4770e33

### DIFF
--- a/extensions/goose/src/goose-cli.ts
+++ b/extensions/goose/src/goose-cli.ts
@@ -76,29 +76,12 @@ export class GooseCLI implements Disposable {
     return !!this.cli?.version;
   }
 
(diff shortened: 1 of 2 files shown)


extensions/goose/src/goose-cli.ts
```diff
@@ -86,14 +105,16 @@ export class GooseCLI implements Disposable {
   }
 
   async getRecipes(options: { path: string }): Promise<Array<{ path: string }>> {
+    if (!this.cli?.path) throw new Error('goose not installed');
```
simonrey1, line 108: It seems already checked on line 110

PR #152, opened 2025-09-01
feat: add support for macOS for goose cli

benoitf:

fixes https://github.com/kortex-hub/kortex/issues/150


extensions/goose/src/goose-downloader.ts
```diff
@@ -51,12 +52,28 @@ export class GooseDownloader implements Disposable {
 
   dispose(): void {}
 
+  extractTarBz2(filePath: string, outDir: string): Promise<void> {
```
simonrey1, line 66: We already have "tar" installed in the npm packages, it could be less hacky than a command line

PR #157, opened 2025-09-01
feat(flow/generate): propagate MCP credentials

axel7083:
https://github.com/user-attachments/assets/24c2a80f-19f3-4860-8b69-1a4e81052bac


packages/renderer/src/lib/chat/components/suggested-actions.svelte
```diff
@@ -19,7 +19,7 @@ const suggestedActions = [
     title: 'What are the last 5 issues of Github',
     label: 'repository podman-desktop/podman-desktop?',
     action: 'What are the last 5 issues of Github repository podman-desktop/podman-desktop?',
-    requiredMcp: ['internal:GitHub MCP server'],
+    requiredMcp: ['internal:123e4567-e89b-12d3-a456-426614172000:0:GitHub MCP server'],
```
simonrey1, line 22: maybe delete the last part "GitHub MCP server" seems unused once we have the remote id

Issue #160, opened 2025-09-01
Change goose install instructions for a Link to the CLI config


simonrey1:
<img width="1162" height="812" alt="Image" src="https://github.com/user-attachments/assets/b3e21c95-6f61-4950-a1c0-fd96ae40cdfc" />

PR #167, opened 2025-09-01
fix: redirect to install cli component instead of instructions

simonrey1:
fixes https://github.com/kortex-hub/kortex/issues/160

### DIFF
--- a/packages/renderer/src/lib/flows/components/NoFlowProviders.svelte
+++ b/packages/renderer/src/lib/flows/components/NoFlowProviders.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 import { Button, Link } from '@podman-desktop/ui-svelte';
 import { toast } from 'svelte-sonner';
+import { router } from 'tinro';
 
 import { combinedInstalledExtensions } from '/@/stores/all-installed-extensions';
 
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/flows/components/NoFlowProviders.svelte
```diff
@@ -41,9 +42,10 @@ let extension = $derived($combinedInstalledExtensions.find(e => e.id === 'kortex
     <div class="max-w-2xl space-y-4">
       <p class="text-gray-500">To get started with flows, please follow these steps:</p>
       
-      <ol class="text-left text-gray-500 space-y-3 list-decimal list-inside">
-        <li>Install the Goose CLI by running:
-          <code class="block bg-gray-800 text-gray-200 p-2 mt-1 rounded font-mono text-sm">curl -fsSL https://github.com/block/goose/releases/download/stable/download_cli.sh | bash</code>
+      <ol class="text-center text-gray-500 space-y-3 list-inside">
+        <li>
+          <Link on:click={(): void => router.goto('/preferences/cli-tools')}
+              >Install the Goose CLI</Link>
```
benoitf, line 48: AFAIK we have handleNavigation method

benoitf: handleNavigation({ page: NavigationPage.CLI_TOOLS });

Issue #154, opened 2025-09-01
use a random/pre-defined name when generating a Flow from a chat


benoitf:
for now, the field is missing and the generate button disabled

idea: use a name like flow-1, flow-2 or like docker's container names :  flow-blue-spoon

PR #171, opened 2025-09-01
feat: add random default name to flow

simonrey1:
fixes https://github.com/kortex-hub/kortex/issues/154

Screenshot 2025-09-01 at 15 07 45

generate as generateWords because generate already exists in this file

also had to disable a eslint on this line, if someone could find a solution let me know


packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -1,5 +1,7 @@
 <script lang="ts">
 import { Button, ErrorMessage, Input } from '@podman-desktop/ui-svelte';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
benoitf, line 3: why this line ?

simonrey1: I don't know why eslint would not pass without it

Issue #161, opened 2025-09-01
Move Generate button to the bottom right of form


simonrey1:
<img width="1459" height="991" alt="Image" src="https://github.com/user-attachments/assets/84ad4295-207e-4e31-b552-3c0d4869fef4" />

PR #172, opened 2025-09-01
fix: move generate button to the right

simonrey1:
fixes https://github.com/kortex-hub/kortex/issues/161

Screenshot 2025-09-01 at 15 16 43


packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -156,10 +156,12 @@ async function generate(): Promise<void> {
...
-              {#if showFlowConnectionSelector}
-              <FlowConnectionSelector class="" bind:value={flowProviderConnectionKey}/>
-              {/if}
-              <Button inProgress={loading} disabled={!flowProviderConnectionKey || !name} onclick={generate}>Generate</Button>
+              <div class="flex w-full">
+                {#if showFlowConnectionSelector}
+                <FlowConnectionSelector class="" bind:value={flowProviderConnectionKey}/>
+                {/if}
+                <Button class="ml-auto" inProgress={loading} disabled={!flowProviderConnectionKey || !name} onclick={generate}>Generate</Button>
+              </div>
```
benoitf: IMHO the button should be on its own line

simonrey1: I tried this at first but I did not find the button so I thought it would be better to keep on the same line

PR #175, opened 2025-09-01
fix: add instructions to README

jeffmaury:
Relates to #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,4 +62,4 @@ Kortex uses the Goose as a flow provider. So there are two options here:
 
 ### End to end workflow demo
 
-![](media/onboarding-demo.gif)
+![](https://raw.githubusercontent.com/kortex-hub/kortex/media/media/onboarding-demo.gif)


README.md
```diff
@@ -1 +1,61 @@
...
+
+```sh
+pnpm watch
+```
+
+## Using Kortex
+
+### Configure your Gemini API Key
+
+Go to `Settings > Resources`, the Gemini provider should be list. Click on `Grab a key` if you don't have an existing
```
simonrey1: typo: be listed

media/onboarding-demo.gif
simonrey1, line 1: Great video! It shows a lot of features that have been implemented in this POC!


Warning: the GitHub personal access token appears in the source of the flow and it is not blurred in the video.

I was also expecting to see the container logs as you streamed your whole Desktop, maybe switching to Podman Desktop window.

For the video size, I think it is possible to upload to the README from GitHub interface (for example Drag and Drop the video while editing README.md in GitHub) and it would generate a link with something like "githubusercontent" that will be hosted by GitHub. So I agree that the size of the video is a bit too big for people to clone this repo easily.

PR #183, opened 2025-09-01
fix: pass model credentials to Goose

### LINKED ISSUE #178: Allow execution of a workflow if Goose has never been configured
For the Gemini provider, we can use env variable GOOGLE_API_KEY

jeffmaury:
Fixes #178

### DIFF
--- a/packages/main/src/plugin/provider-registry.ts
+++ b/packages/main/src/plugin/provider-registry.ts
@@ -1672,33 +1672,6 @@ export class ProviderRegistry {
     return connections;
   }
 
-  registerMCPConnection(provider: Provider, mcpProviderConnection: MCPProviderConnection): Disposable {
-    const providerName = mcpProviderConnection.name;
-    const id = `${provider.id}.${providerName}`;
-    this.mcpProviders.set(id, mcpProviderConnection);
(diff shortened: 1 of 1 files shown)


extensions/goose/src/goose-recipe.ts
```diff
@@ -93,17 +103,28 @@ export class GooseRecipe implements Disposable {
...
+    const content = await readFile(flowPath, 'utf-8');
+    const parsed = parse(content);
+    const gooseProvider = goose2KortexProvider(parsed['settings']['goose_provider']);
+    const gooseModel: string = parsed['settings']['goose_model'];
+    const connection = this.provider
+      .getInferenceConnections()
+      .find(c => c.providerId === gooseProvider && c.connection.models.some(m => m.label === gooseModel));
+    if (!connection) throw new Error(`cannot find connection for ${gooseProvider} ${gooseModel}`);
+    return {
+      GOOGLE_API_KEY: connection.connection.credentials()['gemini:tokens'] ?? 'API_TOKEN',
```
simonrey1, line 116: Maybe throw an error if token not found instead of returning "API_TOKEN" that will make an error on API call

PR #185, opened 2025-09-01
fix: avoid recreating mcp entries

benoitf:
fixes https://github.com/kortex-hub/kortex/issues/151

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-registry.ts
+++ b/packages/main/src/plugin/mcp/mcp-registry.ts
@@ -128,6 +128,14 @@ export class MCPRegistry {
           continue;
         }
 
+        // client already exists ?
+        const existingServers = await this.mcpManager.listMCPRemoteServers();
+        const existing = existingServers.find(srv => srv.id.includes(server.id ?? 'unknown'));
+        if (existing) {
+          console.log(`[MCPRegistry] MCP client for server ${server.id} already exists, skipping`);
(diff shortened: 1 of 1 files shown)


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -128,14 +128,29 @@ export class MCPRegistry {
           continue;
         }
 
+        // client already exists ?
+        const existingServers = await this.mcpManager.listMCPRemoteServers();
+        const existing = existingServers.find(srv => srv.id.includes(server.id ?? 'unknown'));
```
axel7083, line 133: A server details can have multiple remote id, this would prevent registering same serverId different remoteId

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -128,14 +128,29 @@ export class MCPRegistry {
           continue;
         }
 
+        // client already exists ?
+        const existingServers = await this.mcpManager.listMCPRemoteServers();
```
axel7083, line 132: Why this is inside the loop? We should get this info before the starting of the for

packages/main/src/plugin/mcp/mcp-registry.ts
benoitf: for now only the first remote is used

PR #194, opened 2025-09-02
fix(GooseRecipe): check provider is supported, refacto

simonrey1:


extensions/goose/src/goose-recipe.ts
```diff
@@ -186,8 +187,12 @@ export class GooseRecipe implements Disposable {
     });
   }
 
+  private checkSupportedProvider(providerId: string): void {
+    if (providerId !== 'gemini') throw new Error('unsupported provider');
```
axel7083: can we use a switch?

PR #195, opened 2025-09-02
feat(flow): support deletion

axel7083:
https://github.com/user-attachments/assets/a0be1518-0f0f-404d-b2d0-f4df8f10b692

### DIFF
--- a/packages/renderer/src/lib/flows/FlowActions.svelte
+++ b/packages/renderer/src/lib/flows/FlowActions.svelte
@@ -43,7 +43,6 @@ async function deleteFlow(): Promise<void> {
     loading = true;
     try {
       await window.deleteFlow(object.providerId, object.connectionName, object.id);
-      await window.refreshFlows();
     } catch (err: unknown) {
       console.error('Error deleting flow:', err);
     } finally {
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/flows/FlowActions.svelte
```diff
@@ -33,6 +37,22 @@ async function executeFlow(): Promise<void> {
   // redirect to the run tab
   navigateToCreateFlow();
 }
+
+async function deleteFlow(): Promise<void> {
+  withConfirmation(async () => {
+    loading = true;
+    try {
+      await window.deleteFlow(object.providerId, object.connectionName, object.id);
+      await window.refreshFlows();
```
benoitf: deleteFlow should trigger the update from server side (the store should be then updated and then flows are refreshed)

simonrey1: See https://github.com/kortex-hub/kortex/pull/203/files#diff-4f80e5bb5be559a31ef11bc94b0e515cba7335116fca43f80d9030a2f9ea3abfR224 for example

PR #197, opened 2025-09-02
refactor(GooseRecipe): put in common code to get gemini token

simonrey1:


extensions/goose/src/goose-recipe.ts
```diff
@@ -112,9 +113,8 @@ export class GooseRecipe implements Disposable {
       .getInferenceConnections()
       .find(c => c.providerId === kortexProvider && c.connection.models.some(m => m.label === gooseModel));
     if (!connection) throw new Error(`cannot find connection for ${kortexProvider} ${gooseModel}`);
-    this.checkSupportedProvider(kortexProvider);
     return {
-      GOOGLE_API_KEY: connection.connection.credentials()['gemini:tokens'] ?? 'API_TOKEN',
+      GOOGLE_API_KEY: this.getTokenForProvider({ providerId: kortexProvider, connection: connection.connection }),
```
axel7083: You cannot have a generic function for all provider that return `getTokenForProvider`, you need to have `getTokenForProvider` return a Record<string, string>

PR #204, opened 2025-09-02
fix(GooseRecipe): choose model from recipe for kubernetes deploy, refactor

simonrey1:
fix(GooseRecipe): choose model from recipe for kubernetes deploy, refactor #204

### DIFF
--- a/extensions/goose/src/goose-recipe.ts
+++ b/extensions/goose/src/goose-recipe.ts
@@ -15,7 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import { readFile, writeFile } from 'node:fs/promises';
+import { mkdir, readFile, unlink, writeFile } from 'node:fs/promises';
 import { homedir } from 'node:os';
 import { basename, dirname, join } from 'node:path';
 
@@ -94,6 +94,7 @@ export class GooseRecipe implements Disposable {
     const name = parsed['name'].toLowerCase();
 
     const basePath = this.getBasePath();
+    await mkdir(basePath, { recursive: true });
     const fullPath = join(basePath, `${name}.yaml`);
     await writeFile(fullPath, content);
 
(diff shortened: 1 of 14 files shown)


extensions/goose/src/goose-recipe.ts
```diff
@@ -186,32 +172,47 @@ export class GooseRecipe implements Disposable {
...
+  }> {
+    const flowPath = await this.getFlowPath(flowId);
+
+    const content = await readFile(flowPath, 'utf-8');
+
+    const recipeName = basename(flowPath).split('.')[0];
+
+    const parsed = parse(content);
+
+    const providerId = goose2KortexProvider(parsed['settings']['goose_provider']);
```
axel7083, line 197: a bit missleading to have `providerId` here, this is gooseProvider technically ?

extensions/goose/src/goose-recipe.ts
```diff
@@ -186,32 +172,47 @@ export class GooseRecipe implements Disposable {
...
+    const gooseModel: string = parsed['settings']['goose_model'];
+    const connection = this.provider
+      .getInferenceConnections()
+      .find(c => c.providerId === providerId && c.connection.models.some(m => m.label === gooseModel));
+
+    function getEnv(): Record<string, string> {
```
axel7083, line 203: suggestion(non-blocking): Could we have a little bit more explicit function name? Or a comment, because getEnv for what? some ideas `getGooseConfigEnvironment`, `getGooseEnv` etc.

extensions/goose/src/goose-recipe.ts
simonrey1: The method is goose2Kortex provider so it would be kortexProvider but not gooseProvider?

extensions/goose/src/goose-recipe.ts
simonrey1: https://github.com/kortex-hub/kortex/pull/260

PR #223, opened 2025-09-02
fix(flow/run): display flakiness

axel7083:
Fixes https://github.com/kortex-hub/kortex/issues/213


packages/renderer/src/lib/flows/FlowDetails.svelte
```diff
@@ -80,6 +88,10 @@ onMount(() => {
     })
     .catch(console.error);
 });
+
+function onLocalRun(flowExecuteId: string): void {
```
simonrey1: suggestion: rename to set... because it just a setter

Issue #218, opened 2025-09-02
For suggested toast to select MCP github


simonrey1:
- change enable to select
- open combobox for user

PR #225, opened 2025-09-02
fix(chat): suggested actions open combo improve message

simonrey1:
fixes https://github.com/kortex-hub/kortex/issues/218

### DIFF
--- a/packages/api/src/flow-execute-info.ts
+++ b/packages/api/src/flow-execute-info.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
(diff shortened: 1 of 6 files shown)


packages/api/src/flow-execute-info.ts
```diff
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { FlowInfo } from './flow-info';
+import type { FlowInfo } from './flow-info.ts';
```
benoitf, line 19: how this change is related to the fix ?

simonrey1: It is not related but there was an linter error

PR #240, opened 2025-09-02
fix: auto select first model in flow create

### LINKED ISSUE #220: No model selected by default when Flow > Create
Image

jeffmaury:
Fixed #220

### DIFF
--- a/packages/renderer/src/lib/chat/components/model-selector.svelte
+++ b/packages/renderer/src/lib/chat/components/model-selector.svelte
@@ -32,14 +32,18 @@ let groups: Map<string, Array<ModelInfo>> = $derived(
 );
 
 let open = $state(false);
+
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/chat/components/model-selector.svelte
```diff
@@ -32,14 +32,18 @@ let groups: Map<string, Array<ModelInfo>> = $derived(
 );
 
 let open = $state(false);
+
+let usedValue = $derived(value ?? models[0]);
```
axel7083, line 36: This is probably not needed since you select the default in chat.svelte

jeffmaury: I moved it here so that it is reused

PR #263, opened 2025-09-03
fix: default model handling is broken 

### LINKED ISSUE #254: In Flow > Create, default selected model does not enable Generate button
In Flow > Create, default selected model does not enable Generate button

You have to select the model manually for the Generate button to be enabled again

Image

jeffmaury:
Fixes #254 #255

### DIFF
--- a/packages/renderer/src/lib/models/models-utils.ts
+++ b/packages/renderer/src/lib/models/models-utils.ts
@@ -1,3 +1,4 @@
+import type { ModelInfo } from '/@/lib/chat/components/model-info';
 import type { ProviderInfo } from '/@api/provider-info';
 
 export interface ModelInfo {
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/models/models-utils.ts
```diff
@@ -25,3 +26,6 @@ export function getModels(providerInfos: ProviderInfo[]): ModelInfo[] {
     [] as Array<ModelInfo>,
   );
 }
+export function getFirstModel(models: Array<ModelInfo>): ModelInfo | undefined {
+  return models && models.length > 0 ? models[0] : undefined;
```
simonrey1, line 30: suggestion: It takes an array of models, so getFirstModel could be equivalent to `models[0]`

simonrey1: https://github.com/kortex-hub/kortex/pull/272/files

PR #283, opened 2025-09-03
feat(chat): show connect buttons for inference providers

simonrey1:
Help user by listing the inference providers with connect buttons on chat page when he has no inference provider connected

https://github.com/user-attachments/assets/9d786e32-90bb-4056-868e-4adb4dcd7e12


packages/renderer/src/lib/preferences/CreateProviderConnectionButton.svelte
```diff
@@ -0,0 +1,117 @@
...
+onMount(() => {
+  $providerInfos.forEach(provider => {
+    if (
+      providerToBeInstalled &&
+      doExecuteAfterInstallation &&
+      provider.name === providerToBeInstalled.provider.name &&
+      (provider.status === 'ready' || provider.status === 'installed')
+    ) {
+      providerToBeInstalled = undefined;
+      doExecuteAfterInstallation();
```
simonrey1, line 39: It seems that doExecuteAfterInstallation is not called anymore because it was in parent before

PR #300, opened 2025-09-04
fix(chat): reuse export icon in flow create, align

simonrey1:
Screenshot 2025-09-04 at 12 16 10
Screenshot 2025-09-04 at 12 15 41

### DIFF
--- a/packages/renderer/src/lib/chat/components/messages/ExportIcon.svelte
+++ b/packages/renderer/src/lib/chat/components/messages/ExportIcon.svelte
@@ -1,12 +1,12 @@
 <script lang="ts">
 import { faFileExport } from '@fortawesome/free-solid-svg-icons/faFileExport';
-import Fa from 'svelte-fa';
+import Fa, { type IconSize } from 'svelte-fa';
 
 let {
-  scale,
+  size,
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/chat/components/messages/ExportIcon.svelte
```diff
@@ -0,0 +1,12 @@
...
+import Fa from 'svelte-fa';
+
+let {
+  scale,
+}: {
+  scale: number;
+} = $props();
+</script>
+
+<Fa icon={faFileExport} scale={scale}/>
```
benoitf: we're using all the time size, so I'm not sure to use now scale

packages/renderer/src/lib/chat/components/messages/ExportIcon.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+import { faFileExport } from '@fortawesome/free-solid-svg-icons/faFileExport';
+import Fa from 'svelte-fa';
+
+let {
+  scale,
```
benoitf: size

PR #325, opened 2025-09-10
feat(chat): save messages in local SQLite db

### LINKED ISSUE #319: Implement persistence for chat
Chat sessions should be persisted so that user can manage their chat history

simonrey1:
feat(chat): save messages in local SQLite db

Fixes #319

Also I removed a lot of code to focus on the messages history, and I removed the other objects in db like User which are not relevant to our POC in a local electron app;

After this PR, need to read the messages from database to fill history.

Example of "Message" table content:
Screenshot 2025-09-10 at 17 33 13

https://github.com/user-attachments/assets/74d62c8c-e0fa-4193-b7a9-dabda3fb831e

### DIFF
--- a/packages/renderer/src/lib/chat/components/sidebar-history/item.svelte
+++ b/packages/renderer/src/lib/chat/components/sidebar-history/item.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 import { router } from 'tinro';
 
-import type { Chat } from '../../../../../../main/src/chat/db/schema';
+import type { DBChat } from '../../../../../../main/src/chat/db/schema';
 import MoreHorizontalIcon from '../icons/more-horizontal.svelte';
 import TrashIcon from '../icons/trash.svelte';
 import { DropdownMenu, DropdownMenuContent, DropdownMenuItem, DropdownMenuTrigger } from '../ui/dropdown-menu';
@@ -12,7 +12,7 @@ let {
   active,
   ondelete,
 }: {
-  chat: Chat;
+  chat: DBChat;
   active: boolean;
   ondelete: (chatId: string) => void;
 } = $props();


packages/main/src/chat/db/schema.ts
```diff
@@ -1,119 +1,25 @@
+import type { LanguageModelV2Usage } from '@ai-sdk/provider';
 import type { InferSelectModel } from 'drizzle-orm';
-import { sqliteTable, text, primaryKey, foreignKey, integer } from 'drizzle-orm/sqlite-core';
+import { json, jsonb, pgTable, text, timestamp, uuid, varchar } from 'drizzle-orm/pg-core';
```
axel7083: isn't weird to use postgres column type[^1] ?

[^1]: https://orm.drizzle.team/docs/column-types/pg

packages/main/src/chat/db/queries.ts
```diff
@@ -1,404 +1,112 @@
...
-  vote,
-} from './schema.js';
-import { unwrapSingleQueryResult } from './utils.js';
+const require = createRequire(import.meta.url);
+const { createClient } = require('@libsql/client/sqlite3');
 
-// Optionally, if not using email/pass login, you can
-// use the Drizzle adapter for Auth.js / NextAuth
-// https://authjs.dev/reference/adapter/drizzle
+const client = createClient({ url: 'file:local.db' });
```
axel7083: Maybe we should have this somewhere like the `~/.local/share/kortex/configuration`

packages/main/src/chat/db/queries.ts
```diff
@@ -1,404 +1,112 @@
...
-  return safeTry(async function* () {
-    const userResult = yield* fromPromise(
-      db.select().from(user).where(eq(user.email, email)),
-      e => new DbInternalError({ cause: e }),
-    );
-    return unwrapSingleQueryResult(userResult, email, 'User');
-  });
-}
+  const start = Date.now();
+  await migrate(db, { migrationsFolder: './packages/main/src/chat/db/migrations/' });
```
axel7083, line 27: Not sure those migrations would be included when packaging the application :/ need to be careful if we ever ship a binary

packages/main/src/chat/db/schema.ts
simonrey1: I copied and paste again the example from Vercel and got this change, it seems to be working but I will try to not use postgres types you are right

simonrey1: In commit https://github.com/kortex-hub/kortex/pull/325/commits/7d4781f931ecca3ea9025f5d1ff308ad4b8892dd#diff-4cb5c1317acabf9ec8f9a61634b7906b3cef5d52b202fd0310973e845704c6bd
I switched to sqlite core types, thanks for your advice!

packages/main/src/chat/db/queries.ts
simonrey1: Changed and tested here https://github.com/kortex-hub/kortex/pull/325/commits/3ea7b3ecfd7bd8bbc82273da2b3832af0c07ff4d

PR #332, opened 2025-09-12
fix(chat/flow): make prompt from whole conversation 

### LINKED ISSUE #268: discuss: handle multiple messages chat for recipes
The export feature is attached to a single message, we should support exporting the whole chat session that may be composed of several messages

simonrey1:
Instead of creating a flow from a given user prompt, use LLM to generate a prompt from the whole conversation, and store in the flow. So it can be executed automatically.

Demo:

https://github.com/user-attachments/assets/469a7923-766e-4752-8692-7abe87413726

Fixes #268

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2795,13 +2795,13 @@ export class PluginSystem {
         _listener: Electron.IpcMainInvokeEvent,
         providerId: string,
         connectionName: string,
-        modelId: string,
+        model: string,
         mcp: Array<string>,
         messages: UIMessage[],
       ): Promise<string> => {
         const internalProviderId = providerRegistry.getMatchingProviderInternalId(providerId);
         const sdk = providerRegistry.getInferenceSDK(internalProviderId, connectionName);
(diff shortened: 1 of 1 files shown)


packages/main/src/plugin/index.ts
```diff
@@ -2793,24 +2793,36 @@ export class PluginSystem {
...
+          throw new Error('No user message found');
+        }
+
+        //ai sdk/fetch does not support file:URLs
+        const convertedMessages = await convertMessages(messages);
+        const modelMessages = convertToModelMessages(convertedMessages);
+
+        const toolset = await mcpManager.getToolSet(mcp);
+
+        const res = await generateText({
```
benoitf: I don't see why result has been renamed to res as well
why toolSet is now toolset, etc

simonrey1: For result renamed to res, it is because I rewrote this line and I usually go with res instead of result, but I didn't notice that in the changeset, it is now rollbacked.

For toolset and toolSet, it is because I copied the function from the

    ´´´this.ipcHandle(
      'inference:streamText'´´´

But I can rollback it too

PR #336, opened 2025-09-15
chore: use mcp registry v1.0.0

### LINKED ISSUE #335: update mcp registry spec to v1.0.0
before working on other mcp-registry tasks assigned, need to use a fixed version of mcp registry

v1.0.0 has been tagged, should stick to it for now rather than 'main'

benoitf:
fixes #335

### DIFF
--- a/packages/renderer/src/lib/chat/components/suggested-actions.svelte
+++ b/packages/renderer/src/lib/chat/components/suggested-actions.svelte
@@ -32,7 +32,7 @@ const suggestedActions: SuggestedAction[] = [
     title: 'What are the last 5 issues of GitHub',
     label: 'repository podman-desktop/podman-desktop?',
     action: 'What are the last 5 issues of GitHub repository podman-desktop/podman-desktop?',
-    requiredMcp: ['fa3a90b04669d3658a6c1bbea9c079954b2d1509531344604fe56842a1c97182'],
+    requiredMcp: ['123e4567-e89b-12d3-a456-426614172000'],
   },
   {
     title: 'Write code to',


packages/mcp-registry/package.json
```diff
@@ -13,7 +13,7 @@
     "src"
   ],
   "scripts": {
-    "generate": "npx openapi-typescript https://raw.githubusercontent.com/modelcontextprotocol/registry/refs/heads/main/docs/server-registry-api/openapi.yaml -o src/schema.d.ts --export-type"
+    "generate": "npx openapi-typescript https://raw.githubusercontent.com/modelcontextprotocol/registry/refs/tags/v1.0.0/docs/reference/api/openapi.yaml -o src/schema.d.ts --export-type"
```
benoitf, line 16: the path also changed (to reference folder instead of server-registry-api)

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -385,10 +386,29 @@ export class MCPRegistry {
     for (const registryURL of serverUrls) {
       const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
 
-      // now, aggregate the servers from the list
-      serverDetails.push(...serverList.servers);
+      // now, aggregate the servers from the list adding an id being a sha of registryURL::server.name
+      serverDetails.push(
+        ...serverList.servers.map(server => {
+          let id = '';
```
jeffmaury: Maybe we should have the id extraction code as a separate method

benoitf: fixed

Issue #322, opened 2025-09-10
Support local Node MCP servers


jeffmaury:
The MCP servers that are registered on the registry with registry_type=npm

Issue #324, opened 2025-09-10
Support local container based MCP server


jeffmaury:
The MCP servers that are registered on the registry with registry_type=oci

Issue #323, opened 2025-09-10
Support local Python MCP servers


jeffmaury:
The MCP servers that are registered on the registry with registry_type=pypi

PR #341, opened 2025-09-15
feat(mcp): init mcp-spawn package (WIP)

axel7083:
## Description

The logic to spawn local MCP server should be placed in a dedicated package, currently there is no available library to do that from a MCP Registry, but might appear in the future, so let's have it in its own package.

Moreover, it ease the testing etc.

## Related issues

Fixes https://github.com/kortex-hub/kortex/issues/322
Required for
- https://github.com/kortex-hub/kortex/issues/323
- https://github.com/kortex-hub/kortex/issues/324

### DIFF
--- a/packages/renderer/src/lib/mcp/MCPRegistryCreateFromRegistryForm.svelte
+++ b/packages/renderer/src/lib/mcp/MCPRegistryCreateFromRegistryForm.svelte
@@ -57,9 +57,9 @@ async function navigateToMcps(): Promise<void> {
 
           {#if mcpTarget !== undefined}
             {#if 'url' in mcpTarget}  <!-- remote -->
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/ui/PasswordInput.svelte
```diff
@@ -2,13 +2,16 @@
 import { faEye, faEyeSlash } from '@fortawesome/free-regular-svg-icons';
 import { Input } from '@podman-desktop/ui-svelte';
 import { createEventDispatcher, onMount } from 'svelte';
+import type { FormEventHandler } from 'svelte/elements';
```
benoitf: FYI all changes in ui folders should be reported to Podman Desktop first

axel7083: https://github.com/podman-desktop/podman-desktop/pull/13953

PR #362, opened 2025-09-17
fix(chat): implement DB storage

### LINKED ISSUE #319: Implement persistence for chat

simonrey1:
fixes #319

https://github.com/user-attachments/assets/689b7eca-21a1-4fd2-8910-9d570de0ae67

To do in next PRs:

- refetch history when sending first message to display the updated list 
- name chat with a request to the LLM instead of constant "Chat"


simonrey1:
Security warning: exposing the method to delete chat by id does not check userId owns the chat

PR #367, opened 2025-09-17
fix(chat): implement DB storage

simonrey1:
needed for #319

https://github.com/user-attachments/assets/689b7eca-21a1-4fd2-8910-9d570de0ae67

To do in next PRs:

- refetch history when sending first message to display the updated list 
- name chat with a request to the LLM instead of constant "Chat"

Security warning: exposing the method to delete chat by id does not check userId owns the chat
Security warning 2: + you can get the messages for any chat with just the ids

### DIFF
--- a/packages/main/src/chat/chat-manager.ts
+++ b/packages/main/src/chat/chat-manager.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import { randomUUID } from 'node:crypto';
 import { existsSync } from 'node:fs';
 import { mkdir, readFile } from 'node:fs/promises';
 import { join } from 'node:path';
@@ -35,8 +36,13 @@ import type { InferenceParameters } from '/@api/chat/InferenceParameters.js';
 import { MCPManager } from '../plugin/mcp/mcp-manager.js';
 import { ProviderRegistry } from '../plugin/provider-registry.js';
 import { runMigrate } from './db/migrate.js';
+import { ChatQueries } from './db/queries.js';
+import type { Chat, Message } from './db/schema.js';
 
 export class ChatManager {
+  private chatQueries!: ChatQueries;
+  private userId!: string;
+
   constructor(
     @inject(ProviderRegistry)
     private readonly providerRegistry: ProviderRegistry,
@@ -61,12 +67,67 @@ export class ChatManager {
 
     runMigrate(db);
 
+    const chatQueries = new ChatQueries(db);
+
+    this.chatQueries = chatQueries;
+
+    async function getOrCreateUserId(): Promise<string> {
+      const defaultUserEmail = 'default@localhost';
+      const userGetter = await chatQueries.getUser(defaultUserEmail);
+      if (userGetter.isOk()) {
+        return userGetter.value.id;
+      } else {
+        const newUserGetter = await chatQueries.createAuthUser(defaultUserEmail, '');
+        if (newUserGetter.isOk()) {
+          return newUserGetter.value.id;
+        } else {
+          throw new Error('Cannot create user');
+        }
+      }
+    }
+
+    this.userId = await getOrCreateUserId();
+
     this.ipcHandle('inference:streamText', (_, params) => this.streamText(params));
     this.ipcHandle('inference:generate', (_, params) => this.generate(params));
+    this.ipcHandle('mcp-manager:getExchanges', (_, mcpId: string) => this.getExchanges(mcpId));
+    this.ipcHandle('inference:getChats', () => this.getChats());
+    this.ipcHandle('inference:getChatMessagesById', (_, id: string) => this.getChatMessagesById(id));
+    this.ipcHandle('inference:deleteChat', (_, id: string) => this.deleteChat(id));
+  }
 
-    this.ipcHandle('mcp-manager:getExchanges', async (_listener, mcpId: string): Promise<DynamicToolUIPart[]> => {
-      return this.mcpManager.getExchanges(mcpId);
-    });
+  private async getExchanges(mcpId: string): Promise<DynamicToolUIPart[]> {
+    return this.mcpManager.getExchanges(mcpId);
+  }
+
+  private async getChats(): Promise<Chat[]> {
+    return this.chatQueries.getChatsByUserId({ id: this.userId }).match(
+      chats => chats,
+      err => {
+        throw err;
+      },
+    );
+  }
+
+  private async getChatMessagesById(id: string): Promise<{ chat: Chat | null; messages: Message[] }> {
+    return (await this.chatQueries.getMessagesByChatId({ id })).match(
+      async messages => {
+        const chat = (await this.chatQueries.getChatById({ id })).match(
+          chat => chat,
+          err => {
+            throw err;
+          },
+        );
+        return { chat, messages };
+      },
+      err => {
+        throw err;
+      },
+    );
+  }
+
+  private async deleteChat(id: string): Promise<undefined> {
+    await this.chatQueries.deleteChatById({ id });
   }
 
   private async convertMessages(messages: UIMessage[]): Promise<UIMessage[]> {
@@ -93,6 +154,7 @@ export class ChatManager {
     tools: ToolSet;
     stopWhen: StopCondition<ToolSet>;
     system: string;
+    userMessage: UIMessage;
   }> {
     const internalProviderId = this.providerRegistry.getMatchingProviderInternalId(params.providerId);
     const sdk = this.providerRegistry.getInferenceSDK(internalProviderId, params.connectionName);
@@ -112,6 +174,7 @@ export class ChatManager {
 
     return {
       model,
+      userMessage,
       messages,
       tools,
       stopWhen: stepCountIs(5),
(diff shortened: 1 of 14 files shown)


packages/main/src/chat/db/migrate.ts
```diff
@@ -11,7 +11,7 @@ export const runMigrate = (db: BetterSQLite3Database<Record<string, unknown>>):
   // get current folder
   const currentFolder = resolve(__dirname);
 
-  const updateFolder = resolve(currentFolder, '..', 'migrations');
+  const updateFolder = resolve(currentFolder, '..', 'src', 'chat', 'db', 'migrations');
```
benoitf: to check: I think it might be a problem running the binary version (not the development version)

packages/main/src/chat/db/queries.ts
```diff
@@ -29,18 +29,39 @@ import {
...
+import { Directories } from '/@/plugin/directories.js';
+import { join } from 'node:path';
 
 // Optionally, if not using email/pass login, you can
 // use the Drizzle adapter for Auth.js / NextAuth
 // https://authjs.dev/reference/adapter/drizzle
 
 // biome-ignore lint: Forbidden non-null assertion.
-const sqlite = new Database('sqlite.db');
+const sqlite = new Database(join(new Directories().getConfigurationDirectory(), 'sqlite.db'));
```
benoitf: issue: use another specific folder ? configuration directory is for the configuration files

packages/main/src/chat/db/queries.ts
```diff
@@ -29,18 +29,39 @@ import {
...
 sqlite.pragma('foreign_keys = ON');
 export const db = drizzle(sqlite, {});
 
 runMigrate(db);
 
+
+async function initializeUserId() {
+  // Approach suggested by @benoitf
```
benoitf:
```suggestion
```

packages/main/src/plugin/index.ts
```diff
@@ -137,6 +138,16 @@ import type { ViewInfoUI } from '/@api/view-info.js';
...
+import {
+  deleteChatById,
+  getChatById,
+  getChatsByUserId,
+  getMessagesByChatId,
+  saveChat,
+  saveMessages,
+  userIdPromise,
+} from '../chat/db/queries.js';
+import type { Chat, Message } from '../chat/db/schema.js';
```
benoitf: suggestion: use absolute imports with '/'

packages/main/src/plugin/index.ts
```diff
@@ -1434,6 +1446,33 @@ export class PluginSystem {
...
+              chatId,
+              id: userMessage.id,
+              role: 'user',
+              parts: userMessage.parts,
+              createdAt: new Date(),
+              attachments: [],
+            },
+          ],
+        });
+
```
benoitf: issue: move all this content to a method of a dedicated class/file. It'll reduce imports of index.ts
usually index.ts is a no-op code (receive parameters and propagate to another object); no logic

packages/main/src/plugin/index.ts
```diff
@@ -1462,6 +1515,47 @@ export class PluginSystem {
...
+    this.ipcHandle('inference:getChats', async (_listener: Electron.IpcMainInvokeEvent): Promise<Chat[]> => {
+      const id = await userIdPromise;
+
+      return getChatsByUserId({ id }).match(
+        chats => chats,
+        err => {
+          throw err;
+        },
+      );
+    });
```
benoitf: issue: should be in a ChatManager (or aiManager, whatever) component

packages/main/src/chat/db/migrate.ts
simonrey1: Yes, Axel asked the same thing on the closed PR but I am not sure how to test it now

packages/main/src/chat/db/queries.ts
simonrey1: Should I create a new one?

benoitf: yes go ahead, `chat-persistence` (or something in the spirit)

also I think rather than sqlite.db it can be called chat.db (or any more meaningful word)

packages/main/src/chat/db/migrate.ts
benoitf: pnpm compile:current (and then starts the .dmg)

packages/main/src/chat/db/queries.ts
simonrey1: I have a problem if the chat-persistence directory does not exist, I will have to create the directory, so I need to use mkdir from 'fs/promises' as @axel7083 told me to use this, so the exported db const becomes a promise, and I have to await it in the queries files, so I have to make changes in the queries.ts file, but I am not sure if they will be approved?

For example:

```ts
export function deleteChatById({ id }: { id: string }): ResultAsync<undefined, DbError> {
  return safeTry(async function* () {
    const actions = [
      () => db.delete(vote).where(eq(vote.chatId, id)),
      () => db.delete(message).where(eq(message.chatId, id)),
      () => db.delete(chat).where(eq(chat.id, id)),
    ];

    for (const action of actions) {
      yield* fromPromise(action(), e => new DbInternalError({ cause: e }));
    }

    return ok(undefined);
  });
}
```

becomes
```ts

export async function deleteChatById({ id }: { id: string }): Promise<ResultAsync<undefined, DbError>> {
  return safeTry(async function* () {
    const actions = [
      async() => (await db).delete(vote).where(eq(vote.chatId, id)),
      async() => (await db).delete(message).where(eq(message.chatId, id)),
      async() => (await db).delete(chat).where(eq(chat.id, id)),
    ];

    for (const action of actions) {
      yield* fromPromise(action(), e => new DbInternalError({ cause: e }));
    }

    return ok(undefined);
  });
}
```

packages/main/src/chat/db/migrate.ts
simonrey1: Indeed it does not seem to be working, the migrations files are not exported in the built executable (or they are in another location)

Screenshot 2025-09-18 at 12 21 13

simonrey1: Screenshot 2025-09-18 at 12 31 45

I could move the migrations to a subdirectory of existing assets directory, or add extraResources, which one shoud I go with?

benoitf: can you import files so they become vite assets ?

benoitf: else doing like roots.exe works for me

simonrey1: I have added it, and was able to recreate the database by launching the .dmg app

packages/main/src/plugin/index.ts
simonrey1: I have made a PR to extract the existing inference methods in a class

https://github.com/kortex-hub/kortex/pull/381/files

so I can add other in this one

packages/main/src/chat/db/queries.ts
benoitf:
```suggestion
```

packages/main/src/chat/db/queries.ts
benoitf: I would have a init() method called from the chat-manager ?

so init method would also contain the call to runMigrate(db); in async mode.

Once the Chat Manager is ready it means, we can use db field directly

simonrey1: https://github.com/kortex-hub/kortex/pull/395

packages/main/src/chat/chat-manager.ts
```diff
@@ -61,12 +67,67 @@ export class ChatManager {
...
+    return this.chatQueries.getChatsByUserId({ id: this.userId }).match(
+      chats => chats,
+      err => {
+        throw err;
+      },
+    );
+  }
+
+  private async getChatMessagesById(id: string): Promise<{ chat: Chat | null; messages: Message[] }> {
+    return (await this.chatQueries.getMessagesByChatId({ id })).match(
```
simonrey1, line 115: There is no check that the user has rights to read these messages

benoitf:
I'm able to see my previous chats but I've noticed some issues:

1. sidebar seems to grow too much
image

2. if I jump to another section and come back, I don't see my current chat session, I need to manually go to the chat

https://github.com/user-attachments/assets/1cfd52a6-bf48-4550-89a3-ea96e8c6a3c9

3. when coming back to the chat, if sidebar was opened, it's no longer opened (and then I don't see quickly I have a previous chats/messages)

4. It's displayed "Guest", I would expect to see my first "Florent" from my computer/current sessions

benoitf:
all chats are named "chat" making it difficult to navigate back to a previous entry

image

benoitf:
if the sidebar is opened, the new chat button is not there
image


image

simonrey1:
It seems better with relative postionning instead of fixed, I pushed a fix to the existing sidebar here, or should I push it in main first?

Screenshot 2025-09-22 at 12 22 00

benoitf:
@simonrey1 if simpler, do not restore sidebar as part of this PR but make that coming back to the chat window we see what we had before

simonrey1:
Sorry I did not see the messages in the meantime, I have added full name as suggested

Screenshot 2025-09-22 at 12 52 42

simonrey1:
For me it is ok to merge with this set of features, I will continue to work to select the current chat

PR #372, opened 2025-09-17
feat: add description of mcp server

benoitf:
when listing mcp servers from the official mcp registry the id is not meaningful. Having the description helps.

depends on:
- [x] https://github.com/kortex-hub/kortex/pull/374

### DIFF
--- a/packages/renderer/src/lib/mcp/MCPServerListRemoteReady.svelte
+++ b/packages/renderer/src/lib/mcp/MCPServerListRemoteReady.svelte
@@ -1,12 +1,12 @@
 <script lang="ts">
 import { Table, TableColumn, TableRow } from '@podman-desktop/ui-svelte';
-import SimpleColumn from '@podman-desktop/ui-svelte/TableSimpleColumn';
 
 import MCPNameColumn from '/@/lib/mcp/column/MCPNameColumn.svelte';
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/mcp/MCPServerListRemoteReady.svelte
```diff
@@ -31,13 +32,20 @@ const nameColumn = new TableColumn<MCPRemoteServerInfo>('Name', {
   comparator: (a, b): number => b.name.localeCompare(a.name),
 });
 
+const descriptionColumn = new TableColumn<MCPRemoteServerInfo, string>('Description', {
```
simonrey1: suggestion: move common columns (name, description, and other if applicable) to dedicated file with a generic that extends { name: string; description:  string; } and call it with  MCPServerDetail and MCPRemoteServerInfo to get the columns

simonrey1: but it seems that MCPNameColumn does not apply to packages/renderer/src/lib/mcp/MCPServerListRegistryInstall.svelte so I am not sure if this is possible

benoitf: added a common Column

PR #373, opened 2025-09-18
chore: apply lint:fix command

benoitf:


benoitf:
here it's only fixing automated fixes (import sorting), not all the errors

jeffmaury:
> lint:fix fixes all "auto-fixable" issues (no manual/user action)
> 
> but I'm looking at the non automated fixes (will come in a separate PR)

Just guessed I missed the scope so approving

PR #374, opened 2025-09-18
chore: keep description field in the mcp server (if any)

benoitf:
pre-req for https://github.com/kortex-hub/kortex/pull/372

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-registry.ts
+++ b/packages/main/src/plugin/mcp/mcp-registry.ts
@@ -324,6 +324,7 @@ export class MCPRegistry {
     }
 
     const name = serverDetail.name;
+    const description = serverDetail.description;
 
     // create transport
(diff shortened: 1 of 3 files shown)


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -324,6 +324,7 @@ export class MCPRegistry {
     }
 
     const name = serverDetail.name;
+    const description = serverDetail.description;
```
simonrey1: suggestion: avoid introducing const for name and description, as they are very generic names only used in the registerMCPClient call, it is possible to use directly serverDetail.name, and serverDetail.description

simonrey1 (approved):
LGTM, not tested

packages/main/src/plugin/mcp/mcp-registry.ts
benoitf: moved to destruct

Issue #353, opened 2025-09-17
add support only for latest versions coming from a mcp registry


benoitf:
We may have multiple versions of an mcp server defined in a registry.
Grab only latest versions

PR #375, opened 2025-09-18
chore: get only latest versions from the registry

benoitf:
fixes https://github.com/kortex-hub/kortex/issues/353


jeffmaury (approved):
LGTM

simonrey1 (approved):
LGTM
Open question: What happens when we have a MCP server installed in Kortex, and they update it?

Issue #363, opened 2025-09-17
add ci PR check


benoitf:
basic workflow with building/linting/formatting/unit tests

PR #376, opened 2025-09-18
chore: fix missing dependency (linting error)

benoitf:
this dependency is a direct dependency so it should be listed currently it works as it is a transitive dependency

related: https://github.com/kortex-hub/kortex/issues/363


axel7083 (approved):

jeffmaury (approved):
LGTM

PR #377, opened 2025-09-18
chore: fix linting report

benoitf:
here it's more than just adding parameter or return type

looks like condition is always returning the same value, so simplify the function


simonrey1:
This code was copied from https://github.com/vercel/ai-chatbot-svelte/blob/9247ebb9a4bba39284e144c98b22a496c158c271/src/lib/server/ai/prompts.ts#L39

PR #378, opened 2025-09-18
chore: fix import broken by eslint autofix

simonrey1:
follow up of https://github.com/kortex-hub/kortex/pull/373


benoitf:
thanks

PR #381, opened 2025-09-18
refactor(chat): create dedicated class

simonrey1:
prerequisite for https://github.com/kortex-hub/kortex/pull/366 and https://github.com/kortex-hub/kortex/pull/367

### DIFF
--- a/packages/main/src/chat/chat-manager.ts
+++ b/packages/main/src/chat/chat-manager.ts
@@ -1,3 +1,20 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


packages/main/src/chat/chat-manager.ts
```diff
@@ -0,0 +1,121 @@
+import { readFile } from 'node:fs/promises';
```
benoitf, line 18: issue: please add copyright headers

packages/main/src/chat/chat-manager.ts
```diff
@@ -0,0 +1,121 @@
...
+export class ChatManager {
+  constructor(
+    private readonly providerRegistry: ProviderRegistry,
+    private readonly mcpManager: MCPManager,
+    private readonly getWebContentsSender: () => WebContents,
+    private readonly ipcHandle: (
+      channel: string,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      listener: (event: IpcMainInvokeEvent, ...args: any[]) => Promise<void> | any,
+    ) => void,
```
benoitf, line 37: just a thought: I was expecting an existing interface/ better type in our existing code

simonrey1: So do I. Maybe I should create it in podman-desktop?

benoitf: might be interesting for upstreaming/downstreaming process but definitely not a blocker

PR #382, opened 2025-09-18
fix(chat-manager): fix generate method, refactor common method and interface for params of stream and generate text

simonrey1:
draft waiting for https://github.com/kortex-hub/kortex/pull/381 to be merged first

needed for #366

### DIFF
--- a/packages/api/src/chat/InferenceParameters.ts
+++ b/packages/api/src/chat/InferenceParameters.ts
@@ -1,0 +1,9 @@
+import type { UIMessage } from 'ai';
+
+export interface InferenceParameters {
+  providerId: string;
+  connectionName: string;
+  modelId: string;
+  mcp: string[];
+  messages: UIMessage[];
+}
--- a/packages/api/src/chat/InferenceParameters.ts
+++ b/packages/api/src/chat/InferenceParameters.ts
@@ -1,3 +1,21 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


packages/api/src/chat/InferenceParameters.ts
```diff
@@ -0,0 +1,9 @@
+import type { UIMessage } from 'ai';
```
axel7083, line 19: missing headers

packages/main/src/chat/chat-manager.ts
```diff
@@ -64,39 +67,39 @@ export class ChatManager {
...
-    mcp: Array<string>,
-    messages: UIMessage[],
-    onDataId: number,
-  ): Promise<number> {
-    const internalProviderId = this.providerRegistry.getMatchingProviderInternalId(providerId);
-    const sdk = this.providerRegistry.getInferenceSDK(internalProviderId, connectionName);
-    const languageModel = sdk.languageModel(modelId);
+  private async getInferenceComponents(
+    params: InferenceParameters,
+  ): Promise<Parameters<typeof streamText>[0] & Parameters<typeof generateText>[0]> {
```
axel7083: return type is pretty unclear to me, can we add some comments on the function, to explain _shortly_ what it is doing?

packages/main/src/chat/chat-manager.ts
```diff
@@ -64,39 +67,39 @@ export class ChatManager {
...
+      stopWhen: stepCountIs(5),
       system: 'You are a friendly assistant! Keep your responses concise and helpful.',
-      tools: toolset,
+    };
+  }
 
-      stopWhen: stepCountIs(5),
-    });
+  async streamText(
+    _listener: Electron.IpcMainInvokeEvent,
```
axel7083: can't we omit this ?

axel7083 (approved):

packages/main/src/chat/chat-manager.ts
simonrey1: I omitted it but used 

```
(_, params) => this.generate(params)
```

instead of bind, so let me know if you preferred bind

packages/main/src/chat/chat-manager.ts
simonrey1: I explicited the types in a new commit

PR #385, opened 2025-09-18
fix(electron): use electron-builder rebuild capacity instead of other package

simonrey1:
Following PR #344 and another one about electron/rebuild

I have seen that we might be able to rebuild using electron-builder


simonrey1:
Screenshot 2025-09-18 at 16 40 04

it seems to do something for better-sqlite3, LGTM

jeffmaury (approved):
LGTM on Win11Pro

PR #392, opened 2025-09-18
feat: enabled local openai url to be added as an inference server

bmahabirbu:
a poc for the moment!

### DIFF
--- a/packages/renderer/src/lib/chat/components/NoModelsAvailable.svelte
+++ b/packages/renderer/src/lib/chat/components/NoModelsAvailable.svelte
@@ -1,22 +1,81 @@
 <script lang="ts">
+import { Button } from '@podman-desktop/ui-svelte';
+import { toast } from 'svelte-sonner';
+import { router } from 'tinro';
+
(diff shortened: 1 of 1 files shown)


bmahabirbu:
Tested with ramalama and works for non mcp use cases. For some reason using the github mcp has an insane token prompt will continue to debug this because it should work

packages/renderer/src/lib/chat/components/NoModelsAvailable.svelte
```diff
@@ -1,22 +1,81 @@
...
+  }
+}
+
+// Auto-enable OpenAI extension when needed
+async function ensureOpenAIEnabled(): Promise<void> {
+  const openaiProvider = $providerInfos.find(provider => provider.id === 'openai');
+  if (!openaiProvider) {
+    toast.info('Enabling OpenAI extension...');
+    await window.ensureExtensionIsEnabled('kortex.openai');
+    await window.startExtension('kortex.openai');
```
benoitf, line 30: FYI Podman Desktop core cannot depend on a specific extension name. It depends on capabilities like is there a FlowConnection provider, inference server registered, etc.

benoitf:
setting to draft as body is "A POC"

PR #395, opened 2025-09-19
fix(chat): introduce a class for queries, async init the db in the manager

simonrey1:
pre-requisite for #319

### DIFF
--- a/packages/main/src/chat/chat-manager.ts
+++ b/packages/main/src/chat/chat-manager.ts
@@ -28,8 +28,8 @@ import { drizzle } from 'drizzle-orm/better-sqlite3';
 import type { WebContents } from 'electron';
 import { inject } from 'inversify';
 
-import { Directories } from '/@/plugin/directories.js';
 import { IPCHandle, WebContentsType } from '/@/plugin/api.js';
+import { Directories } from '/@/plugin/directories.js';
 import type { InferenceParameters } from '/@api/chat/InferenceParameters.js';
 
 import { MCPManager } from '../plugin/mcp/mcp-manager.js';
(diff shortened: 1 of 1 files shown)


jeffmaury (commented):
Hard to review the queries.ts file: why are there so much changes ?

packages/main/src/plugin/index.ts
```diff
@@ -609,7 +609,7 @@ export class PluginSystem {
     flowManager.init();
 
     const chatManager = new ChatManager(providerRegistry, mcpManager, this.getWebContentsSender, this.ipcHandle);
-    chatManager.init();
+    await chatManager.init();
```
benoitf: I have not catched that earlier, but ChatManager instance should be instantiated by Inversify

packages/main/src/chat/chat-manager.ts
```diff
@@ -40,7 +46,17 @@ export class ChatManager {
     ) => void,
   ) {}
 
-  public init(): void {
+  public async init(): Promise<void> {
+    const directory = new Directories().getChatPersistenceDirectory();
```
benoitf: issue: we should not create a new Directories instance, it needs to be injected through inversify

benoitf (changes requested):

benoitf:
will need https://github.com/kortex-hub/kortex/pull/396

packages/main/src/chat/chat-manager.ts
benoitf: add inject of Directories

benoitf (approved):
LGTM but I've not checked if methods on the DB are working

PR #400, opened 2025-09-19
chore: update pnpm version

benoitf:
dependabot says file is not parseable...

not sure it'll fix neither

https://github.com/kortex-hub/kortex/actions/workflows/dependabot/dependabot-updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,5 +193,5 @@
       "@kubernetes/client-node": "patches/@kubernetes__client-node.patch"
     }
   },
-  "packageManager": "pnpm@10.15.0+sha512.486ebc259d3e999a4e8691ce03b5cac4a71cbeca39372a9b762cb500cfdf0873e2cb16abe3d951b1ee2cf012503f027b98b6584e4df22524e0c7450d9ec7aa7b"
+  "packageManager": "pnpm@10.17.0+sha512.fce8a3dd29a4ed2ec566fb53efbb04d8c44a0f05bc6f24a73046910fb9c3ce7afa35a0980500668fa3573345bd644644fa98338fa168235c80f4aa17aa17fbef"
 }


package.json
```diff
@@ -193,5 +193,5 @@
       "@kubernetes/client-node": "patches/@kubernetes__client-node.patch"
     }
   },
-  "packageManager": "pnpm@10.15.0+sha512.486ebc259d3e999a4e8691ce03b5cac4a71cbeca39372a9b762cb500cfdf0873e2cb16abe3d951b1ee2cf012503f027b98b6584e4df22524e0c7450d9ec7aa7b"
+  "packageManager": "pnpm@10.17.0+sha512.fce8a3dd29a4ed2ec566fb53efbb04d8c44a0f05bc6f24a73046910fb9c3ce7afa35a0980500668fa3573345bd644644fa98338fa168235c80f4aa17aa17fbef"
```
simonrey1, line 196: In my opinion, we should not include the SHA because pnpm does not use it.

See function `parsePackageManager` here:

https://github.com/pnpm/pnpm/blob/702ddb9ee85c3c640b8bfb0f21b5cc31ebdd7626/config/config/src/index.ts#L579

benoitf: looks like you're still on the subject about sha not being useful https://github.com/podman-desktop/podman-desktop/pull/13717#issuecomment-3224172517

integrity check is made by `corepack`

modify the hash and run `corepack install` for example, you will see a hash issue

benoitf: https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages

> The hash is optional but strongly recommended as a security practice

simonrey1: I reproduced the error:

`corepack install`
->
`Internal Error: Mismatch hashes. `

when changing the hash.

Should we use corepack install instead of pnpm install? What is the best practise?

simonrey1: my bad, it's in the README

`- [pnpm v9.x](https://pnpm.io/installation) (`corepack enable pnpm`)`

Maybe you should update this to pnpm v10 in this PR to catch up too?

simonrey1: > https://github.com/nodejs/corepack?tab=readme-ov-file#when-authoring-packages
> 
> > The hash is optional but strongly recommended as a security practice

Yes but it feels like a false security to me if we include the hash and it is not used.

I wonder if Github Action [pnpm/action-setup](https://github.com/pnpm/action-setup) uses it.

benoitf:
the main issue is that dependabot is not working, so want to check if it's related to the version or something else

axel7083 (approved):

PR #405, opened 2025-09-19
chore: update dmg background

benoitf:
provided by @vancura 


benoitf:
well today it displays 'podman desktop' so better have the name of the repository than podman desktop

vancura:
I will prepare a new background image when we have a new name :)

PR #412, opened 2025-09-22
fix: store added MCP registries in a config file

### LINKED ISSUE #346: adding a new mcp registry is not persisted
If I add a custom mcp registry and restart Kortex, it's no longer there

jeffmaury:
Fixes #346

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-registry.ts
+++ b/packages/main/src/plugin/mcp/mcp-registry.ts
@@ -17,6 +17,9 @@
  ***********************************************************************/
 
 import * as crypto from 'node:crypto';
+import { existsSync } from 'node:fs';
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
+import { dirname, join } from 'node:path';
 
 import type * as kortexAPI from '@kortex-app/api';
 import { SecretStorage } from '@kortex-app/api';
@@ -33,6 +36,7 @@ import { InputWithVariableResponse, MCPSetupOptions } from '/@api/mcp/mcp-setup.
 
 import { ApiSenderType } from '../api.js';
 import { Certificates } from '../certificates.js';
+import { Directories } from '../directories.js';
 import { Emitter } from '../events/emitter.js';
 import { Proxy } from '../proxy.js';
 import { Telemetry } from '../telemetry/telemetry.js';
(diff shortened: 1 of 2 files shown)


benoitf (commented):
hello, wondering why it's not part of the default configuration file ?

so it could be seen in the preferences as well (or hidden) but at least it's part of the default configuration

benoitf:
it would also fit in a "managed" configuration file.

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -199,6 +206,9 @@ export class MCPRegistry {
       return Disposable.noop();
     }
     this.registries = [...this.registries, registry];
+    this.saveRegistriesToConfig().catch((error: unknown) => {
+      console.error('[MCPRegistry] Failed to save registries after registration:', error);
+    });
```
benoitf: suggestion: make it non async so we can only call : saveRegistriesToConfig ?

and it's inside the method that we catch the error/log it

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -101,6 +106,7 @@ export class MCPRegistry {
     if (this.proxyEnabled) {
       this.proxySettings = this.proxy.proxy;
     }
+    this.configuration = this.configurationRegistry.getConfiguration('mcp');
```
benoitf: IMHO there is a missing registration of the key into the configuration registry

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -535,4 +548,12 @@ export class MCPRegistry {
     }
     return options;
   }
+
+  private async loadRegistriesFromConfig(): Promise<void> {
+    this.registries = this.configuration.get<kortexAPI.MCPRegistry[]>('registries') ?? [];
```
benoitf: how do we handle registries coming from an extension vs manual registration

I would expect to store only "user" registries, not the one registered by an extension (if I disable an extension the registry should not come again ?)

Also for that I would save only the registry URL (not any other data)

benoitf:
@jeffmaury did you forgot to push changes ?

simonrey1:
> before I was seeing previous implementation and now it's fixed. I don't know if it was my browser's issue or GitHub issue

Have you switched to the "New experience" for reviewing PRs? It seems a bit less buggy to me, but GitHub definilety has trouble to load all commits these days.

PR #416, opened 2025-09-22
fix(chat): keep current chat id in global state

simonrey1:
https://github.com/user-attachments/assets/7c76c997-8fb8-41c7-9c0d-716cebda2daa

Partial fix of #319

### DIFF
--- a/packages/renderer/src/lib/chat/route/CustomChat.svelte
+++ b/packages/renderer/src/lib/chat/route/CustomChat.svelte
@@ -17,7 +17,9 @@ import { convertToUIMessages } from '../utils/chat';
 interface Props {
   chatId?: string;
 }
-const { chatId = currentChatId.value }: Props = $props();
+const { chatId: routerChatId }: Props = $props();
+
+const chatId = $derived(routerChatId ?? currentChatId.value);
 
 const chatsPromise = window.inferenceGetChats();
 const chatHistory = new ChatHistory(chatsPromise);
--- a/packages/renderer/src/lib/chat/route/CustomChat.svelte
+++ b/packages/renderer/src/lib/chat/route/CustomChat.svelte
@@ -7,6 +7,7 @@ import Chat from '/@/lib/chat/components/chat.svelte';
 import { SidebarInset, SidebarProvider } from '/@/lib/chat/components/ui/sidebar';
 import { Toaster } from '/@/lib/chat/components/ui/sonner';
 import { ChatHistory } from '/@/lib/chat/hooks/chat-history.svelte.js';
+import { currentChatId } from '/@/lib/chat/stores/current-chat-id.svelte.js';
 import { sidebarCollapsed } from '/@/lib/chat/stores/sidebar-collapsed';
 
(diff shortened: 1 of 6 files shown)


packages/renderer/src/lib/chat/route/CustomChat.svelte
```diff
@@ -43,7 +44,7 @@ onMount(() => {
     Loading
   {:then { chat, messages }} 
     <SidebarProvider open={!$sidebarCollapsed} onOpenChange={(open: boolean): void => sidebarCollapsed.set(!open)}>
-      <AppSidebar {chatId} />
+      <AppSidebar chatId={chatId} />
```
benoitf: <AppSidebar {chatId} /> is not working ?

packages/renderer/src/lib/chat/stores/current-chat-id.ts
```diff
@@ -0,0 +1,3 @@
+import { writable } from 'svelte/store';
+
+export const currentChatId = writable<string | undefined>();
```
benoitf: replace the writable store by a global $state

packages/renderer/src/lib/chat/route/CustomChat.svelte
simonrey1: oops, could not get used to it coming from ReactJs :D ! Here is the correction

packages/renderer/src/lib/chat/stores/current-chat-id.ts
simonrey1: I tested that it works! just pushed :)

simonrey1: Sorry I think i have an issue with the use in IPCChatTransport

benoitf: what's the issue ?

move it to a .svelte.ts file, not .ts file

simonrey1: I am not sure what you mean by global $state?

Like this for instance?

```ts

let count = $state(0);

export function get() {
  return count;
}

export function set(new_count) {
  count = new_count;
}
```

source: https://mainmatter.com/blog/2025/03/11/global-state-in-svelte-5/

benoitf: yes

simonrey1: I also found this syntax which I could prefer to avoid aving the get and set methods

`export const count = $state({ value: 0 })`

source: https://joyofcode.xyz/how-to-share-state-in-svelte-5

benoitf: I mean, I want a New svelte5 $state, not a writable/store as we're not interested in events

simonrey1: Ok I understand better after a bit of reading about signals.

I pushed a new version using it.

I just noticed it is still in the stores folder, where could I put it?

PR #417, opened 2025-09-22
fix: allow filtering MCP servers

### LINKED ISSUE #348: add search field for MCP servers
adding mcp registries can lead to hunderd of servers

it should be possible to use a search input to filter the MCP servers being listed

jeffmaury:
Filtering is based on server name, can also use description

Fixes #348


packages/renderer/src/lib/mcp/MCPServerListRemoteReady.svelte
```diff
@@ -10,15 +10,23 @@ import { MCPServerDescriptionColumn } from './mcp-server-columns';
...
-  $mcpRemoteServerInfos.map(server => ({
-    ...server,
-    selected: false,
-  })),
+  $mcpRemoteServerInfos
+    .map(server => ({
+      ...server,
+      selected: false,
+    }))
+    .filter(server => server.name.toLowerCase().includes(filter?.toLowerCase() ?? '')),
```
simonrey1: suggestion: only filter when there is a filter because there can be a lot of mcp servers

PR #438, opened 2025-09-23
fix(chat): navigate to other selected chat

simonrey1:
https://github.com/user-attachments/assets/7bbe366c-5f78-478e-842d-d24d53d677ae

Follow up https://github.com/kortex-hub/kortex/pull/437 I am now able to simplify the promises and fix the navigation between chats

### DIFF
--- a/packages/renderer/src/lib/chat/route/CustomChat.svelte
+++ b/packages/renderer/src/lib/chat/route/CustomChat.svelte
@@ -22,7 +22,9 @@ const chatsPromise = window.inferenceGetChats();
 const chatHistory = new ChatHistory(chatsPromise);
 chatHistory.setContext();
 
-const chatMessagesPromise = $derived(chatId ? window.inferenceGetChatMessagesById(chatId) : Promise.resolve(undefined));
+const chatMessagesPromise = $derived(
+  chatId ? window.inferenceGetChatMessagesById(chatId) : Promise.resolve({ chat: undefined, messages: [] }),
+);
 
 let selectedChatModel: SelectedModel | undefined = $state(undefined);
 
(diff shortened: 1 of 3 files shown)


packages/renderer/src/lib/chat/route/CustomChat.svelte
```diff
@@ -46,13 +37,13 @@ onMount(() => {
...
-  {#await dataPromise()}
+  {#await chatMessagesPromise}
     Loading
-  {:then data} 
+  {:then chatMessages} 
     <SidebarProvider open={!$sidebarCollapsed} onOpenChange={(open: boolean): void => sidebarCollapsed.set(!open)}>
       <AppSidebar {chatId} />
       <SidebarInset>
-        <Chat chat={'chatMessages' in data ? data.chatMessages?.chat ?? undefined : undefined} initialMessages={'chatMessages' in data ? convertToUIMessages(data.chatMessages.messages) : []} readonly={false}  />
+        <Chat chat={chatMessages?.chat ?? undefined} initialMessages={chatMessages?.messages ? convertToUIMessages(chatMessages.messages) : []} readonly={false}  />
```
benoitf: suggestion: this part looks complicated because we have code in svelte

simplify the svelte part to just give the values having values updated during an effect()

<Chat {value1} {value2}>

benoitf: suggestion: remove the #await pattern and just make it simple, moving derived into effect

let chatMessages = $state([])

```
$effect(() => {
  window.inferenceGetChatMessagesById(chatId).then((messages) => chatMessages = messages
}
```

in short, do not handle promises in svelte/html, do it in the code

simonrey1: I will simplify

simonrey1: Sorry did not see your second comment, why not using derived and effect from svelte?

benoitf: because you could handle objects rather than promises

code is easier to review/understand.
no "code logic" in the svelte/html part, only display stuff

benoitf: if you have another promise, you'll need to refactor a lot of stuff in the svelte/html and it'll become barely readable

simonrey1: I saw that earlier with the fullname promise it was not working well

benoitf: this is why we're mostly never using #await in the codebase

PR #440, opened 2025-09-23
feat(backend: mcp): support spawn local mcp

axel7083:
## Description

_EXPERIMENTAL_

Introducing a `mcp-setup` package, we could find a better name, but the goal would be to move all the logic for connecting (remote) and spawn (local) MCP servers.

> ⚠️ the errors are pretty terrible, currently the stderr of the spawn command is piped into Kortex stderr, this should be displayed in the Ui in the future
> ⚠️ since https://github.com/kortex-hub/kortex/pull/412 is not merged, I haven't tested with persistence, IN THEORY, MCP should reboot at startup, but this is not guaranteed 

## Screenshots

> ℹ️  You can combine this PR with https://github.com/kortex-hub/kortex/pull/436 to have a nice UI and replicate the video bellow
> ⚠️  I used the registry `https://axel7083.github.io/mcp-registry-online-v1.0.0`, and only the `@modelcontextprotocol/server-filesystem` is working, no other have been tested fixed to works.

https://github.com/user-attachments/assets/499d6f02-099e-4b80-a038-775f4295f837

## Testing

- [x] unit tests for the `mcp-setup` package have been adde

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-registry.ts
+++ b/packages/main/src/plugin/mcp/mcp-registry.ts
@@ -26,6 +26,7 @@ import type { HttpsOptions, OptionsOfTextResponseBody } from 'got';
 import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
 import { inject, injectable } from 'inversify';
 import type { components } from 'mcp-registry';
+import { getMCPSpawner } from 'mcp-setup';
 
 import { SafeStorageRegistry } from '/@/plugin/safe-storage/safe-storage-registry.js';
 import { IConfigurationNode, IConfigurationRegistry } from '/@api/configuration/models.js';
@@ -49,6 +50,9 @@ interface RemoteStorageConfigFormat {
 interface PackageStorageConfigFormat {
   serverId: string;
   packageId: number;
+  packageArguments: { [key: number]: string };
+  runtimeArguments: { [key: number]: string };
+  environmentVariables: { [key: string]: string };
 }
 
 type StorageConfigFormat = RemoteStorageConfigFormat | PackageStorageConfigFormat;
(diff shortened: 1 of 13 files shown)


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -374,6 +416,35 @@ export class MCPRegistry {
...
+        ...argument,
+        value: config.packageArguments[index],
+        variables: undefined,
+      })),
+      environment_variables: (pack.environment_variables ?? []).map(argument => ({
+        ...argument,
+        value: config.environmentVariables[argument.name],
+        variables: undefined,
+      })),
+    };
```
benoitf, line 475: I've noticed that now they changed to camelCase... 😢
https://github.com/modelcontextprotocol/registry/pull/522

axel7083: Was a bit annoying but thanks to typescript, fixed all the problem :+1:

packages/mcp-setup/src/spawner/npm-spawner.ts
```diff
@@ -0,0 +1,77 @@
...
+    const PACKAGE_ARGS: Array<string> = (this.pack.packageArguments ?? []).map(this.getArgument.bind(this));
+
+    // let's use package@version if version is specified
+    const identifier = this.pack.version ? `${this.pack.identifier}@${this.pack.version}` : this.pack.identifier;
+
+    // environments variables
+    const ENVIRONMENTS = this.getEnvironments();
+
+    const transport = new StdioClientTransport({
+      command: NPX_COMMAND,
```
jeffmaury, line 47: Seems to me that -y flag is missing

axel7083: This is unclear if the server should be providing it, or if we should include it ourself, MCP Registry provide a `runtimeArguments`, where the `--yes` could be defined.

Issue #402, opened 2025-09-19
detect ollama running models


benoitf:
ollama extension should report available models

PR #443, opened 2025-09-24
feat: add ollama extension registering ollama models

benoitf:
fixes https://github.com/kortex-hub/kortex/issues/402


extensions/ollama/src/extension.ts
```diff
@@ -0,0 +1,39 @@
...
+  await ollamaExtension?.deactivate();
+  ollamaExtension = undefined;
+}
+
+// Expose ollamaExtension for testing purposes
+if (process.env.NODE_ENV === 'test') {
+  Object.defineProperty(global, 'ollamaExtension', {
+    get: () => ollamaExtension,
+  });
+}
```
axel7083: Why are you defining the global property `ollamaExtension` in test mode, this is not something I would expect in an extension.

simonrey1: I tried to remove this code, and the tests are still passing so I think it can be safely removed.

Image

Image

extensions/ollama/src/extension.ts
```diff
@@ -0,0 +1,39 @@
...
+  await ollamaExtension.activate();
+}
+
+export async function deactivate(): Promise<void> {
+  await ollamaExtension?.deactivate();
+  ollamaExtension = undefined;
+}
+
+// Expose ollamaExtension for testing purposes
+if (process.env.NODE_ENV === 'test') {
```
jeffmaury: Looks weird to me ?

extensions/ollama/src/extension.ts
benoitf: @simonrey1 I don't see how it could pass the test if you remove it ?

benoitf: ps: there are more than 3 tests

extensions/ollama/src/extension.ts
benoitf: how can you check the extension is set to undefined in deactivate otherwise ?

extensions/ollama/src/extension.ts
benoitf: > Why are you defining the global property ollamaExtension in test mode, this is not something I would expect in an extension.

how do you want to add tests on this property if it's not exposed ? also it's only exposed in test mode, not in production mode

simonrey1: oops: I did 

pnpm test ollama

and indeed these tests were not detected

extensions/ollama/src/extension.spec.ts

extensions/ollama/src/ollama-extension.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.#connectionDisposable = undefined;
+      }
+      this.#currentModels = newModelNames;
+      if (newModelNames.length > 0) {
+        const sdk = createOllama();
+        const disposable = ollamaProvider.registerInferenceProviderConnection({
+          name: 'ollama',
+          sdk,
+          status() {
+            return running ? 'started' : 'stopped';
```
simonrey1: Image

It is possible to return 'started' always here as running is true (see screenshot)

extensions/ollama/src/ollama-extension.ts
```diff
@@ -0,0 +1,115 @@
...
+      });
+    }, 30000);
+  }
+
+  protected async updateModelsAndStatus(ollamaProvider: Provider): Promise<void> {
+    let models: Array<{ name: string }> = [];
+    let running = true;
+    try {
+      const res = await fetch('http://localhost:11434/api/tags');
+      if (!res.ok) throw new Error('Ollama not running');
```
simonrey1: suggestion to make error more generic, because it could also be an internal server error, even with Ollama running, and res.ok would be false

extensions/ollama/src/ollama-extension.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    ];
+    server = setupServer(...handlers);
+    server.listen({ onUnhandledRequest: 'error' });
+    await extension.activate();
+    expect(vi.mocked(ollamaProvider.updateStatus)).toHaveBeenCalledWith('stopped');
+    expect(vi.mocked(ollamaProvider.registerInferenceProviderConnection)).not.toHaveBeenCalled();
+  });
+
+  test('should unregister and register new connection if models change', async () => {
+    let callCount = 0;
```
simonrey1: suggestion: my preference goes for this (tested)

```ts

    const models = [{ name: 'm1' }];
    const handlers = [
      http.get('http://localhost:11434/api/tags', () => {
        return HttpResponse.json({ models });
      }),
    ];
    server = setupServer(...handlers);
    server.listen();
    await extension.activate();
    expect(ollamaProvider.registerInferenceProviderConnection).toHaveBeenCalledTimes(1);
    // Simulate timer
    models.push({ name: 'm2' });

    await extension.updateModelsAndStatus(ollamaProvider);
    expect(ollamaProvider.registerInferenceProviderConnection).toHaveBeenCalledTimes(2);

```

extensions/ollama/src/extension.ts
axel7083: expect `OllamaExtension.prototype.deactivate` to have been called is enough, as in `activate` we overwrite the value, so there is not risk after re-activating the extension to reuse an old instance.

IMO avoiding introducing test specific code in the logic is more important than ensuring that this variable is undefined.

extensions/ollama/scripts/build.js
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env node
+/**********************************************************************
+ * Copyright (C) 2022-2023 Red Hat, Inc.
```
gastoner, line 3: 2025

gastoner (approved):
Tried out on Linux ✔️

benoitf:
@jeffmaury qwen3:8b ?

slemeur:
Do we have a way to inform the user that the interaction will not be working if MCP are used with a specific model?
I don't think we should disable the capability if that's what you meant @jeffmaury, but maybe entice the user to choose another model or another model providers

slemeur:
Ok I see this issue: https://github.com/kortex-hub/kortex/issues/457

extensions/ollama/src/extension.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import { activate, deactivate } from './extension';
+import { OllamaExtension } from './ollama-extension';
+
+let extensionContextMock: ExtensionContext;
+
+vi.mock(import('./ollama-extension'));
+
+beforeEach(() => {
+  vi.restoreAllMocks();
+  vi.resetAllMocks();
```
axel7083, line 31: remark (non-blocking): I think the `vi.clearAllMock()` is doing the work of those two

benoitf:
I've rebased the branch but being on hold until we have a way to select tools in MCP

benoitf:
rebased again

PR #451, opened 2025-09-24
feat(chat): remember config from last message

simonrey1:


packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -21,17 +24,31 @@ import NoModelsAvailable from './NoModelsAvailable.svelte';
...
+  messages: DbMessage[];
   readonly: boolean;
 } = $props();
 
 let models: Array<ModelInfo> = $derived(getModels($providerInfos));
 
-let selectedModel = $derived<ModelInfo | undefined>(models[0]);
-let selectedMCP = $state<MCPRemoteServerInfo[]>([]);
+// Partial because existing messages in database have been set an empty config
+const config = messages[messages.length - 1]?.config as Partial<MessageConfig> | undefined;
```
benoitf: issue: we should not do any casting (if not in tests), it breaks all typechecking detection

simonrey1: I agree to limit casting as much as possible but in this case the database helpers stores JSON and I don't know what to do, maybe a zod validation?

benoitf: yes you would need zod, also zod can provide you the type

packages/main/src/chat/chat-manager.ts
```diff
@@ -62,7 +62,7 @@ export class ChatManager {
     if (!existsSync(directory)) {
       await mkdir(directory, { recursive: true });
     }
-    const sqlite = new Database(join(directory, 'chat.db'));
+    const sqlite = new Database(join(directory, 'chats.db'));
```
benoitf, line 65: delete the chat.db if it exists ?

packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -1,16 +1,19 @@
...
 import { ChatHistory } from '/@/lib/chat/hooks/chat-history.svelte';
+import { convertToUIMessages } from '/@/lib/chat/utils/chat';
 import { getModels } from '/@/lib/models/models-utils';
+import { mcpRemoteServerInfos } from '/@/stores/mcp-remote-servers';
 import { providerInfos } from '/@/stores/providers';
+import { MessageConfigSchema } from '/@api/chat/message-config';
 import type { MCPRemoteServerInfo } from '/@api/mcp/mcp-server-info';
 
-import type { Chat as DbChat } from '../../../../../main/src/chat/db/schema';
+import type { Chat as DbChat, Message as DbMessage } from '../../../../../main/src/chat/db/schema';
```
benoitf, line 16: I guess we should move it to api package at some point ? (b/c here renderer uses main stuff which should be prohibited)

benoitf (approved):

packages/main/src/chat/chat-manager.ts
simonrey1: it is very small, I think we can keep it

packages/renderer/src/lib/chat/components/chat.svelte
simonrey1: I agree, I can do a proposition in a follow up PR, thanks for the approval.

simonrey1: https://github.com/kortex-hub/kortex/pull/467

PR #463, opened 2025-09-25
chore(test): initial playwright config

### LINKED ISSUE #448: Initial e2e test playwright integration and basic smoke test

serbangeorge-m:
Closes #448


tests/playwright/playwright.config.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const config: PlaywrightTestConfig = {
+  testDir: './src',
+  timeout: 60_000,
+
+  fullyParallel: false,
+
+  forbidOnly: !!process.env.CI,
+
+  retries: process.env.CI ? 2 : 0,
```
odockal: this is not needed. At least yet.

tests/playwright/playwright.config.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  workers: 1,
+
+  reporter: [
+    ['html', { outputFolder: './output/html-results' }],
```
odockal: I would keep the same folder structure as in pd. `output/tests`

tests/playwright/playwright.config.ts
```diff
@@ -0,0 +1,56 @@
...
+  },
+
+  projects: [
+    {
+      name: 'Electron',
+      testMatch: '**/*.spec.ts',
+    },
+  ],
+
+  outputDir: './output/test-results',
```
odockal: same as above.

tests/playwright/playwright.config.ts
odockal: applies to other test artifacts as well.

tests/playwright/src/fixtures/electron-app.ts
```diff
@@ -0,0 +1,61 @@
...
+    await use(electronApp);
+
+    await electronApp.close();
+  },
+
+  page: async ({ electronApp }, use) => {
+    await new Promise(resolve => setTimeout(resolve, 3000));
+    const page = await electronApp.firstWindow();
+    await page.waitForLoadState('domcontentloaded');
+    await page.setViewportSize({ width: 1280, height: 720 });
```
odockal: we should take what is defined as default in the app core. Not override it here.

tests/playwright/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "0.0.1-next",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "build": "tsc",
+    "test": "playwright test",
+    "test:debug": "playwright test --debug",
+    "test:headed": "playwright test --headed",
+    "test:ui": "playwright test --ui",
+    "report": "playwright show-report ./output/html-report"
```
odockal: Is it used at all?

tests/playwright/package.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "devDependencies": {
+    "@playwright/test": "1.48.1",
```
odockal: You have the same dep in parent package.json. It should be aligned (and using latest version). The best would be to have it defined just once, here. BUT, you need to install playwright deps (chromium) manually or as part of postinstall in subpackage.

tests/playwright/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@kortex/tests-playwright",
```
odockal, line 2: If you have a subpackage defined under the root project, you need to define it in pnpm-workspace.yaml as well. that way you will install deps in subpackage when `pnpm install` is called.

tests/playwright/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "build": "tsc",
+    "test": "playwright test",
+    "test:debug": "playwright test --debug",
+    "test:headed": "playwright test --headed",
+    "test:ui": "playwright test --ui",
+    "report": "playwright show-report ./output/html-report"
+  },
+  "devDependencies": {
+    "@playwright/test": "1.48.1",
+    "@podman-desktop/tests-playwright": "^1.21.0",
```
odockal: this is not used.

odockal (changes requested):

tests/playwright/package.json
serbangeorge-m: yes and no, I would keep it. Might be useful for quick test iteration with debugger

benoitf (commented):
could you please enable them in the pr-check workflow so we'll see if it passes tests

benoitf:
https://github.com/kortex-hub/kortex/blob/main/.github/workflows/pr-check.yaml

.github/workflows/pr-check.yaml
```diff
@@ -271,6 +271,67 @@ jobs:
...
+        mode: [production, development]
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Install Podman v5 using external action
+        # Use the action from the other repository
+        uses: redhat-actions/podman-install@16601a3a718acf7d6986140459092a2f5b941a03
+
+      - name: Setup testenv using external action
+        uses: podman-desktop/e2e/.github/actions/pde2e-testenv-prepare@13e2c57c759137bfc1f437f221967461e8a98e2a
```
benoitf: not sure why we would install podman 5 ?

serbangeorge-m: just followed the example from podman-desktop

benoitf: ok but we don't need podman for kortex

serbangeorge-m: true, I will remove it

PR #466, opened 2025-09-25
fix: support Goose recipes with model settings

### LINKED ISSUE #316: existing goose recipes may not start
existing recipes may not have the goose_provider field, thus it's failing as it's trying to read that field

jeffmaury:
This allows execution of flow without model settings if Goose is configured

Fixes #316


simonrey1 (approved):
Tested on Mac,

it fallsback on Goose configuration for local run, as intended.

Here my API key configured in Goose has expired but this is working as expected.

Image


For Kubernetes, I get an error message as expected too.

Maybe it could be improved because here the provider is not configured.

Image

PR #482, opened 2025-09-30
feat: add RAG to extension API

### LINKED ISSUE #481: Update extension API to support RAG providers

jeffmaury:
Only backend implementation to collect feedback

Fixes #481


jeffmaury:
> I don't see how it's used on the backend side ?
> 
> Questions
> 
>     1. do the backend needs to know something about RAG ?
>        1.can an extension provides RAG and then registers only a MCP tool to get access to the data ? (so no need of this API)

General idea is that RAG environments are managed (started/stopped) through the RAGConnection extension point. Once a RAGConnection is started, the backend can use the MCP Server info to connect to the env (like socket for container connections) and expose it to the chat interface if the RAG environment has been selected

PR #483, opened 2025-09-30
refactor: MCP Manager & MCP Registry (WIP)

axel7083:
## Description

A lot of problem with the way we deals with MCP Registries

1. We load the full catalog in memory and keep a copy in a store
2. The `MCPManager` and `MCPRegistries` are doing too much things, and this is very hard to add new feature
3. We automatically on startup reconnect all remote MCP server, but this is an issue with package, we want to give the user the control of stopping and starting the connections / process

## Screenshots

https://github.com/user-attachments/assets/851df786-d2ae-422e-86b7-2452ebe009df

- MCP Registry selection

> For now, we should only be able to explore one registry at the time

image

- MCP Registry pagination

> Instead of loading the full registry in memory, we only load 30 items at the time, and use the next cursor

image

- Better MCP Configured table

image

1. The MCP status => `running` or `stopped`
2. The MCP Registry this MCP has been taken from
3. The actions => can be `start`, `stop` or `unregister`

image

- mcp-selector disable config when status is stopped

> We could in the future automatically start, but for now, no

image

## Related issues

Requires
- https://github.com/kortex-hub/mcp-runner/pull/11


benoitf (commented):
I still don't get why there are so many changes. It's changing too many things at different places.

For example pagination was not requested (or could go before or after), etc.

benoitf:
outdated, closing

PR #489, opened 2025-10-01
fix(mcp-types): update mcp types, use server name as id without registry

simonrey1:
- Update to new spec of mcp types 1.2.3 to make it work with the registry https://registry.modelcontextprotocol.io

### DIFF
--- a/extensions/mcp-registries/src/extension.ts
+++ b/extensions/mcp-registries/src/extension.ts
@@ -38,7 +38,7 @@ export function deactivate(): void {
 const defaultRegistries: extensionApi.RegistrySuggestedProvider[] = [
   {
     name: 'MCP Registry example',
-    url: 'https://kortex-hub.github.io/mcp-registry-online-v1.2.3',
+    url: 'https://kortex-hub.github.io/mcp-registry-online/v1.2.3/',
     icon: quayIoImage,
   },
 ];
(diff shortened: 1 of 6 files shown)


extensions/mcp-registries/src/extension.ts
```diff
@@ -38,7 +38,7 @@ export function deactivate(): void {
 const defaultRegistries: extensionApi.RegistrySuggestedProvider[] = [
   {
     name: 'MCP Registry example',
-    url: 'https://kortex-hub.github.io/mcp-registry-online-v1.1.0',
+    url: 'https://kortex-hub.github.io/mcp-registry-online-v1.2.3',
```
simonrey1: This does not exist yet, so on this PR the console shows an error and only the other registries that match version 1.2.3 are listed

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -129,23 +129,8 @@ export class MCPRegistry {
...
-        serverId = official.serverId;
-      }
-    }
-    if (!serverId) {
-      const rawId = `${registryURL}::${server.name}`;
-      serverId = crypto.createHash('sha256').update(rawId).digest('hex');
-    }
-    return { ...server, serverId };
+  enhanceServerDetail(server: components['schemas']['ServerDetail']): MCPServerDetail {
+    return { ...server, serverId: encodeURI(server.name) };
```
axel7083, line 133: 👍

extensions/mcp-registries/src/extension.ts
simonrey1: https://github.com/kortex-hub/mcp-registry-online/pull/2/files

benoitf (approved):
LGTM

simonrey1:
@benoitf thanks for the approval, can you force the merge after the build finishes because macOS will not pass?

PR #495, opened 2025-10-02
refactor(mcp-manager): extract exchanges logic to dedicated class

axel7083:
## Description

Since https://github.com/kortex-hub/mcp-runner exports its own MCPManager dealing with storage and logic, we should clean the existing MCPManager, the first step is to extract the exchanges logic (the MCPTransportDelegate)

## Related issues

Split of https://github.com/kortex-hub/kortex/pull/483


benoitf (approved):

jeffmaury (approved):
LGTM

PR #502, opened 2025-10-06
chore: changed export icon to flow icon

gastoner:
Before: 

image

After: 

image

image

Closes https://github.com/kortex-hub/kortex/issues/627

### DIFF
--- a/packages/renderer/src/lib/chat/components/ExportButton.svelte
+++ b/packages/renderer/src/lib/chat/components/ExportButton.svelte
@@ -8,7 +8,7 @@ import { isFlowConnectionAvailable } from '/@/stores/flow-provider';
 import type { MCPRemoteServerInfo } from '/@api/mcp/mcp-server-info';
 import { NavigationPage } from '/@api/navigation-page';
 
-import ExportIcon from './messages/ExportIcon.svelte';
+import FlowIcon from '../../images/FlowIcon.svelte';
 import type { ModelInfo } from './model-info';
 import { Button } from './ui/button';
 
@@ -25,6 +25,8 @@ let {
 } = $props();
 
 let loadingExportAsFlow = $state(false);
+
+let title = $derived($isFlowConnectionAvailable ? 'Export as Flow' : 'Install flow provider to enable save.');
 
 const exportAsFlow = async (): Promise<void> => {
   if (!selectedModel) {
@@ -60,14 +62,15 @@ const exportAsFlow = async (): Promise<void> => {
 </script>
 
 <Button
-	class="h-fit rounded-md p-[7px] hover:bg-zinc-200 dark:border-zinc-700 hover:dark:bg-zinc-900"
+	class="h-fit rounded-2xl p-[5px] mr-1 border"
 	onclick={async(event): Promise<void> => {
 		event.preventDefault();
 		await exportAsFlow();
 	}}
 	disabled={loading || !chatClient.messages.length || loadingExportAsFlow}
-	variant="ghost"
-	title={$isFlowConnectionAvailable? 'Export as Flow' : 'Install flow provider to enable save.'}
+	variant="outline"
+	{title}
 >
-	<ExportIcon size="2x"/>
+  <FlowIcon size={2}/>
+  {title}
 </Button>
--- a/packages/renderer/src/lib/chat/components/multimodal-input.svelte
+++ b/packages/renderer/src/lib/chat/components/multimodal-input.svelte
@@ -174,6 +174,7 @@ $effect.pre(() => {
 	</div>
 
 	<div class="absolute right-0 bottom-0 flex w-fit flex-row items-center justify-end p-2">
+		<ExportButton {chatClient} {selectedModel} {selectedMCP} {loading}/>
 		{#if loading}
 			<Button
 				class="h-fit rounded-full border p-1.5 dark:border-zinc-600"
@@ -198,7 +199,5 @@ $effect.pre(() => {
 				<ArrowUpIcon size={14} />
 			</Button>
 		{/if}
-
-		<ExportButton {chatClient} {selectedModel} {selectedMCP} {loading}/>
 	</div>
 </div>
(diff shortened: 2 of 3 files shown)


benoitf (commented):
not sure here it should be the flow icon as it's an export first

slemeur:
I agree with you @gastoner, I would go with option2. 
I think it would be better if the button was a bit more visible. (maybe not as blue, but most probably in gray to be consistent)
Screenshot 2025-10-20 at 17 37 26

gastoner:
> The logic I have in mind is that:
> 
>     * the buttons in the textarea (or interaction area) are to directly interact with the LLM (and its connected other assets) and continue expanding in the conversation. The user will stay in that context, no navigation to a new page.
> 
>     * The "export as a flow" is not expanding the conversation anymore, it is wrapping it up into a flow by summarizing the entire conversation and bringing the user into a new page of the application.
> 
> 
> That's why to me, it was better if outside. But happy to have other opinions.

That would make sense, but it looks really odd as Florent wrote

I have 2 ideas: 
1. Since the send button is disabled half of the time, what about the buttons would have been switching? When there is no text => export, when is user typing => send button. But there might be a problem e.g. when user types something but wants to export the conversation as flow (needs to remove all of the text) - so not ideal I guess 

2. Move it to top ? e.g. to top right, on top left I feel like there is a lot of options more related to chat options/ etc.. 
Having it on top right would make the button to pop-out more + or looks kinda like https://fontawesome.com/icons/arrow-up-right-from-square?s=solid thanks to its position

slemeur:
Thanks for sharing your thoughts @gastoner .
I think the option 1. with alternate button will not be clear to the end user. 

The proposal you have in 2. is also moving the "Export as Flow" outside of the "interaction area". And it would also get it back to where it was initially. If it stays prominent I'm fine with that approach. 

Why am I thinking it needs to be prominent/visible enough?
In many chat tools, you don't have this capability of "exporting as a flow", that's unique and we will need to educate the users about the capability, but it has to trigger curiosity.

MariaLeonova:
I still think that having a button outside the chat area makes more sense from the user flow POV (not to confuse with the Kortex flow) - it is not essentially the part of the chat, it is a **separate action** that would take the whole conversation and convert it to flow. In that sense what Stevan has shown [here](https://github.com/kortex-hub/kortex/pull/502#issuecomment-3422631185) is more logical. 

For example, if you look at Gemini action buttons inside the chat window, they are all related to actions you can take related to the chat messages and asks: 
<br></br>
Screenshot 2025-10-21 at 10 45 03
<br></br>

Or the ones in ChatGPT:
Screenshot 2025-10-21 at 10 45 30
<br></br>

Same goes for Cursor:
Screenshot 2025-10-21 at 10 46 21

Those are all related to what you are typing it right now and will affect model output.

In case of `Export as flow`, that action is related to your **whole experience** and will not have effect on chat utput; it is a separate action at that.
Having said that, I have a preference for Option 3 in @gastoner 's [proposal](https://github.com/kortex-hub/kortex/pull/502#issuecomment-3422256142).

As it comes last on the page, it logically sums up all the actions and the conversation a user has taken there. 

If that is confusing to people and takes too much space, maybe we could consider an alternative placement on top with all the other 'general' buttons?
Screenshot 2025-10-21 at 10 51 44

simonrey1:
> @axel7083 @slemeur @benoitf @simonrey1 WDYT about the placement?

I prefer the header button action

benoitf:
here are my comments

about the header I've already opened https://github.com/kortex-hub/kortex/issues/582
Alternating actions and selector is disturbing to me (like first item = select something for the chat, second create a new chat (change the page), 3rd one select something for the chat, 4th would be another action that open a new page)

actions on my end should not be in the middle of selectors

IMHO `new chat` should stay at the top (header) or even more disappear ! and stay only in the sidebar but selectors should go in the input area (for example selecting the MCP server /tools and the model should be to me inside the input/interactive area) not at the top of my screen. because it's part of what I'll enter in the input area.

so to me, why export should stay at the header ? my eyes are focused on the interactive chat where I enter data. so to me it can't go at the top but stay around the chat input

slemeur:
@benoitf I agree with your point, but I'm not sure about the conclusion.

While I do agree that your eyes will be focused on the "prompt area", I believe that part of the screen would be more dedicated for the interactive chat, so having actions directly reflected on this screen. Which is not the case for the "export as a flow". 

There is probably a combination of choices we could do.
1/ Maybe we could think about something contextual. After succesfull interactions, we could display the "export as a flow" button (and maybe other actions... I'm thinking about the way it is in a PR review on github... we have all the comments, and then we have the "actions to be taken on the PR" and the "Add a comment" section). There might be something to explore on that area. 

Screenshot 2025-10-23 at 19 27 07

2/ Detecting pattern? showing up the action once we detect a reusable pattern? 

3/ We could also have a kebab menu on the list of chat on the chat sessions where one of the action would be export as a flow.
For 1/ and 2/ it could be like a card at the bottom of the conversation. 


Thoughts on these ideas? 

Now for the sake of getting to a conclusion for the PR and unblocking @gastoner, let's continue the work with option 2. from here: https://github.com/kortex-hub/kortex/pull/502#issuecomment-3422256142 

I have created https://github.com/kortex-hub/kortex/issues/607 to really look at the position of that action and how we can display it in a clear way but there will be other tickets to consider like this: https://github.com/kortex-hub/kortex/issues/582 

Thanks everybody for the comments.

simonrey1 (approved):
LGTM, hover text is hard to read but non blocking

Image


✅  Icon has been updated in flow create 

Image

simonrey1:
> Is that unreadable text (/a11y) in the scope of this task, or worth another ticket?

The text was added in this PR, before there was only the icon, so I guess it is in the scope of this PR

gastoner:
@simonrey1 @vancura updated bg of the button (should be same as buttons on top)

vancura:
Looks great! Pixel perfect! Thanks!

Issue #488, opened 2025-10-01
Flow: Automatically suggest name and description for the flow


slemeur:
## Description

Based on the prompt, it should be possible to generate a meaningfull description and name for the flow.

This would help the user into better identifying the flows created?

Issue #455, opened 2025-09-25
EPIC: Input fields (parameters) for Flows


jeffmaury:
As a AI Builder, my agents are likely going to be templatize so they could be applied to different context with different input values.  

## Pre-requisites  
\--

### Adding Parameters to the flow

* When creating a flow, there will be a new section in the form for “Input Fields”.   
  * An input field is defined by:  
    * A description: which explain what the input field is for  
    * An input “name”: which would be the variable we use in the flow  
    * A default value:an optional default value for the input field  
    * A value type: Text, Number, Boolean, Enum.   
    * A required value: indicating is the input is mandatory for the flow or not  
  * This section should be displaying a table with the description of the input field, the default value and show an icon for input type  
  * There should be a way to manage the input fields (add remove edit)  
  * When clicking on the add (or edit), there will be a modal displayed to provide all the information about the input field  
* Once defined in the flow settings, it should be possible to use the Input Fields in the prompt  
  * In this case, the user will have to put under brackets the input field name:  
    * For Instance: “Take the last 5 issues from {{repository_url}}”   
      

### Executing a flow with Input Fields

* Now that a flow can have input fields, the user who is executing the flow will have to provide them.   
* To begin with (and in the interest of keeping things relatively simple at the moment), when the user will be starting a flow, the user will see a modal with a way to provide the values for the flow.   
  * Only when the “required” input fields will be filled, the user will be able to start the flow.

    

### Impact of deployments of a flow with Input Fields

* In order to make things a simple to start with, a flow will always be deployed with a specific set of values. That means that the user will have to provide the values for each of the required fields.  
* We will handle the flow in a similar fashion that what we are doing to “execute” the flow locally. 

## Corner Cases

- Removing an input field, when used in the prompt

## Stretch

- Provide completion in the “prompt editor” so when the user start with ‘{{‘ it proposes the different input fields that are defined  
- Provide the possibility to select text in the prompt to create an input field

PR #503, opened 2025-10-06
fix(chat): make prompt generation extensible to further add description, title and parameters

simonrey1:
In this PR, I keep the same instructions to generate the prompt, but instead of using raw text output, I specify that I want the LLM to return an object. In this PR, it has only one property but we need to add more properties (title, description, parameters...) for next coming issues.

Pre-requisite for https://github.com/kortex-hub/kortex/issues/488 and https://github.com/kortex-hub/kortex/issues/455
 

Screenshot 2025-10-06 at 12 27 34
Screenshot 2025-10-06 at 12 27 45
Screenshot 2025-10-06 at 12 28 31


benoitf (approved):
will fix the english of the query to

```
Help me create a reproducible prompt that achieves the same result as in the conversation above. The prompt will be executed by another LLM without any further user input, so it must include all the necessary information to reproduce the same outcome.
```

PR #506, opened 2025-10-06
feat(chat/flow): add generation of title and description, move type to API

simonrey1:
In this PR, I move the zod schemas to the API package to be reused in types around the application.

And I added the name and description of the flow to be generated.

I also applied @benoitf suggestion to change to fix the formulation of the prompt.

Example:

Screenshot 2025-10-06 at 15 13 54

### DIFF
--- a/packages/api/src/chat/flow-generation-parameters-schema.ts
+++ b/packages/api/src/chat/flow-generation-parameters-schema.ts
@@ -13,7 +13,7 @@ export const FlowGenerationParametersSchema = z.object({
   prompt: z
     .string()
     .describe(
-      'Help me to build a reproducible prompt to achieve the same result as I got in the conversation above. The prompt will be executed by another LLM without any further user input so it must contain all the information on how to get the same result.',
+      'Help me create a reproducible prompt that achieves the same result as in the conversation above. The prompt will be executed by another LLM without any further user input, so it must include all the necessary information to reproduce the same outcome.',
     ),
 });
 
--- a/packages/api/src/chat/flow-generation-parameters-schema.ts
+++ b/packages/api/src/chat/flow-generation-parameters-schema.ts
@@ -1,0 +1,20 @@
+import z from 'zod';
+
+export const kubernetesNameRegex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/;
+
+export const FlowGenerationParametersSchema = z.object({
+  name: z
+    .string()
+    .regex(kubernetesNameRegex)
+    .describe(
+      `Name of the flow that will be saved. The name must match the kubernetes name regex /${kubernetesNameRegex.source}/.`,
+    ),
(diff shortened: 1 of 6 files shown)


packages/api/src/chat/flow-generation-parameters-schema.ts
```diff
@@ -0,0 +1,20 @@
+import z from 'zod';
```
benoitf, line 19: missing copyrights/headers

packages/api/src/chat/flow-generation-parameters-schema.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const kubernetesNameRegex = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/;
```
benoitf: there is also a maximum length no ?

benoitf (approved):
couple of issues to address before merging

packages/api/src/chat/flow-generation-parameters-schema.ts
simonrey1: It seems to be 253 characters, I will enforce it in the zod schema and use the schema for validation in the form.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

benoitf: if I recall correctly it's 63 characters

packages/api/src/chat/flow-generation-parameters-schema.ts
```diff
@@ -0,0 +1,46 @@
...
+    .max(nameMaxLength)
+    .transform(val =>
+      val
+        .trim()
+        .toLowerCase()
+        .replace(/\s+/g, '-') // Replace spaces with hyphens
+        .replace(/[^a-z0-9-]/g, '') // Remove all other invalid characters
+        .slice(0, nameMaxLength),
+    )
+    .pipe(z.string().regex(/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/, 'Invalid DNS subdomain name after transformation.')),
```
benoitf: should the length be part of the regexp ?

simonrey1: Reading this SO answer, I think it is easier do use the ".max(53)" method on zod string type

"But please note that it won't check for the max length of 255. It may be done using a positive lookahead, but the regex will become very complicated, and it is shorter to be using a string length function ;)"

https://stackoverflow.com/questions/8848981/looking-to-build-some-regex-to-validate-domain-names-rfc-952-rfc-1123

benoitf: but AFAIK stackoverflow is not the ultimate answer ?

`/^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$/,` still look readable

slemeur:
pretty nice!

Issue #403, opened 2025-09-19
detect ramalama/ai-lab running models


benoitf:
local models served by ramalama/ai-labs should be detected and registered

PR #507, opened 2025-10-06
feat: add RamaLama extension

benoitf:
- [ ] depends on https://github.com/kortex-hub/kortex/pull/505

image

register models served by RamaLama

note: for now it is not implemented for Linux

fixes https://github.com/kortex-hub/kortex/issues/403


extensions/ramalama/src/helper/container/container-ramalama-finder.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ContainerRamaLamaFinder } from './container-ramalama-finder';
+
+beforeEach(async () => {
+  vi.resetAllMocks();
+});
+
+test('getRamaLamaContainers filters containers with ai.ramalama label', async () => {
+  const containerRamaLamaFinder = new ContainerRamaLamaFinder();
+
+  const containers = [
```
simonrey1, line 31: suggestion:

```ts

  const getBaseInfos = (): Omit<ContainerInfo, 'Id' | 'Names'> => ({
    Image: '',
    ImageID: '',
    Command: '',
    Created: 0,
    Ports: [],
    State: '',
    Status: '',
    HostConfig: {
      NetworkMode: '',
    },
    NetworkSettings: {
      Networks: {},
    },
    Labels: {},
    Mounts: [],
  });
  const containers: ContainerInfo[] = [
    { ...getBaseInfos(), Id: '1', Names: ['/container1'], Labels: { 'ai.ramalama': 'true' } },
    { ...getBaseInfos(), Id: '2', Names: ['/container2'] },
    { ...getBaseInfos(), Id: '3', Names: ['/container3'], Labels: { 'ai.ramalama': 'true' } },
    { ...getBaseInfos(), Id: '4', Names: ['/container4'] }, // No labels
  ];
```

extensions/ramalama/src/helper/container/container-ramalama-finder.ts
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ContainerInfo } from 'dockerode';
+
+import type { RamalamaContainerInfo } from '../../api/ramalama-container-info';
```
simonrey1, line 21: suggestion: 
```ts
import type { RamalamaContainerInfo } from '/@/api/ramalama-container-info';
```

extensions/ramalama/src/helper/socket-finder/docker/docker-macos-finder.ts
simonrey1, line 1: suggestion: move these helpers outside of ramallama extension

extensions/ramalama/src/api/connection-event.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+// use zod for the json event received by docker/podman
+
+import z from 'zod';
+
+export const DockerEventSchema = z.object({
+  status: z.string().optional(),
```
simonrey1, line 24: In https://docs.docker.com/reference/cli/docker/system/events/#format-the-output, it seems that both Status and Type start with an uppercase?

simonrey1: suggestion: status seems unused, maybe remove it?

benoitf: if you look at your link but at the bottom (where it says what we have using json) you'll see that status has no uppercase

extensions/ramalama/src/handler/conection-handler.ts
```diff
@@ -0,0 +1,116 @@
+/**********************************************************************
```
jeffmaury, line 1: Should be renamed as connection-handler.ts

benoitf:
will need to be able to pickup a given tool in MCP before review

also need to split in two

PR #523, opened 2025-10-07
chore: enforce extensions are using absolute imports (no relative)

benoitf:
see https://github.com/kortex-hub/kortex/pull/507#pullrequestreview-3308912786

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -291,6 +291,28 @@ export default [
   },
 
   {
+    files: ['extensions/**'],
+    languageOptions: {
+      globals: {
+        ...globals.node,
+      },
+    },
+    rules: {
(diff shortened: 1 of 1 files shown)


axel7083:
> could you elaborate why you're against an automatic check that is enforcing a pattern used by all the current extensions in this repository ?

No, this is is not the point I'm making.

> I would like of course to introduce it in Podman Desktop repository, just that it'll take time to update the existing codebase

We share the same objective, as I share the view of wanting to reduce the complexity of imports. However, making such change in the Kortex project (a fork of Podman Desktop) should be done with caution, and creating an upstream issue is clearly required here to ensure that we don't differ too much in the code guidelines and have an unified developer experience.

Issue #524, opened 2025-10-07
Chat allows sending whitespace-only messages


serbangeorge-m:
The chat input field incorrectly enables the "Send" button when I enter only whitespace characters.

<img width="826" height="288" alt="Image" src="https://github.com/user-attachments/assets/aeb0b739-ca98-4dfe-a454-61a0d58ffbed" />

<img width="854" height="1003" alt="Image" src="https://github.com/user-attachments/assets/25eea89c-c494-44a8-959d-1a4593547438" />

PR #528, opened 2025-10-07
fix: trim whitespaces when checking to enable or not

benoitf:
fixes https://github.com/kortex-hub/kortex/issues/524


jeffmaury (approved):
LGTM

Issue #558, opened 2025-10-14
Cannot navigate to MCP_INSTALL_FROM_REGISTRY page with serverId from `registry.modelcontextprotocol.io`


axel7083:
## Description

The following code is responsible of navigating to the MCP Setup page

https://github.com/kortex-hub/kortex/blob/3f522ea63c15de8bf3e42bdf6b905f06e782ede3/packages/renderer/src/navigation.ts#L63

But there is a flaw, the `serverId` is a string that may contain `/` leading to invalid route.

Example, the official serverId for GitHub in the registry is `{serverId: 'io.github.github/github-mcp-server'}`

PR #559, opened 2025-10-14
fix(renderer): invalid route path with modelcontextprotocol registry

axel7083:
Fix https://github.com/kortex-hub/kortex/issues/558

### DIFF
--- a/packages/renderer/src/App.svelte
+++ b/packages/renderer/src/App.svelte
@@ -191,7 +191,7 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
           <MCPDetails id={meta.params.id} />
         </Route>
         <Route path="/mcp-install-from-registry/:serverId/*" breadcrumb="Install MCP Server from Registry" let:meta>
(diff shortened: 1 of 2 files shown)


packages/renderer/src/App.svelte
```diff
@@ -191,7 +191,7 @@ window.events?.receive('kubernetes-navigation', (args: unknown) => {
           <MCPDetails id={meta.params.id} />
         </Route>
         <Route path="/mcp-install-from-registry/:serverId/*" breadcrumb="Install MCP Server from Registry" let:meta>
-          <McpRegistryCreateFromRegistryForm serverId={meta.params.serverId} />
+          <McpRegistryCreateFromRegistryForm serverId={atob(meta.params.serverId)} />
```
simonrey1: question: have you tried to encode the URL?

benoitf: existing routes I think are using encodeURIComponent / decodeURIComponent pattern

benoitf: and older decodeURI/encodeURI

Issue #563, opened 2025-10-14
MCP Form do not support filepath as input type


axel7083:
## Description

Currently only string is supported and is used as fallback

https://github.com/kortex-hub/kortex/blob/90d383858992177984692fd1aca9b94fe3642860/packages/renderer/src/lib/mcp/setup/InputArgument.svelte#L37-L49

PR #564, opened 2025-10-14
feat(renderer: MCP Form): adding support for the filepath input format

axel7083:
## Description

Adding the support for `filepath` input format.

## Screenshots

image

## Related issues

Required for https://github.com/kortex-hub/kortex/issues/322
Fixes https://github.com/kortex-hub/kortex/issues/563

### DIFF
--- a/packages/renderer/src/lib/mcp/setup/InputArgument.spec.ts
+++ b/packages/renderer/src/lib/mcp/setup/InputArgument.spec.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (C) 2023 Red Hat, Inc.
+ * Copyright (C) 2025 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


packages/renderer/src/lib/mcp/setup/InputArgument.spec.ts
```diff
@@ -0,0 +1,88 @@
+/**********************************************************************
+ * Copyright (C) 2023 Red Hat, Inc.
```
benoitf:
```suggestion
 * Copyright (C) 2025 Red Hat, Inc.
```

axel7083: Good catch 👍

benoitf: I hope we could find an eslint rule (not a slow one and a working one to validate this (and auto-fix))

axel7083: This one https://github.com/nikku/eslint-plugin-license-header does not seems to support several dates :/ like `2022-2025`

benoitf: I think we had one that we dropped after the switch to eslint 9 but AFAIK it was not checking date, only mainly that we had a copyright header in the file

PR #567, opened 2025-10-15
feat(main): support local MCP with npx

axel7083:
## Description

Work in pair with https://github.com/kortex-hub/kortex/pull/436

## Screenshots

https://github.com/user-attachments/assets/5f1b5c30-9c5e-40b7-a158-85b4b1928631

## Related issues

Backend part of https://github.com/kortex-hub/kortex/issues/322

Require rebase after:
- https://github.com/kortex-hub/kortex/pull/589

## Steps to reproduce

1. checkout this branch
2. merge `feature/mcp-registry/package-form` in this branch locally
3. add the official model context registry to the list of registries
4. search for `dev.svelte/mcp`
> You can update the suggested action from github mcp to `dev.svelte/mcp` to directly open the correct MCP
5. select the package (local) version
6. click on `spawn`
7. go to the chat
8. ask something like `Tell me about bindable`
9. assert the LLM is using the tools


packages/main/src/plugin/mcp/package/npm-spawner.ts
```diff
@@ -0,0 +1,72 @@
...
+        ...(this.pack.runtimeArguments ?? []),
+        // let's use package@version if version is specified
+        this.pack.version ? `${this.pack.identifier}@${this.pack.version}` : this.pack.identifier,
+        ...(this.pack.packageArguments ?? []),
+      ],
+      env: this.pack.environmentVariables,
+    });
+    this.#disposables.push({
+      [Symbol.asyncDispose]: () => {
+        return transport.close();
```
simonrey1, line 48: suggestion(non-blocking): create a new class `StdioClientTransportDisposable` that implements `AsyncDisposable` so it can be added directly to the list of disposables.

```ts
const transport = new StdioClientTransportDisposable({

[...]

this.#disposables.push(transport);

```

This could help future user of this class not to forget to close the transport.

packages/main/src/plugin/mcp/package/npm-spawner.ts
```diff
@@ -0,0 +1,72 @@
...
+      },
+    });
+    return transport;
+  }
+
+  async enabled(): Promise<boolean> {
+    const { promise, resolve } = Promise.withResolvers<boolean>();
+
+    // eslint-disable-next-line sonarjs/os-command
+    exec(`${NPX_COMMAND} --version`, error => {
```
simonrey1: issue (blocking): The npx command is expensive and it seems that the enabled method is called for every `packages/main/src/plugin/mcp/mcp-registry.ts:219` server (could be thousands) of the registry

packages/main/src/plugin/mcp/package/mcp-spawner.ts
```diff
@@ -0,0 +1,36 @@
...
+  runtimeArguments?: Array<string>;
+  packageArguments?: Array<string>;
+  environmentVariables?: Record<string, string>;
+};
+
+export abstract class MCPSpawner<T extends string = string> implements AsyncDisposable {
+  constructor(protected readonly pack: ResolvedServerPackage & { registryType: T }) {}
+
+  abstract spawn(): Promise<Transport>;
+  abstract [Symbol.asyncDispose](): PromiseLike<void>;
```
benoitf:
```suggestion
  abstract [Symbol.asyncDispose](): Promise<void>;
```

packages/main/src/plugin/mcp/package/mcp-package.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  spawn(): Promise<Transport> {
+    return this.#spawner.spawn();
+  }
+
+  enabled(): Promise<boolean> {
+    return this.#spawner.enabled();
+  }
+
+  [Symbol.asyncDispose](): PromiseLike<void> {
```
benoitf: return type is Promise usually (not PromiseLike in our source code)

packages/main/src/plugin/mcp/package/npm-spawner.ts
```diff
@@ -0,0 +1,72 @@
...
+      } else {
+        resolve(false);
+      }
+    });
+
+    return promise;
+  }
+
+  async [Symbol.asyncDispose](): Promise<void> {
+    await Promise.allSettled(this.#disposables.map(disposable => disposable[Symbol.asyncDispose]()));
```
benoitf: I have a question on the pattern

I would say that if we use asyncDispose it means the runtime is taking care of calling the dispose method.
If we programmatically call dispose it means it's a custom disposable way.

so, should we use kortex `IDisposable` interface if we call ourselves at the end the method. Else to me there is a design pattern conflict.

packages/main/src/plugin/mcp/package/mcp-package.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import type { MCPSpawner, ResolvedServerPackage } from './mcp-spawner.js';
+import { NPMSpawner } from './npm-spawner.js';
+
+export class MCPPackage implements AsyncDisposable {
+  readonly #spawner: MCPSpawner;
+
+  constructor(pack: ResolvedServerPackage) {
+    // By destructuring `registry_type` and reconstructing the object, TypeScript can properly infer the narrowed type in each switch case.
+    const { registryType, ...rest } = pack;
```
jeffmaury, line 30: Why do you need deconstruction and not use pack.registryType

packages/main/src/plugin/mcp/package/mcp-package.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!registryType) throw new Error('cannot determine how to spawn package: registry_type is missing');
+
+    switch (registryType) {
+      case 'npm':
+        this.#spawner = new NPMSpawner({
+          ...rest,
```
jeffmaury, line 36: Could be pack instead ?

packages/main/src/plugin/mcp/package/npm-spawner.ts
```diff
@@ -0,0 +1,72 @@
...
+  async spawn(): Promise<Transport> {
+    if (!this.pack.identifier) throw new Error('missing identifier in MCP Local Server configuration');
+    if (this.pack.fileSha256) {
+      console.warn('specified file sha256 is not supported with npx spawner');
+    }
+
+    const transport = new StdioClientTransport({
+      command: NPX_COMMAND,
+      args: [
+        ...(this.pack.runtimeArguments ?? []),
```
jeffmaury, line 39: -y missing ?

packages/main/src/plugin/mcp/package/mcp-package.ts
axel7083: Yeah, the problem is commented at L27, the problem is typescript cannot infer nested type.

You cannot use pack.registryType with a switch and have typescript properly infer the property. This is a limitation

packages/main/src/plugin/mcp/package/npm-spawner.ts
axel7083: If we use the `IDisposable` this is not allowing async dispose logic, and as we dispose resources like Transport, which have `close` method with Promise return this feel required.

I don't see a clean way of doing the disposable without using `.catch(console.error)` or having some kind of dandling promises.

Any suggestion?

packages/main/src/plugin/mcp/package/npm-spawner.ts
axel7083: should not be required, I guess it only request the user to valide if there is a shell available, and there this is not.

packages/main/src/plugin/mcp/package/npm-spawner.ts
benoitf: I'm fine to have `IAsyncDisposable/ async dispose : Promise<void>` interface 
then we use 'our' interface when it's about removing objects we manage and let Node.js asyncDispose to dispose objects by the runtime

axel7083: So as an action item, I will create an issue in Podman Desktop repository called `IAsyncDisposable` 👍 then backport to Kortex

axel7083: Upstream issue https://github.com/podman-desktop/podman-desktop/issues/14419

packages/main/src/plugin/mcp/package/npm-spawner.ts
```diff
@@ -0,0 +1,72 @@
...
+        // let's use package@version if version is specified
+        this.pack.version ? `${this.pack.identifier}@${this.pack.version}` : this.pack.identifier,
+        ...(this.pack.packageArguments ?? []),
+      ],
+      env: this.pack.environmentVariables,
+    });
+    this.#disposables.push({
+      [Symbol.asyncDispose]: () => {
```
jeffmaury: Add missing return type

packages/main/src/plugin/mcp/package/npm-spawner.ts
axel7083: It is only called for each server configured, so not for all server in the registry

packages/main/src/plugin/mcp/package/npm-spawner.ts
axel7083: Might be follow-up I like the idea, but probably out of scope for now

packages/main/src/plugin/mcp/package/mcp-package.ts
axel7083: Since we deconstruct the pack, only the registryType is properly infered, so this would lead to a type error.

Not really nice, but I don't see a better solution with current typescript limitation

packages/main/src/plugin/mcp/package/npm-spawner.ts
simonrey1: Sill takes time for each of them for a result that could be easily kept in memory.

`npx --version  0.05s user 0.02s system 21% cpu 0.334 total`

axel7083: I removed the enabled method, as this is out of scope of this PR and was not in the requirement of the issue.

packages/main/src/plugin/mcp/package/mcp-spawner.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { components } from '@kortex-hub/mcp-registry-types';
+import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
+
+import type { IAsyncDisposable } from '/@api/async-disposable.js';
+
+export type ResolvedServerPackage = Omit<
+  components['schemas']['Package'],
+  'packageArguments' | 'runtimeArguments' | 'environmentVariables'
+> & {
+  runtimeArguments?: Array<string>;
```
simonrey1, line 27: nitpick: these 3 properties could be included in an interface as they are reused in mcp-registry

```ts

  runtimeArguments?: Array<string>;
  packageArguments?: Array<string>;
  environmentVariables?: Record<string, string>;
```

axel7083: Due to lack of refactor, those properties are not the same, here they are arrays, but in the mcp registry, they are records

  runtimeArguments: Record<number, InputWithVariableResponse>;
  packageArguments: Record<number, InputWithVariableResponse>;
  environmentVariables: Record<string, InputWithVariableResponse>;

To give some explanation, the Record<number,  InputWithVariableResponse> only contain user values, but the Array one contains ALL the arguments, the logic is the following

the mcp-registry will get the server details, and go through all arguments, for each of them it will check if the user has provided a value for one, if yes will replace in the array.

simonrey1: Sorry, it seems like an issue with the PR preview in Github, my comment is outdated.

PR #573, opened 2025-10-17
feat: added dynamic gemini model loading

### LINKED ISSUE #464: Gemini provider users an harcoded list of models
We should retrieve the list dynamically

gastoner:
Adds dynamic loading of gemini models to list

Uses only latest version and the ones that can generate content

Closes #464

### DIFF
--- a/extensions/gemini/src/gemini.spec.ts
+++ b/extensions/gemini/src/gemini.spec.ts
@@ -17,6 +17,8 @@
  ***********************************************************************/
 
 import { createGoogleGenerativeAI, type GoogleGenerativeAIProvider } from '@ai-sdk/google';
+import type { Model, Pager } from '@google/genai';
+import { GoogleGenAI } from '@google/genai';
 import type {
   CancellationToken,
   Disposable,
@@ -44,6 +46,8 @@ vi.mock(import('@ai-sdk/google'), () => ({
   createGoogleGenerativeAI: vi.fn(),
 }));
 
+vi.mock(import('@google/genai'));
+
 const GOOGLE_AI_PROVIDER_MOCK: GoogleGenerativeAIProvider = {} as unknown as GoogleGenerativeAIProvider;
 
 const PROVIDER_API_MOCK: typeof ProviderAPI = {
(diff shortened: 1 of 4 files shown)


benoitf (commented):
do we need a 3rd party library ?

benoitf:
I was wondering if it should then display only these models

Gemma models for example or other models ?

extensions/gemini/src/gemini.ts
```diff
@@ -150,14 +153,7 @@ export class Gemini implements Disposable {
...
       },
-      models: [
-        {
-          label: 'gemini-2.5-flash',
-        },
-        {
-          label: 'gemini-2.5-pro',
-        },
-      ],
+      models: models,
```
benoitf:
```suggestion
      models,
```

extensions/gemini/src/gemini.ts
```diff
@@ -167,6 +163,19 @@ export class Gemini implements Disposable {
     this.connections.set(tokenHash, connectionDisposable);
   }
 
+  private async getGeminiModels(token: string): Promise<Array<{ label: string }>> {
+    const ai = new GoogleGenAI({ apiKey: token });
+    const geminiModels = await ai.models.list();
+    const models: Array<{ label: string }> = [];
+    for await (const model of geminiModels) {
```
benoitf, line 171: why do we need await there ?

extensions/gemini/src/gemini.ts
```diff
@@ -167,6 +163,19 @@ export class Gemini implements Disposable {
...
+      if (model.version?.includes('Latest') && model.name && model.supportedActions?.includes('generateContent')) {
+        const label = model.name.replace('models/', '');
+        models.push({ label: label });
```
benoitf, line 174: I would opt-in for a filter/map approach rather than adding items to an array

```
  return geminiModels
    .filter(
      model =>
        model.version?.includes('Latest') &&
        model.name &&
        model.supportedActions?.includes('generateContent')
    )
    .map(model => ({
      label: model.name.replace('models/', '')
    }));
```

benoitf (approved):

extensions/gemini/src/gemini.ts
gastoner: Me too, but the list is returning an iterator `Promise<Pager<Model>>` so I cant use the filer/map I think :( @benoitf

benoitf: so you can ignore my comments 👍

extensions/gemini/src/gemini.ts
benoitf: b/c not an array, async iterator

extensions/gemini/src/gemini.spec.ts
```diff
@@ -44,8 +45,22 @@ vi.mock(import('@ai-sdk/google'), () => ({
   createGoogleGenerativeAI: vi.fn(),
 }));
 
+vi.mock('@google/genai', () => ({
```
benoitf: use automocking ?

`vi.mock('@google/genai')` should be enough

suggest: 

```ts
vi.mock(import('@google/genai'));
```

extensions/gemini/src/gemini.spec.ts
```diff
@@ -69,6 +84,38 @@ beforeEach(() => {
 
   vi.mocked(PROVIDER_API_MOCK.createProvider).mockReturnValue(PROVIDER_MOCK as Provider);
   vi.mocked(createGoogleGenerativeAI).mockReturnValue(GOOGLE_AI_PROVIDER_MOCK);
+
+  // Mock GoogleGenAI constructor to return our mock instance
+  vi.mocked(GoogleGenAI).mockImplementation(() => GOOGLE_GENAI_MOCK);
```
benoitf: prefer the usage of prototype rather than mockImplementation

extensions/gemini/src/gemini.spec.ts
```diff
@@ -69,6 +84,38 @@ beforeEach(() => {
...
+
+  // Mock secret storage to return empty (no tokens to restore)
+  vi.mocked(SAFE_STORAGE_MOCK.get).mockResolvedValue(undefined);
```
benoitf: I think using prototype before is cleaner

extensions/gemini/src/gemini.spec.ts
```diff
@@ -69,6 +82,34 @@ beforeEach(() => {
 
   vi.mocked(PROVIDER_API_MOCK.createProvider).mockReturnValue(PROVIDER_MOCK as Provider);
   vi.mocked(createGoogleGenerativeAI).mockReturnValue(GOOGLE_AI_PROVIDER_MOCK);
+
+  GoogleGenAI.prototype.models = GOOGLE_GENAI_MOCK.models;
```
benoitf: should it be 
vi.mocked(GoogleGenAI.prototype.models.....) ?

Issue #411, opened 2025-09-22
Loading MCP servers is rejected as a whole


jeffmaury:
When loading MCP server from the registries, it any of one reject the list of servers (bad url, service down) the `listMCPServersFromRegistries` method throw an error and the whole operation is rejected

PR #583, opened 2025-10-20
chore: handle fetch mcp servers from registries error

SoniaSandler:
This PR adds handling for fetch mcp servers from registries error in `listMCPServersFromRegistries`.

Closes https://github.com/kortex-hub/kortex/issues/411

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-registry.ts
+++ b/packages/main/src/plugin/mcp/mcp-registry.ts
@@ -478,16 +478,12 @@ export class MCPRegistry {
       .concat(this.suggestedRegistries.map(registry => registry.url));
 
     for (const registryURL of serverUrls) {
-      let serverList: components['schemas']['ServerList'] | undefined;
       try {
-        serverList = await this.listMCPServersFromRegistry(registryURL);
-      } catch (error: unknown) {
-        console.error(`Failed fetch for registry ${registryURL}`, error);
-      }
-
-      if (serverList) {
+        const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
         // now, aggregate the servers from the list ensuring each server has an id
         serverDetails.push(...serverList.servers.map(({ server }) => this.enhanceServerDetail(server)));
+      } catch (error: unknown) {
+        console.error(`Failed fetch for registry ${registryURL}`, error);
       }
     }
     return serverDetails;


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -478,10 +478,14 @@ export class MCPRegistry {
...
     for (const registryURL of serverUrls) {
-      const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
+      try {
+        const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
 
-      // now, aggregate the servers from the list ensuring each server has an id
-      serverDetails.push(...serverList.servers.map(({ server }) => this.enhanceServerDetail(server)));
+        // now, aggregate the servers from the list ensuring each server has an id
+        serverDetails.push(...serverList.servers.map(({ server }) => this.enhanceServerDetail(server)));
+      } catch (error: unknown) {
```
simonrey1: suggestion(non-blocking): can we display the error in the console.error stream?

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -478,10 +478,14 @@ export class MCPRegistry {
       .concat(this.suggestedRegistries.map(registry => registry.url));
 
     for (const registryURL of serverUrls) {
-      const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
+      try {
+        const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
```
simonrey1: nitpick: My preference would be to limit the try catch block to the `listMCPServersFromRegistry` call that had errors, the next line (serverDetails.push...) seems ok

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -478,10 +478,14 @@ export class MCPRegistry {
...
 
-      // now, aggregate the servers from the list ensuring each server has an id
-      serverDetails.push(...serverList.servers.map(({ server }) => this.enhanceServerDetail(server)));
+        // now, aggregate the servers from the list ensuring each server has an id
+        serverDetails.push(...serverList.servers.map(({ server }) => this.enhanceServerDetail(server)));
+      } catch (error: unknown) {
+        console.error(`Failed fetch for registry ${registryURL}`);
```
jeffmaury:
```suggestion
        console.error(`Failed fetch for registry ${registryURL}`, error);
```

jeffmaury (approved):
LGTM

packages/main/src/plugin/mcp/mcp-registry.ts
simonrey1: remark: this is the same suggestion as https://github.com/kortex-hub/kortex/pull/583#discussion_r2447035757

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -478,10 +478,17 @@ export class MCPRegistry {
       .concat(this.suggestedRegistries.map(registry => registry.url));
 
     for (const registryURL of serverUrls) {
-      const serverList: components['schemas']['ServerList'] = await this.listMCPServersFromRegistry(registryURL);
+      let serverList: components['schemas']['ServerList'] | undefined;
+      try {
+        serverList = await this.listMCPServersFromRegistry(registryURL);
```
jeffmaury: suggestion: 

```
try {
  const serverList = await this.listMCPServersFromRegistry(registryURL);
  serverDetails.push(...serverList.servers.map(
} catch (error: unknown) {
  console.error(....)
}
```

simonrey1: this was done at first and I suggested to limit the try block here https://github.com/kortex-hub/kortex/pull/583#discussion_r2447042210

SoniaSandler: @simonrey1 @jeffmaury I don't have a particular preference for this. We can leave it as it is for now, or if someone has strong preference, we could change it back to 
```
try {
  const serverList = await this.listMCPServersFromRegistry(registryURL);
  serverDetails.push(...serverList.servers.map(
} catch (error: unknown) {
  console.error(....)
}
```

simonrey1: My comment was only a nitpick (https://github.com/kortex-hub/kortex/pull/583#discussion_r2447042210) so you can rollback to Jeff's solution no problem for me

PR #592, opened 2025-10-22
feat: register the command line generation for flow connections

benoitf:
related to scheduling


jeffmaury (approved):
LGTM

My only concern is about flow engines not exposing a command line

PR #593, opened 2025-10-22
refactor: use ipc handle in flow manager to handle commands

benoitf:
allow to handle commands (rather than adding them in index.ts)

### DIFF
--- a/packages/main/src/plugin/flow/flow-manager.ts
+++ b/packages/main/src/plugin/flow/flow-manager.ts
@@ -211,12 +211,14 @@ export class FlowManager implements Disposable {
     this.ipcHandle('flows:getLogCurrent', async (_listener): Promise<string> => {
       return this.getLogCurrent();
     });
+
     this.ipcHandle(
       'flows:dispatchLog',
       async (_listener, providerId: string, connectionName: string, flowId: string, taskId: string): Promise<void> => {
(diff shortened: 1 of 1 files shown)


benoitf:
yes, that's sad :-)

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -198,6 +200,42 @@ export class FlowManager implements Disposable {
...
+    });
+
+    this.ipcHandle('flows:listExecute', async (_listener): Promise<Array<FlowExecuteInfo>> => {
+      return this.listExecutions();
+    });
+
+    this.ipcHandle('flows:getLogCurrent', async (_listener): Promise<string> => {
+      return this.getLogCurrent();
+    });
+    this.ipcHandle(
```
simonrey1, line 215: nitpick: insert a new lines between handles

simonrey1 (approved):
LGTM, not tested

Maybe update the PR title to something like : "move methods to flow manager" or anything

packages/main/src/plugin/flow/flow-manager.ts
benoitf: solved

jeffmaury (approved):
LGTM

Issue #519, opened 2025-10-06
EPIC: Scheduling of Flows


slemeur:
As a AI Builder, I should be able to schedule the execution of the flows, so I can have the have the result handy on a certain set of criteria. 

## Pre-requisite

- Kortex is up and running

## Description

- In the the flow details, we would add a new tab for “Scheduling”  
- The user will have options for configuring the scheduling with different section  
  - When to run  
    - Start date & time  
    - End date or number of runs (optional)  
  - Frequency  
    - One-time  
    - Every X minutes/hours (e.g., every 15 min; with a min interval)  
    - Daily (every N days)  
    - Weekly (choose days: Mon–Sun, every N weeks)  
    - Monthly  
    - By date (e.g., 1st, 15th, 31st → handle short months)  
    - By weekday pattern (e.g., 2nd Tue, last Fri)  
    - Yearly

nichjones1:
Closing, no longer a priority

PR #603, opened 2025-10-23
chore: include API for registering scheduling extensions

benoitf:
Add a way to register a scheduler.

### DIFF
--- a/packages/main/src/plugin/scheduler/scheduler-registry.ts
+++ b/packages/main/src/plugin/scheduler/scheduler-registry.ts
@@ -1,0 +1,55 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


packages/main/src/plugin/scheduler/scheduler-registry.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor(
+    @inject(ApiSenderType)
+    private apiSender: ApiSenderType,
+  ) {}
+
+  async init(): Promise<void> {}
+
+  // Register a new scheduler
+  register(providerImpl: ProviderImpl, scheduler: ProviderScheduler): Disposable {
```
MarsKubeX, line 39: WDYT to have unit tests for this function?

benoitf: added

packages/main/src/plugin/scheduler/scheduler-registry.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.#schedulers.set(providerImpl.name, scheduler);
```
jeffmaury, line 40: An provider(extension) can register a single scheduler ?

benoitf: yes

benoitf: but an extension can register multiple providers

simonrey1 (approved):

PR #609, opened 2025-10-24
feat: added json validation when adding mcp registry

### LINKED ISSUE #347: validate mcp registry data before adding/handling the JSON
we're assuming that the JSON returned by mcp registries are compliant with the OpenAPI types generated for the v1.0.0.0

but there is no guarantee it's compliant. We should check if the JSON is more/less compliant or if the registry is compliant with v1.0.0 spec before creating entries as it can throw weird errors from end user POV once we've added the registry

gastoner:
Adds validation from yaml specification using `ajv` when adding/loading the mcp registry

Closes #347 

How to test: 
start the app 
Add https://registry.modelcontextprotocol.io/ registry if you don't have it already
and you should see some warinings in console (there will be plenty of warnings ~140 validation warnings)

Bonus:
https://github.com/kortex-hub/kortex/pull/849 adds on top of this PR indicator in mcp install list so you can see which one did not pass the validation

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-schema-validator.spec.ts
+++ b/packages/main/src/plugin/mcp/mcp-schema-validator.spec.ts
@@ -261,4 +261,19 @@ describe('validateSchemaData', () => {
     expect(result).toBe(false);
     expect(console.warn).toHaveBeenCalled();
   });
+
+  test('should not warn when suppressWarnings is true', () => {
+    const invalidRepository = {
+      url: 'https://github.com/example/repo',
+      // Missing 'source' field
+    };
+
+    const result = validator.validateSchemaData(invalidRepository, 'Repository', undefined, true);
+
+    expect(result).toBe(false);
+    expect(console.warn).not.toHaveBeenCalledWith(
+      expect.stringContaining('[MCPSchemaValidator] Failed to validate data against schema'),
+      expect.anything(),
+    );
+  });
 });
--- a/packages/main/src/plugin/mcp/mcp-schema-validator.ts
+++ b/packages/main/src/plugin/mcp/mcp-schema-validator.ts
@@ -31,13 +31,19 @@ export class MCPSchemaValidator {
    * @param jsonData - The data to validate
    * @param schemaName - The schema component name (e.g., 'ServerList', 'ServerResponse')
    * @param contextName - Optional context name for error messages (e.g., registry URL)
+   * @param suppressWarnings - Optional flag to suppress warnings (default: false)
    * @returns true if valid, false if invalid
    */
-  validateSchemaData(jsonData: unknown, schemaName: keyof components['schemas'], contextName?: string): boolean {
+  validateSchemaData(
+    jsonData: unknown,
+    schemaName: keyof components['schemas'],
+    contextName?: string,
+    suppressWarnings: boolean = false,
+  ): boolean {
     const validator = createValidator(schemaName);
     const isValid = validator(jsonData);
 
-    if (!isValid) {
+    if (!isValid && !suppressWarnings) {
       const context = contextName ? ` from '${contextName}'` : '';
       console.warn(
         `[MCPSchemaValidator] Failed to validate data against schema '${schemaName}'${context}.`,


packages/api/src/mcp/mcp-server-info.ts
```diff
@@ -10,3 +11,36 @@ export interface MCPRemoteServerInfo {
   description: string;
   url: string;
 }
+
+// Zod schema matching the OpenAPI ServerDetail type
+const MCPServerDetailSchema = z.object({
```
benoitf: I'm wondering here if we should generate the zod definition from ourselves

definition comes from `@kortex-hub/mcp-registry-types` where it's automatically extracted from openapi type.

so I would be cautious bringing our 'own manual' schema

if we update the library defining the type, we may forget to update to update the zod schema.

I would say that either zod type should come from the library or we define the zod type from the typescript definition

simonrey1: +1

I also think these should be generated from the OpenAPI spec in the repo you mention

pnpm-lock.yaml
```diff
@@ -186,8 +186,8 @@ importers:
         specifier: 'workspace:'
         version: link:packages/extension-api
       '@kortex-hub/mcp-registry-types':
-        specifier: ^1.2.3
-        version: 1.2.3
+        specifier: link:../mcp-registry-types
```
simonrey1: issue(blocking): version of package was replaced by a link

We should keep the version specifier and update it.

gastoner: will recreate it

benoitf:
I think the spec is more complicated than what is being done today.

Today we assume that the repository is based on a given version of the specification.

but in the specification, each server entry also have its own schema

```json
curl -s "https://registry.modelcontextprotocol.io/v0/servers/tools.ref%2Fref-tools-mcp/versions/latest"                                                                                        {
  "server": {
    "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
```


```json
curl -s "https://registry.modelcontextprotocol.io/v0/servers/ai.shawndurrani%2Fmcp-registry/versions/latest"
{
  "server": {
    "$schema": "https://static.modelcontextprotocol.io/schemas/2025-09-29/server.schema.json",
    "name": "ai.shawndurrani/mcp-registry",
    "description": "Search the public MCP Registry; discover servers and copy SSE URLs.",```
```

so I don't think it would work if there is only one schema to validate the field.

I guess the validation on the server entry should be done using the available `$schema`

gastoner:
@simonrey1 @benoitf I've updated the PR now it should check the schema from the specification

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -511,6 +516,108 @@ export class MCPRegistry {
...
+      if (!server?.name) {
+        continue;
+      }
+
+      const serverName = server.name;
+      const schemaUrl = server.$schema;
+
+      // If the server declares a $schema, validate against it
+      if (schemaUrl) {
+        const validator = await this.getServerSchemaValidator(schemaUrl);
```
simonrey1: question: In issue https://github.com/kortex-hub/kortex/issues/347, the problem seems to be that the server data is

>  incompatible with the OpenAPI types generated for the v1.0.0.0

Here we validate against the self-declared $schema.

But still it could be a totally different schema of what we expect so can you explain why this could fix the issue?

package.json
```diff
@@ -197,7 +197,8 @@
       "vitest>vite": "^6.1.0",
       "esbuild-register>esbuild": "^0.25.0",
       "tsx>esbuild": "^0.25.0",
-      "vite>esbuild": "^0.25.0"
+      "vite>esbuild": "^0.25.0",
+      "@kortex-hub/mcp-registry-types": "link:../mcp-registry-types"
```
gastoner: Will be updated when @kortex-hub/mcp-registry-types new version is released

pnpm-lock.yaml
simonrey1, line 1: issue: I see a lot of diffs here on this lock file, can you check again why it has so many differences?

Image

gastoner: Should be fixed now, will update the version to newer version of mcp-registry-types later after it is merged

packages/main/src/plugin/mcp/mcp-registry.spec.ts
```diff
@@ -16,82 +16,375 @@
...
+        },
+      ],
+      metadata: {
+        nextCursor: undefined,
+      },
+    };
+
+    // Mock two registry data fetches for pagination
+    fetchMock
+      .mockResolvedValueOnce({
```
simonrey1: nitpick: It is a bit too generic to mock any call to fetch with this value, I would prefer to mock the resolved value of a explicit function instead of the global fetch

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -518,6 +523,32 @@ export class MCPRegistry {
...
 
+  /**
+   * Validates JSON data against the ServerList schema.
+   * Uses the cached validator loaded during initialization.
+   *
+   * @param jsonData - The data to validate
+   * @param registryName - The name/URL of the registry (for error messages)
+   * @returns true if valid or if validation is skipped, false if invalid
+   */
+  private validateServerListData(jsonData: unknown, registryName: string): boolean {
```
simonrey1: suggestion(non-blocking): move this method to a new class to refactor the code, as the current class is quite big already.

gastoner:
@benoitf @simonrey1 @axel7083 can you PTAL?

gastoner:
@simonrey1 I could also add an idicator into the list of servers that the server has some error (e.g. red dot )

image (9)

cc @vancura

simonrey1:
> @simonrey1 I could also add an idicator into the list of servers that the server has some error (e.g. red dot )

Perfect, please follow up with that in the next PR, let's merge this one to start with?

gastoner:
@axel7083 @benoitf can you PTAL?

PR #626, opened 2025-10-28
feat: moved logic for hiding secrets to core kortex

### LINKED ISSUE #566: Move logic for hiding/showing secrets to Kortex core
Currently each extension has its own logic for hiding/showing secrets in logs. To be more consistent, this logic must be moved to the Kortex core.

MarsKubeX:
This PR centralizes secret hiding logic in the Kortex core by introducing a new `SecretManager` class in `@kortex-app/api`.

**Core Changes:**
- Added `SecretManager` class (`packages/api/src/secret-manager.ts`) that handles secret masking in YAML content using two strategies:
  - **By key**: Hides header values in recipe extensions based on sensitive key names (e.g., `authorization`, `x-api-key`)
  - **By value**: Hides exact secret values (e.g., API tokens) by replacing them with asterisks
- Supports both direct recipe YAML and Kubernetes ConfigMap-embedded recipes
- Added comprehensive unit tests for the `SecretManager`

**Issue**

closes #566


benoitf (commented):
hello, I'm not sure to understand why we would ask getSecretValues of getSensisitveKeys

The global idea of having it managed by the core means that the extension has nothing to do, all secrets should be hidden

I mean, the core has access to all the values of secrets so if a secret appears somewhere it can be hidden

And in the opposite, if an extension forget to say it's a sensitive key but store the secret through secured secrets then it should be automatically be 'sensitive'. Here it looks like there are two things that need to be synched and it'll likely fail at some point

packages/api/src/secret-manager.ts
```diff
@@ -0,0 +1,39 @@
...
+export class SecretManager {
+  /** Fixed-length mask to hide secrets without revealing their actual length */
+  private static readonly MASK = '********************';
+
+  constructor(private readonly secretValues: string[]) {}
+
+  /**
+   * Hides secrets in YAML content by replacing exact secret values with a fixed-length mask
+   */
+  hideSecretsInYaml(content: string): string {
```
simonrey1: nitpick: `hideSecretsInYaml` could be renamed to maskSecrets because it is not specific to YAML

packages/main/src/plugin/flow/flow-secret-collector.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  private async collectProviderCredentials(parsed: ParsedFlowContent): Promise<string[]> {
+    const secrets: string[] = [];
+
+    if (!parsed.settings?.goose_provider) {
+      return secrets;
+    }
+
+    // Map goose provider name to kortex provider ID
+    const inferenceProviderId = this.GOOSE_TO_KORTEX_PROVIDER_MAP[parsed.settings.goose_provider];
```
simonrey1: issue(blocking): this is specific to the "Goose" FlowProvider, and should not be included in the main

packages/main/src/plugin/flow/flow-secret-collector.ts
```diff
@@ -0,0 +1,135 @@
...
+    try {
+      const flowProvider = this.providerRegistry.getProvider(flowProviderId);
+      const flowConnection = flowProvider.flowConnections.find(({ name }) => name === connectionName);
+
+      if (!flowConnection) {
+        console.warn(`[FlowSecretCollector] Flow connection ${connectionName} not found`);
+        return [];
+      }
+
+      const flowContent = await flowConnection.flow.read(flowId);
```
simonrey1: issue(non-blocking): the performance might be degraded when having multiple flows, because when collecting secrets, you open each of the flow files. But you have no easier solution for now so this is not blocking.

packages/api/src/secret-manager.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export class SecretManager {
```
benoitf: AFAIK class/implementation do not belong to the the api package, ---> main service

benoitf (changes requested):
we do have a secret storage (SafeStorageRegistry)

so why it's not this registry that has the service to hide secrets for a given content.
It has access to all the secrets being used, why do we need to collect in flow-manager or flow-secret-collector ?


it has also be commented elsewhere, but flow objects can't have any dependency to an extension

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -93,6 +96,51 @@ export class FlowManager implements Disposable {
...
+    // Generate the Kubernetes YAML
+    const { resources } = await flowConnection.flow.generateKubernetesYAML({
+      flowId,
+      namespace,
+      hideSecrets: false,
+    });
+
+    // Hide secrets if requested
+    if (hideSecrets) {
+      return await this.safeStorageRegistry.hideSecretsInContent(resources);
```
benoitf, line 138: issue on the style. as it's a promise you can return the promise directly (no need to await)

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -23,6 +23,7 @@ import { ProviderRegistry } from '/@/plugin/provider-registry.js';
 import type { FlowExecuteInfo } from '/@api/flow-execute-info.js';
 import type { FlowInfo } from '/@api/flow-info.js';
 
+import { SafeStorageRegistry } from '../safe-storage/safe-storage-registry.js';
```
benoitf, line 26: prefer `/@/....` import if possible

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -93,6 +96,51 @@ export class FlowManager implements Disposable {
...
   }
 
+  /**
+   * Get flow information by flowId
+   * @param flowId  flow identifier
+   * @returns flow information or undefined if not found
+   */
+  getFlow(flowId: string): FlowInfo | undefined {
+    return this.all().find(f => f.id === flowId);
+  }
```
benoitf, line 106: question: how this method is used ?

packages/main/src/plugin/flow/flow-manager.ts
```diff
@@ -93,6 +96,51 @@ export class FlowManager implements Disposable {
...
+    // Get the flow provider to use
+    const flowProvider = this.provider.getProvider(providerId);
+    const flowConnection = flowProvider.flowConnections.find(({ name }) => name === connectionName);
+    if (!flowConnection) throw new Error(`cannot find flow connection with name ${connectionName}`);
+
+    // Generate the Kubernetes YAML
+    const { resources } = await flowConnection.flow.generateKubernetesYAML({
+      flowId,
+      namespace,
+      hideSecrets: false,
```
benoitf, line 133: question: from my understanding, there will be a follow up to no longer give this option to flow ? as it's managed by the core now

packages/main/src/plugin/safe-storage/safe-storage-registry.ts
```diff
@@ -154,6 +169,82 @@ export class SafeStorage {
...
+    const SECRET_KEYS = new Set([
+      'authorization',
+      'bearer',
+      'token',
+      'password',
+      'secret',
+      'api-key',
+      'apikey',
+      'x-api-key',
+    ]);
```
benoitf, line 231: not sure to follow why there is a whitelist ?

AFAIK it should replace everywhere no ?

packages/main/src/plugin/safe-storage/safe-storage-registry.ts
```diff
@@ -154,6 +169,82 @@ export class SafeStorage {
...
+      // Only add if the parent key is in the whitelist
+      const parentKey = currentPath[currentPath.length - 1]?.toLowerCase();
+      if (parentKey && SECRET_KEYS.has(parentKey)) {
+        secretValues.add(obj);
+      }
+    } else if (Array.isArray(obj)) {
+      for (const item of obj) {
+        this.extractSecretValuesFromObject(item, secretValues, currentPath);
+      }
+    } else if (obj !== null && typeof obj === 'object') {
```
benoitf, line 242: would it be null or undefined as AFAIK we've rules to avoid the usage of `null`

packages/main/src/plugin/safe-storage/safe-storage-registry.ts
MarsKubeX: MCP configurations are stored as JSON objects. To hide secrets in the YAML, I need to extract individual values from this JSON and compare them. Without the whitelist, I'd extract all strings from the JSON, causing false positives. With the whitelist, I only extract values whose parent key indicates they're secrets (authorization, token, api-key, etc.)

packages/main/src/plugin/flow/flow-manager.ts
MarsKubeX: you're right, it was a function I've created for the old strategy. I'll remove it.

packages/main/src/plugin/safe-storage/safe-storage-registry.ts
benoitf: hello, what would be the false positives ?
we store basically tokens which are long strings, so how can it be used inside the file without being the right secret ?

benoitf: also, let say I use this function for a different use-case, like displaying registry configurations, I don't think it may have authorization field at all.

I expect that calling this function works OOTB

MarsKubeX: Currently, when I decrypt the MCP config from SafeStorage:
`safeStorage.getDecrypted('core.mcp:registry:configurations');`
Since mcp configs are stored as JSON objecst the decrypted result is something like:
`[
  {
    "remoteId": 0,
    "serverId": "com.github.mcp",
    "headers": {
      "Authorization": "Bearer github_pat_token..."
    }
  },
  ...
]`
When I compare this with the generated YAML, non-secret values also match (like "com.github.mcp"), causing false positives—hiding service names and IDs that aren't actually secrets.

I understand your point about having an "OOTB" solution. An alternative would be to refactor mcp-registry.ts to store only header secrets in SafeStorage (not the entire config object), and keep metadata elsewhere. However:
- This would require significant changes to how MCP configs are stored/retrieved
- It needs proper analysis to avoid breaking existing functionality
- It would require data migration for existing users

What I think for this PR, is to keep the whitelist approach (it's safe and targeted). If we want to refactor the storage model, let's handle it in a separate issue.

benoitf: > What I think for this PR, is to keep the whitelist approach (it's safe and targeted). If we want to refactor the storage model, let's handle it in a separate issue.

we can't. SafeStorage class should not have any knowledge about 'Authentication". It's something that MCPRegistry is aware but not SafeStorage.

MarsKubeX: Created issue #673 to refactor mcp-registry.

PR #640, opened 2025-10-30
chore(test): create mcp github connection

### LINKED ISSUE #639: Configure and verify GitHub MCP Server installation

serbangeorge-m:
Closes #639

### DIFF
--- a/tests/playwright/src/fixtures/provider-fixtures.ts
+++ b/tests/playwright/src/fixtures/provider-fixtures.ts
@@ -19,7 +19,7 @@ import { type ElectronApplication, type Page } from '@playwright/test';
 
 import { MCP_SERVERS, type MCPServerId } from '../model/core/types';
 import { NavigationBar } from '../model/navigation/navigation';
-import { getMCPToken, hasMCPToken, PROVIDERS, type ResourceId } from '../utils/resource-helper';
+import { PROVIDERS, type ResourceId } from '../utils/resource-helper';
 import { type ElectronFixtures, getFirstPage, launchElectronApp, test as base } from './electron-app';
 
 interface WorkerFixtures {
@@ -108,18 +108,17 @@ export const test = base.extend<ElectronFixtures, WorkerFixtures>({
         const mcpPage = await workerNavigationBar.navigateToMCPPage();
 
         for (const id of mcpServers) {
-          if (hasMCPToken(id)) {
-            try {
-              const server = MCP_SERVERS[id];
-              const token = getMCPToken(id);
-              if (!token) {
-                throw new Error(`${server.envVarName} environment variable is not set`);
-              }
-              await mcpPage.createServer(server.serverName, token);
-              configured.push(id);
-            } catch (error) {
-              console.warn(`MCP setup skipped for ${id}:`, error);
-            }
+          const server = MCP_SERVERS[id];
+          const token = process.env[server.envVarName];
+          if (!token) {
+            throw new Error(`${server.envVarName} environment variable is not set`);
+          }
+
+          try {
+            await mcpPage.createServer(server.serverName, token);
+            configured.push(id);
+          } catch (error) {
+            console.warn(`MCP setup skipped for ${id}:`, error);
           }
         }
 
@@ -148,10 +147,6 @@ export const test = base.extend<ElectronFixtures, WorkerFixtures>({
 
   page: async ({ workerPage }, use): Promise<void> => {
     await use(workerPage);
-  },
-
-  navigationBar: async ({ workerNavigationBar }, use): Promise<void> => {
-    await use(workerNavigationBar);
   },
 });
 
--- a/tests/playwright/src/utils/resource-helper.ts
+++ b/tests/playwright/src/utils/resource-helper.ts
@@ -15,8 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import type { MCPServerId, SettingsResourceId } from 'src/model/core/types';
-import { MCP_SERVERS } from 'src/model/core/types';
+import type { SettingsResourceId } from 'src/model/core/types';
 
 export interface ResourceConfig {
   readonly envVarName: string;
@@ -47,13 +46,4 @@ export function getProviderCredentials(providerId: ResourceId): string | undefin
 
 export function hasApiKey(providerId: ResourceId): boolean {
   return !!getProviderCredentials(providerId);
-}
-
-export function getMCPToken(mcpServerId: MCPServerId): string | undefined {
-  const server = MCP_SERVERS[mcpServerId];
-  return process.env[server.envVarName];
-}
-
-export function hasMCPToken(mcpServerId: MCPServerId): boolean {
-  return !!getMCPToken(mcpServerId);
 }


tests/playwright/src/fixtures/provider-fixtures.ts
```diff
@@ -59,14 +74,33 @@ export const test = base.extend<TestFixtures, WorkerFixtures>({
...
       try {
-        await createResource(workerPage, resource);
+        const provider = PROVIDERS[resource];
+        const credentials = process.env[provider.envVarName];
+        if (!credentials) {
+          throw new Error(`${provider.envVarName} environment variable is not set`);
+        }
+
+        await workerNavigationBar.settingsLink.click();
+        const settingsPage = new SettingsPage(workerPage);
```
cbr7:
```suggestion
        const settingsPage = await workerNavigationBar.navigateToSettingsPage();
```

tests/playwright/src/fixtures/provider-fixtures.ts
```diff
@@ -59,14 +74,33 @@ export const test = base.extend<TestFixtures, WorkerFixtures>({
...
+        await settingsPage.createResource(resource, credentials);
+
         await use();
       } finally {
         try {
-          await deleteResource(workerPage, resource);
+          await workerNavigationBar.settingsLink.click();
+          const settingsPage = new SettingsPage(workerPage);
```
cbr7:
```suggestion
          const settingsPage = await workerNAvigationBar.navigateToSettingsPage();
```

tests/playwright/src/model/pages/mcp-ready-tab-page.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  async deleteServer(serverName: string): Promise<void> {
+    await expect
+      .poll(
+        async () => {
+          const count = await this.findServer(serverName).count();
+          if (count === 0) return true;
+
+          await this.findServer(serverName).first().getByRole('button', { name: 'Remove instance of MCP' }).click();
+          await this.page.waitForTimeout(500);
```
cbr7: This kind of pattern should be avoided at all costs, need to wait for observable behavior, not hardcoded waits, this will cause problems in higher lag environments.

tests/playwright/src/model/pages/mcp-ready-tab-page.ts
```diff
@@ -0,0 +1,71 @@
...
+      return false;
+    }
+  }
+
+  async deleteServer(serverName: string): Promise<void> {
+    await expect
+      .poll(
+        async () => {
+          const count = await this.findServer(serverName).count();
+          if (count === 0) return true;
```
cbr7: method returns void, why does it return true here?

tests/playwright/src/model/pages/mcp-ready-tab-page.ts
```diff
@@ -0,0 +1,71 @@
...
+
+          await this.findServer(serverName).first().getByRole('button', { name: 'Remove instance of MCP' }).click();
+          await this.page.waitForTimeout(500);
+          return false;
```
cbr7: return statement does nothing, execution of the method is already finished, and also it shouldn't return a value anyway since method is defined as void.

tests/playwright/src/model/pages/mcp-ready-tab-page.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.table.getByRole('row').filter({ hasText: serverName });
+  }
+
+  async isServerConnected(serverName: string, timeout = 5_000): Promise<boolean> {
+    try {
+      await expect.poll(async () => await this.findServer(serverName).count(), { timeout }).toBeGreaterThan(0);
+      return true;
+    } catch {
+      return false;
+    }
```
cbr7:
```suggestion
      return await this.findServer(serverName).count() > 0;
```

cbr7: If you want to also have a waiter function for it that is possible but it should not be implemented inside this boolean function, not to mention that the assert can be done in the test itself, if you expect the value to be higher than 0 you can check that in the test.

tests/playwright/src/model/pages/mcp-ready-tab-page.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  async deleteServer(serverName: string): Promise<void> {
+    await expect
```
cbr7: In general this pattern of having asserts in try catch blocks should be avoided, if you expect some type of value it should be handled in the test itself, the delete method should just try to delete and nothing else.

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +45,23 @@ export class McpInstallTabPage extends BaseTablePage {
...
+  }
+
+  async installRemoteServer(serverName: string, token: string): Promise<void> {
+    const serverRow = this.findServer(serverName);
+    await expect(serverRow).toBeVisible();
+
+    const installButton = serverRow.getByRole('button', { name: 'Install Remote server' });
+    await installButton.click();
+
+    const passwordInput = this.page.locator('input[type="password"]');
```
cbr7: this locator should be defined for the entire class scope, not just locally in this method and also that field as aria-label I think so it should work as getByLabel('password');

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +45,23 @@ export class McpInstallTabPage extends BaseTablePage {
...
+    await expect(passwordInput).toBeVisible({ timeout: 5_000 });
+    await passwordInput.fill(token);
```
cbr7:
```suggestion
    await passwordInput.fill(token);
    await expect(passwordInput).toHaveValue(token);
```

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +45,23 @@ export class McpInstallTabPage extends BaseTablePage {
...
+
+    const connectButton = this.page.getByRole('button', { name: 'Connect' });
+    await connectButton.click();
```
cbr7:
```suggestion
    await expect(connectButton).toBeEnabled();
    await connectButton.click();
```

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +45,23 @@ export class McpInstallTabPage extends BaseTablePage {
...
+  findServer(serverName: string): Locator {
+    return this.table.getByRole('row').filter({ hasText: serverName });
+  }
+
+  async installRemoteServer(serverName: string, token: string): Promise<void> {
+    const serverRow = this.findServer(serverName);
+    await expect(serverRow).toBeVisible();
+
+    const installButton = serverRow.getByRole('button', { name: 'Install Remote server' });
+    await installButton.click();
```
cbr7, line 63:
```suggestion
    await expect(installButton).toBeEnabled();
    await installButton.click();
```

tests/playwright/src/model/pages/mcp-page.ts
```diff
@@ -41,20 +42,35 @@ export class McpPage extends BasePage {
...
+    if (await readyTab.isServerConnected(serverName)) {
+      console.log(`MCP server ${serverName} is already connected, skipping...`);
+      return;
+    }
+
+    const installTab = await this.openInstallTab();
+    await installTab.installRemoteServer(serverName, token);
+
+    const readyTabAfterInstall = await this.openReadyTab();
+    expect(await readyTabAfterInstall.isServerConnected(serverName, 15_000)).toBe(true);
```
cbr7: In combination with the suggestion from the method itself this assert needs to be refactored, it should look like


```suggestion
    await expect.poll(async () => readyTabAfterInstall.isServerConnected(serverName, 15_000)).toBeTruthy();
```

tests/playwright/src/utils/resource-helper.ts
```diff
@@ -44,51 +39,19 @@ export const PROVIDERS = {
...
 export type ResourceId = keyof typeof PROVIDERS;
 
-async function navigateToResourcesPage(page: Page): Promise<SettingsResourcesPage> {
-  await waitForNavigationReady(page);
-  const navigationBar = new NavigationBar(page);
-  const settingsPage = await navigationBar.navigateToSettingsPage();
-  return await settingsPage.openResources();
+export interface MCPServerConfig {
+  readonly envVarName: string;
+  readonly serverName: string;
```
cbr7: should be stored in types.ts file

tests/playwright/src/utils/resource-helper.ts
```diff
@@ -44,51 +39,19 @@ export const PROVIDERS = {
...
-  if ((await resourcesPage.getCreatedResourceFor(provider.resourceId).count()) > 0) {
-    // Resource already exists
-    return;
-  }
+export const MCP_SERVERS = {
+  github: {
+    envVarName: 'GITHUB_TOKEN',
+    serverName: 'com.github.mcp',
+  },
+} as const satisfies Record<string, MCPServerConfig>;
```
cbr7: should be stored in types.ts file

tests/playwright/src/model/pages/mcp-install-tab-page.ts
simonrey1: I believe that playwright checks for the button to be enabled before clicking, but I can't find the exact documentation

simonrey1: https://playwright.dev/docs/actionability
Screenshot 2025-10-30 at 11 57 43

cbr7: @simonrey1 this is true somewhat, in that there are checks for actionability, however it does not wait for actionability, meaning that if the check fails it will throw an error, the click action itself does not have a waiter function, not truly.

The assert will check and recheck for actionality in a cycle until either the timeout expires or the conditions are met.

simonrey1: I am pretty sure it waits for the element to be enabled when you do the .click()

cbr7: @simonrey1 it waits only in so far that it takes time to check all the requirements once, but if one of the checks fails, say for example the button takes some time to be enabled because some background validation being done in the backend it will not recheck that requirement again, if it worked the way you think there wouldnt be any point to having asserts in PW.

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +47,25 @@ export class McpInstallTabPage extends BaseTablePage {
...
+
+    const installButton = serverRow.getByRole('button', { name: 'Install Remote server' });
+    await expect(installButton).toBeEnabled();
+    await installButton.click();
+
+    await expect(this.passwordInput).toBeVisible({ timeout: 5_000 });
+    await this.passwordInput.fill(token);
+    await expect(this.passwordInput).toHaveValue(token);
+
+    const connectButton = this.page.getByRole('button', { name: 'Connect' });
```
cbr7: missed this one initially, this should also be defined at the top of the class.

tests/playwright/src/model/pages/mcp-install-tab-page.ts
```diff
@@ -45,4 +47,25 @@ export class McpInstallTabPage extends BaseTablePage {
...
+
+  async installRemoteServer(serverName: string, token: string): Promise<void> {
+    const serverRow = this.findServer(serverName);
+    await expect(serverRow).toBeVisible();
+
+    const installButton = serverRow.getByRole('button', { name: 'Install Remote server' });
+    await expect(installButton).toBeEnabled();
+    await installButton.click();
+
+    await expect(this.passwordInput).toBeVisible({ timeout: 5_000 });
```
cbr7: Default waiting time is 5 seconds, so adding an explicite timeout of 5 seconds does nothing, if you wanted to make it some other value then it makes sense, but 5 seconds is default.

tests/playwright/src/model/pages/mcp-page.ts
```diff
@@ -41,20 +42,38 @@ export class McpPage extends BasePage {
...
+  }
 
-    const installTabPage = new McpInstallTabPage(this.page);
-    await installTabPage.waitForLoad();
-    return installTabPage;
+  async deleteServer(serverName: string): Promise<void> {
+    const readyTab = await this.openReadyTab();
+    await readyTab.deleteServer(serverName);
+    await expect.poll(async () => await readyTab.isServerConnected(serverName), { timeout: 10_000 }).toBeFalsy();
+    await readyTab.verifyEmpty();
```
cbr7: This validation does not seem to be correct here, method is only in charge of delete one mcp server, why should there be an expectation that the list be empty when there could be multiple servers to begin with, this is a different validation that should be called separatly within a test when the expectation is that the list must be empty, but not here.

tests/playwright/src/model/pages/mcp-install-tab-page.ts
simonrey1: I have setup a demo repository to show how the .click() can wait 3s for the button to be enabled.

https://github.com/simonrey1/playwright-click-demo

tests/playwright/src/fixtures/provider-fixtures.ts
```diff
@@ -75,13 +100,59 @@ export const test = base.extend<TestFixtures, WorkerFixtures>({
...
     await use(workerElectronApp);
   },
 
   page: async ({ workerPage }, use): Promise<void> => {
     await use(workerPage);
   },
+
+  navigationBar: async ({ workerNavigationBar }, use): Promise<void> => {
+    await use(workerNavigationBar);
+  },
```
cbr7: This fixture isn't needed I don't think, since there is an identical one in the electron-app fixture, which gets inherited by this one.

Issue #598, opened 2025-10-23
Display name of the flows instead of path


slemeur:
As a user, I'll differentiate the flow by their name and not by the path. The path should remain a secondary information. 

We should update the flow list:

<img width="787" height="142" alt="Image" src="https://github.com/user-attachments/assets/2fe0cc15-cdfd-4f61-bcb7-8711b7c48a4e" />

As well as the detail screen for the flow (in the header).

slemeur:
are you thinking about it like a hover action? or an icon that would be always visible?
also, I think in this context as a user I would like to "open the folder" where the recipe is stored, not necessarily copy its path.

benoitf:
for now if you click, as other items, you go to the details page

vancura:
BTW, one more idea. for `Open the folder`, consider rewording to:

- macOS: `Show in Finder` (or `Reveal in Finder` - both are used, although `Show in Finder` is more common in system apps, especially recently (reveal is slowly deprecated, from what I have been seeing recently). If we want to go the extra mile, we can query what's the default system file browser, so people using Path Finder get "Show in Path Finder", but that's a stretch goal, but would be an awesome detail that would make this look native)

- Windows: `Reveal in File Explorer` (standard)

- Linux: `Open Containing Folder` (standard, yet, you know, there are no HIGs on Linux, except for Gnome and Elementary, if I am right)

PR #642, opened 2025-10-31
chore: add flow name to flow list and use path as secondary text

SoniaSandler:
This PR updates the flow/path column in the flow list so that the main text is the flow's name and the secondary text is the flow's path (duplicate flows in screenshot are due to goose bug)

Screenshot 2025-10-31 at 12 05 18 AM

Closes https://github.com/kortex-hub/kortex/issues/598

### DIFF
--- a/packages/renderer/src/lib/flows/FlowDetails.svelte
+++ b/packages/renderer/src/lib/flows/FlowDetails.svelte
@@ -10,6 +10,7 @@ import Route from '/@/Route.svelte';
 import { flowsInfos } from '/@/stores/flows';
 import { executeFlowsInfo } from '/@/stores/flows-execute';
 import { providerInfos } from '/@/stores/providers';
+import type { FlowInfo } from '/@api/flow-info';
 import type { ProviderFlowConnectionInfo } from '/@api/provider-info';
 
 import FlowActions from './FlowActions.svelte';
@@ -30,24 +31,7 @@ let provider = $derived($providerInfos.find(provider => provider.id === provider
 let connection: ProviderFlowConnectionInfo | undefined = $derived(
   provider?.flowConnections.find(connection => connection.name === connectionName),
 );
-let { path, name } = $derived.by(() => {
-  const matchingFlow = $flowsInfos.find(
-    flowInfo =>
-      flowInfo.id === flowId && flowInfo.providerId === providerId && flowInfo.connectionName === connectionName,
-  );
-  if (!matchingFlow) {
-    return { path: atob(flowId), name: atob(flowId) };
-  }
-  return matchingFlow;
-});
-
-let flowInfo = $derived({
-  providerId,
-  id: flowId,
-  path,
-  name,
-  connectionName,
-});
+let flowInfo: FlowInfo | undefined = $derived($flowsInfos.find(({ id }) => id === flowId));
 
 let flowContent: string | undefined = $state(undefined);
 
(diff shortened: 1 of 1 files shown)


SoniaSandler:
@vancura @slemeur I have added the flow's name to the table. Is there any related functionality you'd like me to change or adjust due to this UI update?
Based on https://github.com/kortex-hub/kortex/issues/598#issuecomment-3441555564, it looks like the preferred approach (details page) is outside of this issue's scope.

MarsKubeX (approved):
LGTM. Tested and working as well.

packages/renderer/src/lib/flows/flow-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  const fileNameFromPath = flowPath.split(/[/\\]/).pop();
+  if (fileNameFromPath) {
+    const name = fileNameFromPath.split('.');
+    extractedName = fileNameFromPath.startsWith('.') ? name[1] : name[0];
+    if (extractedName) {
+      return extractedName;
+    }
+  }
+  return flowPath;
+}
```
benoitf: hello, I'm wondering why it's not using classic methods from Node.js path

like `path.basename`, `path.extname`, etc.

https://nodejs.org/api/path.html#pathbasenamepath-suffix

it looks like we're reimplementing existing functions.

SoniaSandler: @benoitf since it's on the frontend, using the path node module cause the `Error: Module "node:path" has been externalized for browser compatibility` error. Is there another way to use it on the frontend?
Or we could create a separate function in the backend that uses `path` and call it on the frontend

benoitf: I'm expecting to have it on the backend and adding property into `FlowInfo` object (adding `name` property in addition to the path, etc

IMHO frontend should just display things. Here it's error prone, might not work on some cases, etc.

SoniaSandler: @benoitf I updated the code. However, one difference between this approach and the previous one is handling file names with multiple `.` like `file-name-1.config.yaml`.
With this approach, it will return `file-name-1.config` instead of `file-name-1`.
It's a nitpick, but I just wanted to point it out in case there is a preference for a specific return value.

packages/renderer/src/lib/flows/FlowDetails.svelte
```diff
@@ -29,12 +30,22 @@ let provider = $derived($providerInfos.find(provider => provider.id === provider
...
+let { path, name } = $derived.by(() => {
+  const matchingFlow = $flowsInfos.find(
+    flowInfo =>
+      flowInfo.id === flowId && flowInfo.providerId === providerId && flowInfo.connectionName === connectionName,
+  );
+  if (!matchingFlow) {
+    return { path: atob(flowId), name: atob(flowId) };
+  }
+  return matchingFlow;
+});
```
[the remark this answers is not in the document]
benoitf: need to sync with @axel7083 
I see in his PR

https://github.com/kortex-hub/kortex/pull/672/files#diff-e22e1239f1e269f605e935e7f0fcff1212a58619149d05b61187a923c25bf9ec

```
let flowInfo: FlowInfo | undefined = $derived($flowsInfos.find(({ id }) => id === flowId));
```

benoitf (approved):
issue: before merging, clarify the way to get Flow from its id in FlowDetails page as there is another PR doing something different so it will result in a conflict so better to agree before

packages/renderer/src/lib/flows/flow-utils.ts
benoitf: thanks

packages/renderer/src/lib/flows/FlowDetails.svelte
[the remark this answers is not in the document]
axel7083: I have a preference for the solution without default, let's be undefined if this is not found, and the component should have ` {#if flowInfo`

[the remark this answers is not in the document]
SoniaSandler: @axel7083 @benoitf I updated the PR with some of the changes made in https://github.com/kortex-hub/kortex/pull/672

Issue #645, opened 2025-10-31
Input fields on "Install Remote MCP Server" page have static IDs and aria-labels


amisskii:
On the "Install Remote MCP Server" configuration screen (the page for "Adding com.github.mcp" shown in the screenshot), the input fields for variables (like the GitHub Personal Access Token) are failing to receive correct, unique identifiers.

Instead of having a functional, unique ID and name, attributes are rendered as static text.

<input id="password-{id}" name="password-undefined" type="password" ...>

<img width="1150" height="694" alt="Image" src="https://github.com/user-attachments/assets/dfc53486-12de-47fb-821e-4a9244db6aa2" />

PR #647, opened 2025-10-31
chore: refactor IDs and aria-labels for MCP server inputs

amisskii:
Signed-off-by: Anton Misskii <amisskii@redhat.com>

This PR updates the id and aria-label attributes for input fields on the Adding MCP Server page

Example for GitHub token field:

Before: id="password-" aria-label="password"

After: id="variable-github_mcp_pat-value" aria-label="Header used for authentication"

issue: https://github.com/kortex-hub/kortex/issues/645

### DIFF
--- a/packages/renderer/src/lib/mcp/setup/InputArgument.svelte
+++ b/packages/renderer/src/lib/mcp/setup/InputArgument.svelte
@@ -37,7 +37,7 @@ function onSelectChange(
 </script>
 
 <Markdown markdown={object.description} />
-<div class="flex flex-row items-center gap-x-2" aria-label="object-value-{id}">
+<div class="flex flex-row items-center gap-x-2">
   {#if object.isSecret}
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/ui/PasswordInput.svelte
```diff
@@ -8,11 +8,13 @@ import Fa from 'svelte-fa';
 type Props = Omit<ComponentProps<Input>, 'value'> & {
   password?: string;
   passwordHidden?: boolean;
+  description?: string;
 };
 
 let {
   id,
   name,
+  description,
```
benoitf, line 17: is there a matching podman desktop change?

This file comes unmodified from podman desktop so it should be a cherry-pick from the upstream project

amisskii:
> I'm really unclear why this PR is needed to be honest.

Currently, the input fields have incorrect aria-labels and IDs. In this pr I’m refactoring them so they will have the correct aria-labels as intended. You can also check the related issue for the details.

Issue #659, opened 2025-11-03
App hangs due to a race condition if users starts and stops the app very quickly


cbr7:
There seems to be a race condition in the app which causes the app to hang on startup if the users opens/closes the app very quickly in sequence, it seems the issue occurs whenever the app is closed before extensionLoader finishes to initialize.

PR #660, opened 2025-11-03
chore: fix for extensionLoader initialization race condition on startup

cbr7:
Fix for extensionLoader initialization race condition on startup/shutdown.

https://github.com/kortex-hub/kortex/issues/659


benoitf (changes requested):
as reported by @simonrey1 these changes should not be in Kortex

cbr7:
@simonrey1 sure, not a problem

Should I make a PR in PD repo with this?

Issue #458, opened 2025-09-25
Kubernetes deployment for parametrized flow


jeffmaury:
Current implementation is based on jobs. 

If there are parameters, that could be:

- jobs with config map
- rest service
- rest service + web ui

slemeur:
In order to make things a simple to start with, a flow will always be deployed with a specific set of values. That means that the user will have to provide the values for each of the required fields.

PR #672, opened 2025-11-04
feat(flow): recognise flow parameters

axel7083:
## Description

A goose recipe can have a series of parameters in its definition, e.g;

```yaml
[...]
parameters:
  - key: pr
    input_type: string
    requirement: required
    description: name of the pull request
  - key: repo
    input_type: string
    requirement: optional
    description: name of the repo. uses the current one if not selected
    default: "https://github.com/kortex-hub/kortex"
```

Thoses may have a default, but in the case they don't we need to ask the user those values before starting it locally or deploying to Kubernetes.

First step is to recognise those parameters; let's add it to the Flow interface in the extension-api.

## Screeshots

I updated the details UI to display nicely the flow details including the parameters when there are some.

| Before | After |
| --- | --- |
| image| image |

## Related issues

Required for
- https://github.com/kortex-hub/kortex/issues/458

## Testing

You can try a recipe with some parameters to see them display, (checkout https://github.com/block/goose/blob/main/documentation/src/pages/recipes/data/recipes/analyze-pr.yaml)

### DIFF
--- a/extensions/goose/src/goose-recipe.ts
+++ b/extensions/goose/src/goose-recipe.ts
@@ -61,9 +61,9 @@ export interface RecipeExtension {
 export interface RecipeParameter {
   key: string;
   input_type: string;
-  requirement: 'optional' | 'required';
+  requirement: 'optional' | 'required' | 'user_prompt';
   description: string;
-  default: string;
+  default?: string;
 }
 
 export interface RecipeWithExtensions {
@@ -254,7 +254,7 @@ export class GooseRecipe implements Disposable {
     const parsed = parse(content);
 
     const parameters: Array<FlowParameter> = (parsed['parameters'] ?? []).map((param: RecipeParameter) => ({
-      required: param.requirement === 'required',
+      required: param.requirement !== 'optional',
       name: param.key,
       description: param.description,
       default: param.default,
--- a/packages/api/src/flow-info.ts
+++ b/packages/api/src/flow-info.ts
@@ -15,6 +15,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+import type { FlowParameter } from '@kortex-app/api';
 
 export interface FlowInfo {
   /**
@@ -26,11 +27,5 @@ export interface FlowInfo {
   id: string;
   path: string;
   name: string;
-  parameters?: Array<{
-    required: boolean;
-    name: string;
-    description: string;
-    format: string;
-    default: string;
-  }>;
+  parameters?: Array<FlowParameter>;
 }
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -408,7 +408,7 @@ declare module '@kortex-app/api' {
     name: string;
     description: string;
     format: string;
-    default: string;
+    default?: string;
   }
 
   export interface Flow {


extensions/goose/src/goose-recipe.ts
```diff
@@ -57,6 +58,14 @@ export interface RecipeExtension {
   headers?: Record<string, string>;
 }
 
+export interface RecipeParameter {
+  key: string;
+  input_type: string;
+  requirement: 'optional' | 'required';
```
gastoner: Question: 
Can we use instead of `requirement: 'optional' | 'required';` => `optional: boolean` / `required: boolean` ?

Since this can be really just 2 values right?

extensions/goose/src/goose-recipe.ts
```diff
@@ -236,6 +253,14 @@ export class GooseRecipe implements Disposable {
 
     const parsed = parse(content);
 
+    const parameters: Array<FlowParameter> = (parsed['parameters'] ?? []).map((param: RecipeParameter) => ({
+      required: param.requirement === 'required',
```
gastoner: Folowing https://github.com/kortex-hub/kortex/pull/672/files#r2493135264
Then this would be e.g. `required: param.isRequired`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -403,9 +403,18 @@ declare module '@kortex-app/api' {
     vmTypeDisplayName?: string;
   }
 
+  export interface FlowParameter {
+    required: boolean;
+    name: string;
+    description: string;
+    format: string;
+    default: string;
```
benoitf: issue: IMHO default should be optional

if it's not required why should I bring a default value

extensions/goose/src/goose-recipe.ts
benoitf: looking at goose documentation https://block.github.io/goose/docs/guides/recipes/recipe-reference/#parameter-requirements there are:


```
required: Parameter must be provided when using the recipe
optional: Can be omitted if a default value is specified
user_prompt: Will interactively prompt the user for input if not provided
```

question: if a user recipe has `user_prompt`, what happens with this PR ?

packages/api/src/flow-info.ts
```diff
@@ -25,4 +25,11 @@ export interface FlowInfo {
 
   id: string;
   path: string;
+  parameters?: Array<{
+    required: boolean;
+    name: string;
+    description: string;
+    format: string;
+    default: string;
+  }>;
```
benoitf: question: I have the same question than coderabbit, if we only store data collected from extensions, it makes sense to use the extension API. If it's an internal data (with more stuff, it doesn't)

packages/extension-api/src/extension-api.d.ts
axel7083: Yes sure

packages/api/src/flow-info.ts
axel7083: Replaced with extension api interface :heavy_check_mark:

extensions/goose/src/goose-recipe.ts
axel7083: I added the `user_prompt` but I think we should keep the `required: boolean;` in the extension-api.

Technically if this is user_prompt, this is required

packages/renderer/src/lib/flows/FlowDetails.svelte
```diff
@@ -95,11 +99,71 @@ function setSelectedFlowExecuteId(flowExecuteId: string): void {
...
+                <DetailsCell>{connection?.name}</DetailsCell>
+              </tr>
+              <tr>
+                <DetailsCell>Path</DetailsCell>
+                <DetailsCell>{flowInfo.path}</DetailsCell>
+              </tr>
+            </DetailsTable>
+          </div>
+
+          {#if (flowInfo.parameters ?? []).length > 0}
```
simonrey1, line 123: suggestion(non-blocking): this could be done in a separate Svelte component

PR #684, opened 2025-11-06
chore: add ramalama extension

benoitf:
search for containers that are RamaLama containers and then register inference connection to them (if any)

⚠️ rely on the container extension https://github.com/kortex-hub/kortex/pull/678

so it can't work without the other one.

extracted from the full extension that has been cut into 2 parts https://github.com/kortex-hub/kortex/pull/507

### DIFF
--- a/extensions/ramalama/README.md
+++ b/extensions/ramalama/README.md
@@ -5,6 +5,7 @@ Integrates RamaLama local AI models with Kortex. Automatically discovers and reg
 ## Requirements
 
 - [RamaLama](https://ramalama.ai/) installed and running locally
+- Container engine running (Podman or Docker)
 
 ## Usage
 


extensions/ramalama/src/helper/container-ramalama-finder.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ContainerInfo } from 'dockerode';
+
+import type { RamalamaContainerInfo } from '/@/api/ramalama-container-info';
+
+export class ContainerRamaLamaFinder {
+  private isRamalamaContainer(container: ContainerInfo): container is RamalamaContainerInfo {
```
simonrey1, line 24: suggestion(non-blocking): these two methods can be static

```ts

  private static isRamalamaContainer(container: ContainerInfo): container is RamalamaContainerInfo {
    return container.Labels?.['ai.ramalama'] !== undefined;
  }

  static async getRamaLamaContainers(containers: ContainerInfo[]): Promise<RamalamaContainerInfo[]> {
    // It should include label 'ai.ramalama'
    return containers.filter(ContainerRamaLamaFinder.isRamalamaContainer);
  }

```

benoitf: -1
the pattern here is to use DI. We should not use static methods while this is something we inject through dependency injection

extensions/ramalama/src/helper/container-model-extractor.ts
```diff
@@ -0,0 +1,33 @@
...
+ ***********************************************************************/
+
+import type { ModelInfo } from '/@/api/model-info';
+import type { RamalamaContainerInfo } from '/@/api/ramalama-container-info';
+
+export class ContainerModelExtractor {
+  async extractModelInfo(container: RamalamaContainerInfo): Promise<ModelInfo> {
+    const modelName = container.Labels['ai.ramalama.model'];
+    const portLabel = container.Labels['ai.ramalama.port'];
+    const port = portLabel ? Number.parseInt(portLabel, 10) : 0;
```
simonrey1, line 26: issue: as the bot said in https://github.com/kortex-hub/kortex/pull/684/files#r2498329582, port could be `NaN`

extensions/ramalama/src/helper/container-model-extractor.ts
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ModelInfo } from '/@/api/model-info';
+import type { RamalamaContainerInfo } from '/@/api/ramalama-container-info';
+
+export class ContainerModelExtractor {
+  async extractModelInfo(container: RamalamaContainerInfo): Promise<ModelInfo> {
```
simonrey1, line 23: suggestion(non-blocking): this method can be static

extensions/ramalama/src/helper/container-model-extractor.ts
benoitf: the bot said something but either the label is there with a port or the label is not there so looks like a false positive (not an issue)

extensions/ramalama/src/helper/container-model-extractor.ts
benoitf: -1, DI here

benoitf:
@simonrey1 it searches for containers, container engine requirements added

PR #692, opened 2025-11-07
chore: introduce skeleton for RamaLama extension

benoitf:
introduce a skeleton for RamaLama extension

it does nothing, but expect to get the container extension being available

- [x] depends on https://github.com/kortex-hub/kortex/pull/691

### DIFF
--- a/extensions/ramalama/src/manager/inference-model-manager.ts
+++ b/extensions/ramalama/src/manager/inference-model-manager.ts
@@ -17,7 +17,9 @@
  ***********************************************************************/
 
 import type { Disposable } from '@kortex-app/api';
+import { injectable } from 'inversify';
 
+@injectable()
 export class InferenceModelManager implements Disposable {
   async init(): Promise<void> {}
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -511,9 +511,6 @@ importers:
 
   extensions/ramalama:
     dependencies:
-      '@ai-sdk/openai-compatible':
-        specifier: ^1.0.19
-        version: 1.0.26(zod@4.1.12)
       inversify:
         specifier: ^7.7.1
         version: 7.7.1(reflect-metadata@0.2.2)
@@ -524,18 +521,12 @@ importers:
       '@kortex-app/container-extension-api':
         specifier: workspace:*
         version: link:../container/packages/api
-      '@types/dockerode':
-        specifier: ^3.3.42
-        version: 3.3.42
       adm-zip:
         specifier: ^0.5.16
         version: 0.5.16
       mkdirp:
         specifier: ^3.0.1
         version: 3.0.1
-      msw:
-        specifier: ^2.10.4
-        version: 2.10.4(@types/node@24.1.0)(typescript@5.9.2)
       vite:
         specifier: ^7.1.2
         version: 7.1.2(@types/node@24.1.0)(jiti@2.4.2)(lightningcss@1.30.1)(terser@5.36.0)(tsx@4.20.3)(yaml@2.8.1)
@@ -8218,6 +8209,7 @@ snapshots:
       '@inquirer/type': 3.0.7(@types/node@24.1.0)
     optionalDependencies:
       '@types/node': 24.1.0
+    optional: true
 
   '@inquirer/core@10.1.14(@types/node@22.17.0)':
     dependencies:
@@ -8244,6 +8236,7 @@ snapshots:
       yoctocolors-cjs: 2.1.2
     optionalDependencies:
       '@types/node': 24.1.0
+    optional: true
 
   '@inquirer/figures@1.0.12': {}
 
@@ -8254,6 +8247,7 @@ snapshots:
   '@inquirer/type@3.0.7(@types/node@24.1.0)':
     optionalDependencies:
       '@types/node': 24.1.0
+    optional: true
 
   '@internationalized/date@3.8.2':
     dependencies:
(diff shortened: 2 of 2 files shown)


simonrey1 (approved):
LGTM, after a rebase of PR 691, I was able to start and stop RamaLama extension without errors in the interface

Remark: `2. Models will automatically appear in Kortex's models` in the extension details page is not yet true because this is only a skeletton, but I guess it is fine to wait for the implementation.

extensions/ramalama/src/manager/_manager-module.ts
axel7083, line 1: why using an underscore for the filename?

extensions/ramalama/src/manager/inference-model-manager.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+export class InferenceModelManager {
```
axel7083: can't we implements the Disposable from `@kortex/extension-api` ?

axel7083 (commented):
I rebased this PR onto https://github.com/kortex-hub/kortex/pull/691 to try it

I can confirm on windows, the ramalama extension is not failing 

Image

extensions/ramalama/src/manager/_manager-module.ts
benoitf: it was a way to declare the inversify modules at the top of the directory (_ was for something not really related to the code, so having it always as the first entry in the folder was giving a quick introduction of)

I was tempted to always name it `_module.ts` (it's like metadata for the current folders)

extensions/ramalama/src/manager/inference-model-manager.ts
benoitf: it's satisfying the Disposable pattern with the dispose method (not required to implement it like in Java)

extensions/ramalama/src/manager/_manager-module.ts
axel7083: maybe we should create a dedicated folder, instead of using some weird naming? we created an `inject` folder in the podman extension, make sense to me, especially if we want to have separation of concern

benoitf: usually you declare bindings in the folders where the files are and then you include the module in the upper directories

so you would create an inject folder in each folder ?

axel7083: You are saying that now with inversify we have to do the binding in separate files, that would not be centralised ?

Moreover this feel very globalish and not testable, in the contrary of the podman extension where we introduce a class `InversifyBinding` responsible of binding all the services

axel7083: I don't understand why we are doing both here (Having an InversifyBinding & `_module` files), when we could have a single place doing the logic, this is new and may be the "way" that inversify recommend, but this is not the way we dealt with it in podman extension and I would rather have code guidelines, we should not define or decide new way of doing things in PR

extensions/ramalama/src/manager/inference-model-manager.ts
axel7083: For readability, and have a better understanding of the classes lifecycle, I would rather see directly when opening a file that a class is disposable than finding in the middle of other functions that there is a dispose one.

extensions/ramalama/src/manager/_manager-module.ts
benoitf: it's not globalish or not testable. It's how you better scope things

as written at
https://inversify.io/docs/api/container-module/

> Container modules help manage the complexity of bindings in large applications.

all extensions using currently inversify (and all large scale projects) are using modules.

> I would rather have code guidelines, we should not define or decide new way of doing things in PR

I could argue that the way podman extension introduce it is not the way all other usage is doing
or that this extension existed before "podman extension" was trying to use inversify ?

extensions/ramalama/src/manager/inference-model-manager.ts
benoitf: added

extensions/ramalama/src/inject/inversify-binding.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+let inversifyBinding: InversifyBinding;
+
+const containerExtensionAPIMock = {} as unknown as ContainerExtensionAPI;
+const extensionContextMock = {} as ExtensionContext;
+const telemetryLoggerMock = {} as TelemetryLogger;
+const providerMock = {} as Provider;
+
+// mock inversify
+vi.mock(import('inversify'));
```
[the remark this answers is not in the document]
benoitf: we're using this syntax to be sure that in case we move/remove the string specified inside the mock still exists

for example vi.mock(import('../foo.js')); will fail if foo.js is moved

extensions/ramalama/src/manager/_manager-module.ts
benoitf: will defer to a possible standard for naming modules

benoitf:
PR rebased and updated

sill do nothing

MarsKubeX (approved):
LGTM

Issue #490, opened 2025-10-01
display context/size tokens used by a query in chat


benoitf:
Might be interesting to know the "resources" consumed by my query

For example: one query without MCPs or with multiple MCPs, all the tools from a MCP or a subset of tools, etc.

so for example between two queries looking almost similar, there could be a huge difference

slemeur:
Done <img width="963" height="200" alt="Image" src="https://github.com/user-attachments/assets/8ea8944f-37a3-42d2-984b-b61065e9fd3b" />

PR #699, opened 2025-11-07
fix: display the number of tokens consumed in the chat UI

jeffmaury:
Relates to #490

Might request reset of the chat database as a new field is added into the Message table

tokens


packages/main/src/chat/db/migrations/0000_ambiguous_dragon_man.sql
```diff
@@ -25,6 +25,7 @@ CREATE TABLE "Message" (
     "attachments" json NOT NULL,
     "createdAt" timestamp NOT NULL,
     "config" json NOT NULL,
+    "tokens" integer,
```
simonrey1: issue(blocking): we cannot force users to reset the database, you can alter the table in a new migration file

benoitf: or just use a new filename to store the database by incrementing the version as I would say it's still okish to change that as there is no 'stable' release yet.

packages/renderer/src/lib/chat/components/ipc-chat-transport.ts
```diff
@@ -47,9 +51,10 @@ export class IPCChatTransport<T extends UIMessage> implements ChatTransport<T> {
             console.error('Error during inferenceStreamText:', error);
             controller.error(error);
           },
-          () => {
-            console.log('IPCChatTransport: Stream completed');
+          (tokens: number) => {
+            console.log('IPCChatTransport: Stream completed tokens:', tokens);
```
simonrey1: nitpick: add a space before the number of tokens (or use string interpolation to do it)

packages/main/src/chat/chat-manager.ts
```diff
@@ -251,6 +253,7 @@ export class ChatManager {
               chatId,
               attachments: [],
               config,
+              tokens: totalUsage.totalTokens ?? null,
```
benoitf: we should use undefined, not null

benoitf:
removed~/.local/share/kortex/chat-persistence/chats.db

now I see the tokens value.

I have then a question around the location of the tokens display

here it's at the bottom right, I do one query and it display the tokens

now, I ask a new query and it's changing the tokens value to the last query.

then, it means I don't know if tokens value is for the last query, the sum of all queries and then I don't see the tokens used for my previous query (if I want to compare)

packages/main/src/chat/chat-manager.ts
jeffmaury: > we should use undefined, not null

Can use undefined here as the type is infered by drizzle and has the number | null type

benoitf: can't it be not null and default is 0 ?

benoitf:
> The tokens displayed here is for the whole conversation. That's per design. We could display next to each response as it's attached to Message table but wondering about the UI

IMHO it's what is confusing. underlying arch is providing that information on each response not for the full chat.

I would expect on my end to see at the bottom the total of token consumed (sum of each queries, not the last one)
each time I do a query, it increments the number of tokens I have used since I created my chat.

packages/main/src/chat/db/migrations/0000_ambiguous_dragon_man.sql
jeffmaury: Fixed by 801f843ab7b8e42db1e6b670efbacfa3ca12ffb5

packages/main/src/chat/chat-manager.ts
jeffmaury: Fixed by c1970a9eec9422f2c5323a0deb93a7efc6221018

benoitf:
> But it is the case as all messages from a chat are part of a single conversation or I missed something ?

the number being displayed was not incrementing on my end. (first query was higher than the second one) using the same query prompt in the chat.

so it's for the last response

vancura:
There's not enough space for the counter badge down there. How about something like this?

image

(The badge is top right, same height as the buttons there)

/c @MariaLeonova

packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -51,6 +51,14 @@ let selectedMCP = $state<MCPRemoteServerInfo[]>(
 
 const chatHistory = ChatHistory.fromContext();
 
+let totalTokens = $state(
+  messages[messages.length - 1] !== undefined &&
+    typeof messages[messages.length - 1] === 'object' &&
```
simonrey1: question: why is all the typecheck needed?

Would it be enough to go with?

```ts
let totalTokens = $state(messages[messages.length - 1].tokens);
```

packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -106,17 +117,24 @@ const hasModels = $derived(models && models.length > 0);
...
                     loading={chatClient.status === 'streaming' || chatClient.status === 'submitted'}
                     messages={chatClient.messages}
                 />
-                <form class="bg-background mx-auto flex w-full gap-2 px-4 pb-4 md:max-w-3xl md:pb-6">
+                <form class="bg-background mx-auto flex w-full gap-2 px-4 md:max-w-3xl">
                     {#if !readonly}
                         <MultimodalInput {attachments} {chatClient} {selectedModel} {selectedMCP} bind:mcpSelectorOpen={mcpSelectorOpen} class="flex-1" />
                     {/if}
                 </form>
+                {#if !readonly}
```
simonrey1: suggestion(non-blocking): introduce a small Svelte component for the token count

packages/renderer/src/lib/chat/components/ipc-chat-transport.ts
```diff
@@ -47,9 +51,10 @@ export class IPCChatTransport<T extends UIMessage> implements ChatTransport<T> {
             console.error('Error during inferenceStreamText:', error);
             controller.error(error);
           },
-          () => {
-            console.log('IPCChatTransport: Stream completed');
+          (tokens: number) => {
+            console.log('IPCChatTransport: Stream completed tokens: ', tokens);
```
simonrey1, line 55: suggestion(non-blocking): String interpolation can also be used here

simonrey1 (approved):
LGTM
Image

I left some comments but they are not blocking

packages/renderer/src/lib/chat/components/chat.svelte
jeffmaury: Fixed by bd8604b1d5459cc7cda80a99e9c517a050dbb9e6

packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -51,6 +51,8 @@ let selectedMCP = $state<MCPRemoteServerInfo[]>(
 
 const chatHistory = ChatHistory.fromContext();
 
+let totalTokens = $state(messages[messages.length - 1]?.tokens ? messages[messages.length - 1].tokens : 0);
```
benoitf: $derived

jeffmaury: Cannot use $derived as the messages array is not updated but client.messages

packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -101,18 +106,16 @@ const hasModels = $derived(models && models.length > 0);
...
         {#if hasModels}
-            <div class="flex flex-col flex-3/4"> 
+            <div class="flex flex-col flex-3/4">
+              <ChatHeader bind:mcpSelectorOpen={mcpSelectorOpen} {readonly} models={models} bind:selectedModel={selectedModel} bind:selectedMCP={selectedMCP} tokens={totalTokens}/>
                 <Messages
                     {readonly}
                     loading={chatClient.status === 'streaming' || chatClient.status === 'submitted'}
                     messages={chatClient.messages}
                 />
-                <form class="bg-background mx-auto flex w-full gap-2 px-4 pb-4 md:max-w-3xl md:pb-6">
```
gastoner, line 155: I think that shouldnt be removed

packages/renderer/src/lib/chat/components/chat-header.svelte
```diff
@@ -86,8 +88,11 @@ const noMcps = $derived($mcpRemoteServerInfos.length === 0);
                 </div>
             {/if}
         </div>
+    <div class="order-3 ml-auto bg-muted flex justify-end gap-4 rounded-lg py-2 text-sm text-muted-foreground">
```
gastoner: py-2 => to p-2

gastoner: With py-2
image

With p-2
image

packages/renderer/src/lib/chat/components/chat.svelte
gastoner: Without this change
image

With this change:
image

jeffmaury: Fixed by commit f7cb8a827f4550c11ac4290286f1c23d6cb76973

packages/renderer/src/lib/chat/components/chat-header.svelte
jeffmaury: Fixed by 611875bdd7e608cdf05865e66eed276f48a2f7ba

gastoner (approved):
LGTM, thanks for the changes

fbricon:
Closing as the Chat view is being removed #2853

Issue #518, opened 2025-10-06
Executing a flow with Input Fields


slemeur:
* Now that a flow can have input fields, the user who is executing the flow will have to provide them.   
* To begin with (and in the interest of keeping things relatively simple at the moment), when the user will be starting a flow, the user will see a modal with a way to provide the values for the flow.   
  * Only when the “required” input fields will be filled, the user will be able to start the flow.

PR #715, opened 2025-11-10
feat(flow): allow passing parameters to flow execution

simonrey1:
**Partial fix of #518**

This PR implements the requested feature to allow users to pass parameters when executing a flow. This is crucial for flows that require dynamic input at runtime.

### Changes Included:

1.  **`packages/extension-api/src/extension-api.d.ts`**: The `execute` method in `FlowAPI` now accepts an optional `params?: Record<string, string>` argument.
2.  **`packages/main/src/plugin/flow/flow-manager.ts`**:
    * Updated the internal `execute` method signature to accept the `params`.
    * Passed the `params` through to the flow provider's `flowConnection.flow.execute`.
    * The IPC handler for `flows:execute` is updated to include `params` in the flow payload and passes them along.
3.  **`packages/preload/src/index.ts`**: The `flowExecute` exposed method is updated to accept the optional `params` in the flow object payload.
4.  **`extensions/goose/src/goose-cli.ts`**:
    * The `generateCommandLine` and `run` methods are updated to accept and handle the optional `params`.
    * The `params` are converted into the necessary `--params key=value` command-line arguments for the underlying `goose` CLI execution.

This ensures the parameter passing is supported end-to-end, from the API call down to the CLI execution.

You can see my WIP here for the modal:

https://github.com/simonrey1/kortex/compare/flows/execute/add-params...simonrey1:kortex:feature/svelte-5-runes-modal?expand=1

### DIFF
--- a/extensions/goose/src/goose-cli.ts
+++ b/extensions/goose/src/goose-cli.ts
@@ -88,12 +88,14 @@ export class GooseCLI implements Disposable {
     options: { path: string; env?: Record<string, string>; params?: Record<string, string> },
   ): FlowGenerateCommandLineResult {
     if (!this.cli?.path) throw new Error('goose not installed');
-    const args = ['run', '--recipe', flowPath];
-    if (options.params) {
-      for (const key in options.params) {
-        args.push('--params', `${key}=${options.params[key]}`);
-      }
-    }
+    const args = [
+      'run',
+      '--recipe',
+      flowPath,
+      ...(options.params
+        ? Object.entries(options.params).flatMap(([key, value]) => ['--params', `${key}=${value}`])
+        : []),
+    ];
     return {
       command: this.cli.path,
       args,


axel7083 (approved):
Codewise LGTM

packages/extension-api/src/extension-api.d.ts
```diff
@@ -497,7 +497,7 @@ declare module '@kortex-app/api' {
       /**
        * @experimental expect change
        */
-      execute(flowId: string, logger: Logger): Promise<void>;
+      execute(flowId: string, logger: Logger, params?: Record<string, string>): Promise<void>;
```
benoitf, line 500: suggestion: prefer FlowOptions interface, that will allow to add more parameters later. Here we need to change the method signature

extensions/goose/src/goose-cli.ts
```diff
@@ -85,17 +85,27 @@ export class GooseCLI implements Disposable {
...
-    options: { path: string; env?: Record<string, string> },
+    options: { path: string; env?: Record<string, string>; params?: Record<string, string> },
   ): FlowGenerateCommandLineResult {
     if (!this.cli?.path) throw new Error('goose not installed');
+    const args = ['run', '--recipe', flowPath];
+    if (options.params) {
+      for (const key in options.params) {
+        args.push('--params', `${key}=${options.params[key]}`);
+      }
+    }
```
benoitf: suggestion
```
const args = [
  'run',
  '--recipe',
  flowPath,
  ...(options.params
    ? Object.entries(options.params).flatMap(([key, value]) => ['--params', `${key}=${value}`])
    : [])
];

```

basically iterate on entries (that gives you both key and value items) rather than key to then get the value using the key

packages/extension-api/src/extension-api.d.ts
simonrey1: As discussed online, because the FlowOptions interface still needs to be created, I will follow-up this PR with a proposition to introduce it

simonrey1: https://github.com/kortex-hub/kortex/pull/725

Issue #476, opened 2025-09-29
should be able to select tools we want to include from a given mcp server


benoitf:
having many tools can increase the context size a lot

having a way to restrict the tools that we want to enable will decrease the context size and improve performance on local inference servers

another advantage is to exclude tools that could do operations we don't want

It would mean that the predefined GitHub query should probably require a given tool(or tools) from all available from a given MCP server rather than requiring all of them

PR #716, opened 2025-11-10
feat: filter mcp tools

axel7083:
## Description

I tried to followed the mockup in https://github.com/kortex-hub/kortex/issues/476#issuecomment-3512820117 and the instructions in https://github.com/kortex-hub/kortex/issues/476#issuecomment-3574260046

## Screenshots

https://github.com/user-attachments/assets/a16538f7-9ef8-464b-b705-6fd90c1c8d1a

## Related issues

Fixes https://github.com/kortex-hub/kortex/issues/476

## Testing

**Manual reproducer**

1. checkout this branch
2. assert you have the GitHub server
3. open the right panel on the chat interface
4. assert you see the GitHub mcp
5. click on the `filter` icon
6. assert you see all tools of the GitHub MCP
7. select the `list_issues`
8. ask the bot to list issues in Podman Desktop
9. assert it works

### DIFF
--- a/packages/renderer/src/lib/chat/components/mcp-filter-card.svelte
+++ b/packages/renderer/src/lib/chat/components/mcp-filter-card.svelte
@@ -1,8 +1,5 @@
 <script lang="ts">
-import { faChevronDown } from '@fortawesome/free-solid-svg-icons/faChevronDown';
-import { faChevronLeft } from '@fortawesome/free-solid-svg-icons/faChevronLeft';
 import { Checkbox } from '@podman-desktop/ui-svelte';
-import Fa from 'svelte-fa';
 
 import { Tooltip, TooltipContent, TooltipTrigger } from '/@/lib/chat/components/ui/tooltip';
 import type { MCPRemoteServerInfo } from '/@api/mcp/mcp-server-info';
@@ -31,23 +28,13 @@ let selection: 'all' | 'none' | 'partial' = $derived.by(() => {
   else if (selectedTools.size === tools.length) return 'all';
   return 'partial';
 });
-let expanded: boolean = $state(false);
-
-function toggleExpand(): void {
-  expanded = !expanded;
-}
 </script>
 
 <div
-  class="flex flex-col items-center flex justify-between rounded-md"
-  class:border={expanded}
-  class:border-[var(--pd-content-table-border)]={expanded}
+  class="flex flex-col items-center flex justify-between rounded-md border border-[var(--pd-content-table-border)]"
 >
   <div
     class="flex flex-row bg-[var(--pd-content-card-hover-bg)] py-5 px-2 w-full rounded-tr-md rounded-tl-md justify-between"
-    class:rounded-b-md={!expanded}
-    class:border={!expanded}
-    class:border-[var(--pd-content-table-border)]={!expanded}
   >
     <div
       class="flex gap-x-2">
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/chat/components/chat.svelte
```diff
@@ -49,6 +50,10 @@ let selectedMCP = $state<MCPRemoteServerInfo[]>(
   config?.mcp?.flatMap(mcpId => $mcpRemoteServerInfos.find(r => r.id === mcpId) ?? []) ?? [],
 );
 
+let selectedMCPTools = $state<Map<string, Set<string>>>(
```
gastoner: Use SvelteMap?

gastoner (commented):
https://github.com/user-attachments/assets/c338bd31-947c-4a3f-b0a0-2923cd76d2c8

When I select some tool and then hide/show the filter mine pick was cleared, is this expected? I think that when I prompt it remembered my picks

packages/main/src/plugin/mcp/mcp-manager.ts
```diff
@@ -75,23 +75,45 @@ export class MCPManager implements IAsyncDisposable {
...
-  public async getToolSet(selected: Array<string> | undefined = undefined): Promise<ToolSet> {
-    const tools = await Promise.all(
-      (
-        selected?.reduce(
-          (accumulator, current) => {
-            const client = this.#client.get(current);
-            if (client) {
-              accumulator.push(client);
+  public async getToolSet(
+    selectedMCP: Array<string> | undefined = undefined,
```
MarsKubeX: Should not be better as optional parameter?

axel7083: Currently the function takes `selectedMCP: Array<string> | undefined = undefined,` and I just copy paste, I could change to `options?: { ... }` if this is important for you 👍

packages/renderer/src/lib/chat/components/chat.svelte
axel7083: Good catch!

axel7083:
> When I select some tool and then hide/show the filter mine pick was cleared, is this expected? I think that when I prompt it remembered my picks

yes I did that on purpose, it makes sense to me to reset, but might not, we can probably work on that in a follow-up

benoitf (commented):
I was able to filter tools. Funny thing is that if you unselect list_issues it's still returning data (but in that case you don't see that the tool has been called)

I still have questions on:
- lag when clicking on the filter buttons. (I need to wait for quite some time)
- colors of the button (red color when it's opened)

image

- chevron direction:  here it's going to the up direction (so it's yet another direction when it's opened)
https://github.com/podman-desktop/podman-desktop/issues/14165

- filter button displays : "filter toggled tools" on hover/tooltip


- tools is replacing MCP on the right. should it be a replacement or an addition 
image

like there is the MCP possibility and then Tools (like a right toolbar) (not clear from my side from the issue or mockup to see what is expected)

now (MCP is no longer there)
image

benoitf:
in the mockup I saw that for each tool we're displaying the description of the tool (it's not implemented here). Are we missing that information ?

benoitf:
and I don't see filtering box to filter for example all the tools having the name "issue" in it

axel7083:
Following private discussions

- The chevron should be inverted
- We should store the information about the list of tools of all the MCP
  - This imply that the filter buttons will be removed by a chevron as the info will be available
- We should have a search bar
- The checkbox should be improved
  - Checkbox of the MCP should be selected when all tools are selected
  - Checkbox of the MCP should be indeterminate if a few tools are selected
  - Checkbox of the MCP should be unchecked if none are selected
- Description of each tool should be visible
- The auto-select tools will not be included

packages/api/src/chat/InferenceParameters.ts
```diff
@@ -22,6 +22,10 @@ export interface InferenceParameters {
   providerId: string;
   connectionName: string;
   modelId: string;
-  mcp: string[];
+  mcp: Array<string>;
+  /**
+   * The key is the MCP ID, and the value is an array of tool to use for that MCP.
+   */
+  tools: Record<string, Array<string>>;
```
simonrey1, line 25: suggestion(non-blocking): merge `mcp` and `tools` to simplify

axel7083: This has more implication that it seems 👁️

axel7083: https://github.com/kortex-hub/kortex/pull/716/commits/26f5f624b6ed1fdc2099a452f65034a091dfb9f3

packages/preload/src/index.ts
```diff
@@ -1736,13 +1736,6 @@ export function initExposure(): void {
     return ipcInvoke('mcp-registry:getMcpRegistryServers');
   });
 
-  contextBridge.exposeInMainWorld(
```
axel7083, line 1739: this was not used, but was using stuff preventing refactoring

packages/main/src/plugin/index.ts
```diff
@@ -2169,22 +2169,6 @@ export class PluginSystem {
       },
     );
 
-    this.ipcHandle(
```
axel7083, line 2172: Removing as not used https://github.com/kortex-hub/kortex/pull/716/files#r2560672845 but preventing refactoring

axel7083:
> two chevrons do not like to be in their expected state

I removed the collapsing mechanism for the MCP tools.

https://github.com/user-attachments/assets/0b7727ae-bfe8-474a-9d9e-14bddfff6a6d

> issue: if I click on the card 'Give me last 5 issues' there is nothing being reported in the UI but in the console I see

I cannot reproduce, see above recording

MarsKubeX (approved):
LGTM in MacOS. The only thing I see that could be a weird behavior is after clicking "what are the last 5 issues of Github" after retrieving the issues, the "tools section" at the right is closed but not-blocking.

https://github.com/user-attachments/assets/418e0bd2-932c-4d7a-860c-dd884e781b8c

PR #717, opened 2025-11-10
feat(loader): add the loader SVG

### LINKED ISSUE #368: loader image is the seal on a rocket
Image

should be changed to something with 

Image

vancura:
This adds the loader SVG.

Fixes #368.

Co-Authored-By Claude

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -55,7 +55,7 @@
     "yaml": "^2.8.1"
   },
   "dependencies": {
-    "@ai-sdk/svelte": "^4.0.28",
+    "@ai-sdk/svelte": "^4.0.31",
     "@ai-sdk/ui-utils": "^1.2.11",
     "@lucide/svelte": "^0.539.0",
     "@podman-desktop/ui-svelte": "^1.22.0-202509181129-3b9dd05d418",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -645,8 +645,8 @@ importers:
   packages/renderer:
     dependencies:
       '@ai-sdk/svelte':
-        specifier: ^4.0.28
-        version: 4.0.28(svelte@5.38.1)(zod@4.1.13)
+        specifier: ^4.0.31
+        version: 4.0.31(svelte@5.38.1)(zod@4.1.13)
       '@ai-sdk/ui-utils':
         specifier: ^1.2.11
         version: 1.2.11(zod@4.1.13)
@@ -845,12 +845,6 @@ packages:
   '@adobe/css-tools@4.4.3':
     resolution: {integrity: sha512-VQKMkwriZbaOgVCby1UDY/LDk5fIjhQicCvVPFqfe+69fWaPWydbWJ3wRt59/YzIwda1I81loas3oCoHxnqvdA==}
 
-  '@ai-sdk/gateway@3.0.11':
-    resolution: {integrity: sha512-gLrgNXA95wdo/zAlA0miX/SJEYKSYCHg+e0Y/uQeABLScZAMjPw3jWaeANta/Db1T4xfi8cBvY3nnV8Pa27z+w==}
-    engines: {node: '>=18'}
-    peerDependencies:
-      zod: ^3.25.76 || ^4.1.8
-
   '@ai-sdk/gateway@3.0.13':
     resolution: {integrity: sha512-g7nE4PFtngOZNZSy1lOPpkC+FAiHxqBJXqyRMEG7NUrEVZlz5goBdtHg1YgWRJIX776JTXAmbOI5JreAKVAsVA==}
     engines: {node: '>=18'}
@@ -887,12 +881,6 @@ packages:
     peerDependencies:
       zod: ^3.25.76 || ^4.1.8
 
-  '@ai-sdk/provider-utils@4.0.4':
-    resolution: {integrity: sha512-VxhX0B/dWGbpNHxrKCWUAJKXIXV015J4e7qYjdIU9lLWeptk0KMLGcqkB4wFxff5Njqur8dt8wRi1MN9lZtDqg==}
-    engines: {node: '>=18'}
-    peerDependencies:
-      zod: ^3.25.76 || ^4.1.8
-
   '@ai-sdk/provider-utils@4.0.5':
     resolution: {integrity: sha512-Ow/X/SEkeExTTc1x+nYLB9ZHK2WUId8+9TlkamAx7Tl9vxU+cKzWx2dwjgMHeCN6twrgwkLrrtqckQeO4mxgVA==}
     engines: {node: '>=18'}
@@ -911,8 +899,8 @@ packages:
     resolution: {integrity: sha512-HrEmNt/BH/hkQ7zpi2o6N3k1ZR1QTb7z85WYhYygiTxOQuaml4CMtHCWRbric5WPU+RNsYI7r1EpyVQMKO1pYw==}
     engines: {node: '>=18'}
 
-  '@ai-sdk/svelte@4.0.28':
-    resolution: {integrity: sha512-4Pag4Ek7vIH8fuehKZD2Yfb3oGe5sL2t3oLQ9eVggqFQ9xBELTLLQzUD2+CSEXH61NavGPq4ve2rG+znvGzLpg==}
+  '@ai-sdk/svelte@4.0.31':
+    resolution: {integrity: sha512-p+VfBQ+YGt/VxyrZZ1mKg0ArfP7/28CeTgSiPZgjbXEwhtejWgsdu1nEqkrM8wahBedFNEfuY/8b41Efxvst7Q==}
     peerDependencies:
       svelte: ^5.31.0
 
@@ -2895,12 +2883,6 @@ packages:
   aggregate-error@3.1.0:
     resolution: {integrity: sha512-4I7Td01quW/RpocfNayFdFVk1qSuoh0E7JrbRJ16nH01HhKFQ88INq9Sd+nd72zqRySlr9BmDA8xlEJ6vJMrYA==}
     engines: {node: '>=8'}
-
-  ai@6.0.28:
-    resolution: {integrity: sha512-TshF7p7Vhg8Ty8MMijiUKB3k334GTqgD7eK5XQzaxAcrc4JXuZClRrS0nhxrbs0LOtCVbqYTRLb7qJ7VG2djnQ==}
-    engines: {node: '>=18'}
-    peerDependencies:
-      zod: ^3.25.76 || ^4.1.8
 
   ai@6.0.31:
     resolution: {integrity: sha512-aAn62jsDueAK7oiY4jeqJcA4zFctDqVHGyEaUDaWxEXzz4kbMdoByfYlYZhO1V3nhkeVoI8qNyFfiZusAubQLQ==}
@@ -7835,19 +7817,19 @@ snapshots:
 
   '@adobe/css-tools@4.4.3': {}
 
-  '@ai-sdk/gateway@3.0.11(zod@4.1.13)':
-    dependencies:
-      '@ai-sdk/provider': 3.0.2
-      '@ai-sdk/provider-utils': 4.0.4(zod@4.1.13)
-      '@vercel/oidc': 3.1.0
-      zod: 4.1.13
-
   '@ai-sdk/gateway@3.0.13(zod@4.1.11)':
     dependencies:
       '@ai-sdk/provider': 3.0.2
       '@ai-sdk/provider-utils': 4.0.5(zod@4.1.11)
       '@vercel/oidc': 3.1.0
       zod: 4.1.11
+
+  '@ai-sdk/gateway@3.0.13(zod@4.1.13)':
+    dependencies:
+      '@ai-sdk/provider': 3.0.2
+      '@ai-sdk/provider-utils': 4.0.5(zod@4.1.13)
+      '@vercel/oidc': 3.1.0
+      zod: 4.1.13
 
   '@ai-sdk/google@3.0.7(zod@4.3.5)':
     dependencies:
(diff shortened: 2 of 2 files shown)


benoitf:
could you please avoid using AI to generate comments

image

vancura:
Do you have technical, performance, or architectural concerns with embedded PNG data? I'm here to discuss the actual implementation.

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,136 @@
+<script lang="ts">
+import { onMount } from 'svelte';
+
+export let size = 400;
+
+let dotsGroup: SVGGElement | null = null;
+let eyesElement: SVGUseElement | null = null;
```
benoitf: issue: null should not be used as per https://github.com/kortex-hub/kortex/blob/main/eslint.config.mjs#L30 rule

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,136 @@
...
+      dotsGroup.appendChild(circle);
+    }
+  }
+
+  function blink(): void {
+    if (eyesElement) {
+      eyesElement.style.transition = 'opacity 0.05s ease-in';
+      eyesElement.style.opacity = '1';
+
+      setTimeout(() => {
```
benoitf, line 77: issue: timeout should be cleared when the element is destroyed

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,136 @@
...
+        if (eyesElement) {
+          eyesElement.style.transition = 'opacity 0.15s ease-out';
+          eyesElement.style.opacity = '0';
+        }
+      }, 80);
+
+      // eslint-disable-next-line sonarjs/pseudo-random
+      setTimeout(blink, 500 + Math.random() * 2000);
```
benoitf: suggestion: numbers > 1000 are written using the format 1_000, 2_000 in the codebase

axel7083 (commented):
Giving my two cents here

I don't think we should embed base64 encoded images in svelte components; when dealing with `<image` the browser will have to allocate memory to store, and decode the base64 string before drawing it. This is CPU intensive, may be a CPU spike on startup, using pure HTML elements `<path> etc.` would avoid this step, therefore I am in favour of avoiding hardcoding images.

Possible solution

- Use only HTML elements
- Move the files in a separate folder and import them

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,136 @@
...
+import { onMount } from 'svelte';
+
+export let size = 400;
+
+let dotsGroup: SVGGElement | null = null;
+let eyesElement: SVGUseElement | null = null;
+
+onMount(() => {
+  // Configuration - scale from 800pt design
+  const scale = size / 800;
```
benoitf: suggestion: global constants are usually not in the onMount part

packages/renderer/src/lib/images/LoaderAnimation.svelte
vancura: BTW, it would be good to have this in Husky.

axel7083: I opened https://github.com/podman-desktop/podman-desktop/issues/14842 there are an existing rule that we can enforce

simonrey1:
> Does anyone know what's DCO's problem? These strings are the same!
Screenshot 2025-11-25 at 15 05 04

I copied them in VSCode and it seems that `Václav` `Vančura` strings are not the same, maybe an issue with encoding.

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,178 @@
...
+            eyesElement.style.transition = `opacity ${BLINK_FADE_OUT_S}s ease-out`;
+            eyesElement.style.opacity = '0';
+          }
+        }, BLINK_CLOSE_DELAY_MS),
+      );
+
+      // eslint-disable-next-line sonarjs/pseudo-random
+      timeoutIds.push(setTimeout(blink, BLINK_MIN_INTERVAL_MS + Math.random() * BLINK_RANDOM_INTERVAL_MS));
+    }
+  }
```
gastoner, line 89: (not-blocking) Personally, I would move this function outside of `onMount`

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,178 @@
...
+    circle.setAttribute('r', dotRadius.toString());
+    circle.setAttribute('fill', DOT_COLOR);
+
+    // eslint-disable-next-line svelte/no-dom-manipulating
+    dotsGroup.appendChild(circle);
+  }
+}
+
+onMount(() => {
+  const timeoutIds: ReturnType<typeof setTimeout>[] = [];
```
gastoner: Cant we use `NodeJS.Timeout` directly?

packages/renderer/src/lib/images/LoaderAnimation.spec.ts
```diff
@@ -0,0 +1,60 @@
+/**********************************************************************
+ * Copyright (C) 2024 Red Hat, Inc.
```
gastoner: 2025

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,184 @@
...
+$effect(() => {
+  if (dotsGroup) {
+    // Clear existing dots
+    // eslint-disable-next-line svelte/no-dom-manipulating
+    dotsGroup.innerHTML = '';
+
+    // Scale from 800pt design
+    const scale = size / BASE_DESIGN_SIZE;
+    const radius = BASE_RADIUS * scale;
+    const dotRadius = BASE_DOT_RADIUS * scale;
```
gastoner: Cant we move this outside of effect? => whis way it will become derived/ states

gastoner: This way we would not compute this every time, but only when size changes

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,189 @@
...
+
+function blink(timeoutIds: NodeJS.Timeout[]): void {
+  if (eyesElement) {
+    eyesElement.style.transition = `opacity ${BLINK_FADE_IN_S}s ease-in`;
+    eyesElement.style.opacity = '1';
+
+    timeoutIds.push(
+      setTimeout(() => {
+        if (eyesElement) {
+          eyesElement.style.transition = `opacity ${BLINK_FADE_OUT_S}s ease-out`;
```
gastoner, line 79: NIT: `opacity ${BLINK_FADE_OUT_S}s ease-out` cant be this moved to CONST on the top of the file?

gastoner: Same for the `opacity ${BLINK_FADE_IN_S}s ease-in`

packages/renderer/src/Loader.spec.ts
```diff
@@ -24,11 +24,49 @@ import { tick } from 'svelte';
...
+    ({
+      matches: false,
+      media: query,
+      onchange: null,
+      addListener: () => {},
+      removeListener: () => {},
+      addEventListener: () => {},
+      removeEventListener: () => {},
+      dispatchEvent: () => false,
+    }) as MediaQueryList;
```
gastoner, line 45: Are we using all of those props? if not we can just fill those that we uses e.g media or matches and then do `as undefined as MediaQueryList` to avoid those lines (maybe more): 
```
addListener: () => {},
removeListener: () => {},
addEventListener: () => {},
removeEventListener: () => {},
```

packages/renderer/src/Loader.spec.ts
```diff
@@ -24,11 +24,49 @@ import { tick } from 'svelte';
...
+    },
+    get length(): number {
+      return localStorageData.size;
+    },
+    key: (index: number): string | null => [...localStorageData.keys()][index] ?? null,
+  };
+
+  Object.defineProperty(globalThis, 'matchMedia', { value: matchMediaMock, writable: true });
+  Object.defineProperty(globalThis, 'localStorage', { value: localStorageMock, writable: true });
+});
```
gastoner, line 69: Does this needs to be in `hoisted`?

packages/renderer/src/Loader.spec.ts
```diff
@@ -24,11 +24,49 @@ import { tick } from 'svelte';
...
+    removeItem: (key: string): void => {
+      localStorageData.delete(key);
+    },
+    clear: (): void => {
+      localStorageData.clear();
+    },
+    get length(): number {
+      return localStorageData.size;
+    },
+    key: (index: number): string | null => [...localStorageData.keys()][index] ?? null,
```
gastoner, line 63: Cant we use `undefined` rather than `null`

packages/renderer/src/lib/images/LoaderAnimation.svelte
```diff
@@ -0,0 +1,189 @@
...
+    <g clip-path="url(#backgroundClip)">
+        <use id="shadow1Element" xlink:href="#shadow1" style="mix-blend-mode: soft-light;"/>
+        <use id="shadow2Element" xlink:href="#shadow2" style="mix-blend-mode: soft-light;"/>
+    </g>
+
+    <g id="headGroup">
+        <use xlink:href="#head"/>
+        <use id="eyesElement" bind:this={eyesElement} xlink:href="#eyes" style="opacity: 0;"/>
+    </g>
+  </svg>
```
gastoner, line 174: Thinking if we can move the svg component to another .svelte file and then just import it + add the style here too? Not sure if it is possible tbh
WDYT other reviewers?

gastoner (commented):
Left some comments

packages/renderer/src/lib/images/LoaderAnimation.svelte
vancura: Thanks for the feedback, @gastoner! I was thinking about this, but I believe the current approach is good:

1. The timing values (`BLINK_FADE_IN_S`, `BLINK_FADE_OUT_S`) are already extracted as constants at the top of the file (https://github.com/kortex-hub/kortex/pull/717/files#diff-ad2e35474aac0c43de150e6cd52b18328bda057aaa5679e0494d0d47ad82ac0eR12-R13)
2. The full transition strings are only used once each within the `blink()` function.
3. Keeping the template literal inline makes it clear what is the CSS property that is animated (and with what easing).

If we extract the entire strings like `const TRANSITION_FADE_IN = 'opacity 0.05s ease-in'`, we'll lose the flexibility to reuse the numeric timing values elsewhere if needed.

That said, if you feel strongly about it, I'm okay to make the change. Let me know! :)

packages/renderer/src/Loader.spec.ts
vancura: I considered a more minimal mock, but I kept the full interface because:

- This mock is consumed by Svelte 5 internals (`DevicePixelRatio`) and mode-watcher. We don't have full visibility into which methods they might call.
- Using `as unknown as MediaQueryList` would bypass TS type checking, which could hide issues if the `MediaQueryList` interface or its usage changes.
- The empty stubs are essentially no-cost and prevent potential runtime errors if any of these methods get called

That said, I could simplify by removing just `addListener()`/`removeListener()` as those are deprecated APIs. Would that be a reasonable middle ground?

packages/renderer/src/Loader.spec.ts
vancura: Yup, it needs to be because (according to Claude):
- `vi.hoisted()` ensures these mocks are set up before any module imports are processed. This is necessary because:
  - Svelte 5's runtime accesses `matchMedia` during module initialization (for DevicePixelRatio)
  - `mode-watcher` accesses `localStorage` during its initialization

Apparently without hoisting, the imports at the top of the file would execute first, and these libraries would get `undefined` for these browser APIs, and the tests woiuld fail.

packages/renderer/src/Loader.spec.ts
vancura: If I recall correctly, a couple of days ago, @benoitf told me to not use `undefined` and rather `null`, but I am not sure now, to be honest.

Claude tells me that `null` is correct here. The [Web Storage API specification](https://developer.mozilla.org/en-US/docs/Web/API/Storage/key) mentions that `Storage.key()` returns `null` when the index doesn't exist, not `undefined`.

Since this is a mock of the browser `localStorage` API, it should match the real behavior. I think that using `undefined` would make the mock inaccurate and could hide bugs if code depended on the `null` return value.

packages/renderer/src/lib/images/LoaderAnimation.svelte
gastoner: Thank you for the explanation, keep it as it is

packages/renderer/src/Loader.spec.ts
```diff
@@ -24,11 +24,49 @@ import { tick } from 'svelte';
...
+  };
+
+  Object.defineProperty(globalThis, 'matchMedia', { value: matchMediaMock, writable: true });
```
jeffmaury: **issue:** the window methods are mocked by default so you must use `vi.mocked(window.xx).mockXXXX` rather than `Object.defineProperty`

vancura: Fixed in 59625db74c9cff84d33ac86932af8652e14f3d9d

packages/renderer/src/lib/images/LoaderAnimation.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+  });
+
+  test('should calculate correct scale for different sizes', () => {
+    const { container } = render(LoaderAnimation, { props: { size: 800 } });
+    const svg = container.querySelector('svg');
+    expect(svg?.getAttribute('viewBox')).toBe('0 0 800 800');
+  });
+
+  test('should handle multiple blink cycles', async () => {
+    vi.useFakeTimers();
```
gastoner: Nit: move this to beafore each, and add `shouldAdvanceTime: true` -> might lead to removing some lines of code like `vi.advanceTimersByTime` ? just guessing here tbh but see 
https://github.com/podman-desktop/podman-desktop/blob/main/CODE-GUIDELINES.md#using-fake-timers-with-svelte-components

vancura: Thanks, fixed in a7c417c65723583d93a6dfdfd8161da6994ba004

gastoner (approved):
LGTM

PR #725, opened 2025-11-12
refactor(flows): use object with all options instead of multiple parameters

simonrey1:
Follow up of comment https://github.com/kortex-hub/kortex/pull/715#discussion_r2517482395,
introducing a `FlowExecuteOptions` interface to allow adding more parameters in the future.

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -431,6 +431,12 @@ declare module '@kortex-app/api' {
     flowId: string;
   }
 
+  export interface FlowExecuteParams {
+    flowId: string;
+    logger: Logger;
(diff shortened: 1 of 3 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -497,7 +503,7 @@ declare module '@kortex-app/api' {
       /**
        * @experimental expect change
        */
-      execute(flowId: string, logger: Logger, params?: Record<string, string>): Promise<void>;
+      execute(options: FlowExecuteOptions): Promise<void>;
```
benoitf: suggestion: might be FlowExecuteParams if you replace all the parameters by the interface and not only the 'optional' part (previous last parameter)

PR #726, opened 2025-11-12
chore(container-extension): add docker finder of socket paths for macOS

benoitf:
Adds a way to detect docker engine on macOS

unit tests added

### DIFF
--- a/extensions/container/packages/extension/src/helper/socket-finder/docker/docker-macos-finder.ts
+++ b/extensions/container/packages/extension/src/helper/socket-finder/docker/docker-macos-finder.ts
@@ -1,0 +1,39 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


extensions/container/packages/extension/src/helper/socket-finder/docker/docker-macos-finder.ts
```diff
@@ -0,0 +1,39 @@
...
+export class DockerSocketMacOSFinder implements SocketFinder {
+  async findPaths(): Promise<string[]> {
+    // default socket path is at $HOME/.docker/run/docker.sock
+    const socketPath = resolve(homedir(), '.docker/run/docker.sock');
+
+    // exists ?
+    if (!existsSync(socketPath)) {
+      return [];
+    }
+    return [socketPath];
```
gastoner, line 37: I think that resolve return only `string` 
So I don't understand why we would return it in the array

benoitf: the interface implemented by various finders allows to return multiple path.

while for now it returns only one path for docker, it could return more than one. (default path for example)

I pickup a simple interface returning only an array. I could have pickup both string or Array signature but it was looking overcomplicated for such a simple case

PR #729, opened 2025-11-13
chore: update publish step to use Node.js v24

benoitf:
OIDC publication requires npm > 11.5.1 while Node.js v22 has 10.9.2

https://docs.npmjs.com/trusted-publishers
image


jeffmaury (approved):
LGTM

benoitf:
it may still not work but at least this requirement will be met

PR #730, opened 2025-11-13
feat(flow): add modal for params

simonrey1:
This PR adds a modal before running a flow locally when it has parameters. The modal displays for each parameters:
- name
- description
- an input to fill it
- the default value of the input is set if given in the flow definition

The Cancel button closes the modal, the Run button is disabled if one of the fields is empty, and it can be clicked otherwise.

When no parameters, we keep the same behavior as before : no modal, the flow runs.

I have not added UT yet but they can be added after getting some feedback here.

Closes https://github.com/kortex-hub/kortex/issues/518

Screenshot 2025-11-13 at 11 57 02
Screenshot 2025-11-13 at 11 57 04
Screenshot 2025-11-13 at 11 57 10
Screenshot 2025-11-13 at 11 56 58

Example yaml for Goose recipe:

```yaml
title: flow-pr
name: flow-pr
description: Analyse a pr

prompt: |
     Your job is to analyse and explain PR {{ pr }} of repo {{ repo }}
instructions: Your job is to analyse and explain a PR

parameters:
  - key: pr
    input_type: string
    requirement: required
    description: name of the pull request
  - key: repo
    input_type: string
    requirement: optional
    default: podman-desktop/podman-desktop
    description: name of the repo


extensions:
  - name: "com.github.mcp"
    type: "streamable_http"
    uri: "GitHub's official MCP Server"
    headers:
      Authorization: "Bearer ***************************"


settings:
  goose_provider: google
  goose_model: gemini-flash-latest

```

### DIFF
--- a/packages/renderer/src/lib/flows/FlowActions.svelte
+++ b/packages/renderer/src/lib/flows/FlowActions.svelte
@@ -8,6 +8,7 @@ import type { FlowInfo } from '/@api/flow-info';
 import { NavigationPage } from '/@api/navigation-page';
 
 import ListItemButtonIcon from '../ui/ListItemButtonIcon.svelte';
+import RunFlowModal from './RunFlowModal.svelte';
 
 interface Props {
   object: FlowInfo;
@@ -17,6 +18,7 @@ interface Props {
 const { object, onLocalRun }: Props = $props();
 
 let loading: boolean = $state(false);
+let showModal: boolean = $state(false);
 
 function navigateToRunFlow(): void {
   handleNavigation({
@@ -29,15 +31,32 @@ function navigateToRunFlow(): void {
   });
 }
 
-async function executeFlow(): Promise<void> {
+async function executeFlow(params?: Record<string, string>): Promise<void> {
   // execute the flow
-  const flow = { providerId: object.providerId, connectionName: object.connectionName, flowId: object.id };
-  const taskId: string = await window.flowExecute(flow);
+  const flow = { providerId: object.providerId, connectionName: object.connectionName, flowId: object.id, params };
+  const taskId: string = await window.flowExecute($state.snapshot(flow));
   // execute callback if any
   onLocalRun?.(taskId);
 
   // redirect to the run tab
   navigateToRunFlow();
+}
+
+async function handleRunClick(): Promise<void> {
+  if (object.parameters && object.parameters.length > 0) {
+    showModal = true;
+  } else {
+    await executeFlow({});
+  }
+}
+
+async function handleModalRun(params: Record<string, string>): Promise<void> {
+  showModal = false;
+  await executeFlow(params);
+}
+
+function handleModalCancel(): void {
+  showModal = false;
 }
 
 async function deleteFlow(): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/renderer/src/lib/flows/FlowActions.svelte
```diff
@@ -55,5 +74,14 @@ async function deleteFlow(): Promise<void> {
 }
 </script>
 
-<ListItemButtonIcon title="Run this recipe" icon={faPlay} onClick={executeFlow} />
+<ListItemButtonIcon title="Run this recipe" icon={faPlay} onClick={handleRunClick} />
 <ListItemButtonIcon inProgress={loading} title="Delete" icon={faTrash} onClick={deleteFlow} />
+
+{#if showModal}
+  <RunFlowModal
+    bind:showModal={showModal}
```
benoitf: issue: here we usually use a onClose or onExit callback

and it's in this page's callback that we do showModal = false

RunFlowModal should not handle this attribute

packages/renderer/src/lib/flows/RunFlowModal.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+const valid = $derived.by(() => {
+  if (flow.parameters) {
+    return flow.parameters.every(param => inputValues[param.name] && inputValues[param.name].trim() !== '');
+  }
+  return true;
+});
+
+$effect(() => {
+  if (showModal) {
```
benoitf: question: in which case do we have this effect with showModal = false as if showModal is false, it's no longer in DOM/destroyed

but showModal should not be there (previous comment)

packages/renderer/src/lib/flows/RunFlowModal.svelte
```diff
@@ -0,0 +1,89 @@
...
+  inputValues = {
+    ...inputValues,
+    [paramName]: (event.target as HTMLInputElement).value,
+  };
+}
+
+function handleRun(): void {
+  if (onRun) {
+    onRun(inputValues);
+  }
```
benoitf: question: in which case onRun would not be defined as it's a mandatory property ?

packages/renderer/src/lib/flows/RunFlowModal.svelte
```diff
@@ -0,0 +1,89 @@
...
+  showModal = false;
+}
+
+function handleCancel(): void {
+  if (onCancel) {
```
benoitf: question: in which case onCancel would not be defined as it's a mandatory property ?

packages/renderer/src/lib/flows/RunFlowModal.svelte
```diff
@@ -0,0 +1,89 @@
...
+    onCancel();
+  }
+  showModal = false;
+}
+</script>
+
+{#if showModal}
```
benoitf: question: I don't see why there is a if there

it's the parent component that ask to display the modal if it's required. So it doesn't need to do anything about the display in this component.
this component assumes it's always being displayed. It's the parent that create/destroy the component (and then manage its visibility)

benoitf:
> For comment https://github.com/kortex-hub/kortex/pull/730, I was thinking maybe I should use onMount?

I suspect the link is not the expected one but I guess you're speaking about the $effect.

if you pickup onMount, if the parameters change (I don't know about how reactivity has been applied on the upper loop) but it means it won't trigger any changes if you do it only in the onMount

PR #732, opened 2025-11-13
chore: changed default location for new chat button

### LINKED ISSUE #582: new chat button location
It seems strange to me to get an action in the middle of selectors

Image

gastoner:
Changes `new chat` button location to be default in sidebar & makes the sidebar non-colapsible

Will remove the toggle if needed 

[Screencast_20251118_071746.webm](https://github.com/user-attachments/assets/5c49683e-4f4a-4ac6-83e8-0fdcadce9d00)

Closes #582

### DIFF
--- a/packages/renderer/src/lib/chat/components/app-sidebar.svelte
+++ b/packages/renderer/src/lib/chat/components/app-sidebar.svelte
@@ -3,7 +3,7 @@ import { router } from 'tinro';
 
 import { currentChatId } from '/@/lib/chat/state/current-chat-id.svelte';
 
-import PenToSquareIcon from './icons/PenToSquareIcon.svelte';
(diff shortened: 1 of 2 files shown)


slemeur:
@gastoner thanks for the PR on this. I agree with you that it looks like a step back.
What I'd like to suggest is that:

1. when the sidebar is opened, we display the plus button (and make it more obvious as you did)
Screenshot 2025-11-14 at 14 10 20

2. when the sidebar is collapsed:
we would make the plus button next to the sidebar button, so we would be moving it from one position on the left.

Later on, this will be reworked as the buttons to choose models, mcp+tools+rag will change.

jeffmaury (approved):
LGTM

PR #736, opened 2025-11-13
chore(container-extension): add container handler that check connections are up-to-date

benoitf:
it also provides two events
- one when there is a new connection change
- one when there is a container change


jeffmaury (approved):
LGTM

PR #738, opened 2025-11-13
chore(ramalama-extension): add container model extractor

benoitf:
extract name of the model and its port from ContainerInfo metadata


jeffmaury (approved):
LGTM

PR #739, opened 2025-11-14
refactor: introduce a method to get flow connection

benoitf:
#519 
for scheduling a flow, need to get the same connection


axel7083 (approved):
Might be useful to link why this is needed in the PR description, and what issue it is required for

If this is pure refactor for refactor, you can just say it is not required but is a nice to have

benoitf:
yes I was adding the description before being asked elsewhere, should be ok now

PR #742, opened 2025-11-14
chore: add methods to interact from IPC with scheduler/flow manager

benoitf:
Scheduler-registry is handling the registration of schedulers, but it was missing methods from IPC to interact with schedulers

like, remotely asking to schedule a job, list the jobs, get the scheduler names. This PR is adding those "IPC" commands

it also adds the IPC commands at the flow level (so that flow manager can ask the scheduler-registry to delegate the execution of jobs)

SchedulerRegistry : interact with the scheduler-extensions (low-level)

FlowManager: allow to schedule Flows delegating to the "internal" scheduler-registry. (it adds the metadata 'flow-scheduler' on all jobs to be able to later retrieve all of them (filter out other jobs that may have been scheduled that are not flows)

### DIFF
--- a/packages/main/src/plugin/scheduler/scheduler-registry.ts
+++ b/packages/main/src/plugin/scheduler/scheduler-registry.ts
@@ -15,13 +15,19 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-import type { ProviderScheduler } from '@kortex-app/api';
+import type {
+  ProviderScheduleExecution,
+  ProviderScheduler,
+  ProviderScheduleResult,
+  ProviderSchedulerOptions,
+} from '@kortex-app/api';
 import { inject, injectable } from 'inversify';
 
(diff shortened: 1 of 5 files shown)


packages/main/src/plugin/scheduler/scheduler-registry.ts
```diff
@@ -31,9 +37,77 @@ export class SchedulerRegistry implements IDisposable {
...
 
-  async init(): Promise<void> {}
+  async init(): Promise<void> {
+    // handle get all schedulers
+    this.ipcHandle('scheduler:get-all', async (): Promise<string[]> => {
+      return this.getSchedulerNames();
+    });
+
+    this.ipcHandle(
+      'scheduler:list-items',
```
axel7083: question: I don't see the corresponding exposeInMainWorld?

benoitf: yes it was an error, it's no longer there

PR #744, opened 2025-11-17
chore(scheduler-extension): adds helper to be able to grab each value of a cron expression

benoitf:
when manipulating a cron/string expression, nice to be able to quickly read each space entry (and the opposite, be able to generate back the string from individual entries)

### DIFF
--- a/extensions/scheduler/src/helper/cron-parser.spec.ts
+++ b/extensions/scheduler/src/helper/cron-parser.spec.ts
@@ -1,0 +1,106 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


extensions/scheduler/src/helper/cron-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+ ***********************************************************************/
+
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import { type CronComponents, CronParser } from './cron-parser';
+
+let cronParser: CronParser;
+beforeEach(async () => {
+  vi.resetAllMocks();
+  cronParser = new CronParser();
```
simonrey1, line 26: remark(non-blocking): you already told me about prefer dependency injection over static method, but in this test `new CronParser()`is not injected and in my opinion it would have been easier to declare `static parse` and use as `CronParser.parse`.

jeffmaury (approved):
LGTM

extensions/scheduler/src/helper/cron-parser.spec.ts
benoitf: for tests, I'm using the DI container when there is an injected attribute. If you think I should use DI to get the object all the time, I'll update all subsequent PR to use.

> in my opinion it would have been easier to declare static parse and use as CronParser.parse.

using DI injection makes the unit tests easier when it's consumed as you control everything that is given to the object. Also there is a singleton so we don't create more than one instance of the class.

simonrey1: No I think it is ok to use the `new` in UT, no problem.

PR #746, opened 2025-11-17
feat: update extension API with RAG providers

### LINKED ISSUE #481: Update extension API to support RAG providers

jeffmaury:
Update extension API to support RAG extension points

Fixes #481

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -682,6 +682,8 @@ declare module '@kortex-app/api' {
     config: MCPServerConfig;
   };
 
+  export type RegisterServerResult = Disposable & { serverId: string };
+
   export type RagProviderConnection = {
     name: string;
     mcpServer: MCPServer;
@@ -1531,7 +1533,7 @@ declare module '@kortex-app/api' {
     export const onDidUpdateRegistry: Event<MCPRegistry>;
     export const onDidUnregisterRegistry: Event<MCPRegistry>;
 
-    export function registerServer(server: MCPServerDetail): Disposable & { serverId: string };
+    export function registerServer(server: MCPServerDetail): RegisterServerResult;
     export function unregisterServer(serverId: string): void;
   }
 
--- a/packages/main/src/plugin/extension/extension-loader.ts
+++ b/packages/main/src/plugin/extension/extension-loader.ts
@@ -20,7 +20,7 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 
 import type * as containerDesktopAPI from '@kortex-app/api';
-import { ChunkProvider } from '@kortex-app/api';
+import { ChunkProvider, RegisterServerResult } from '@kortex-app/api';
 import { components } from '@kortex-hub/mcp-registry-types';
 import AdmZip from 'adm-zip';
 import { app, clipboard as electronClipboard } from 'electron';
@@ -32,6 +32,7 @@ import {
   KubeGeneratorRegistry,
   type KubernetesGeneratorProvider,
 } from '/@/plugin/kubernetes/kube-generator-registry.js';
+import { RegisterServerResultImpl } from '/@/plugin/mcp/register-server-result-impl.js';
 import { MenuRegistry } from '/@/plugin/menu-registry.js';
 import { NavigationManager } from '/@/plugin/navigation/navigation-manager.js';
 import { WebviewRegistry } from '/@/plugin/webview/webview-registry.js';
@@ -1035,15 +1036,12 @@ export class ExtensionLoader implements IAsyncDisposable {
         disposables.push(registration);
         return registration;
       },
-      registerServer: (server: components['schemas']['ServerDetail']): Disposable & { serverId: string } => {
+      registerServer: (server: components['schemas']['ServerDetail']): RegisterServerResult => {
         const serverId = analyzedExtension.id + '.' + server.name;
         mcpRegistryInstance.registerInternalMCPServer({ ...server, serverId });
-        const disposable = Disposable.create(() => {
+        return new RegisterServerResultImpl(() => {
           mcpRegistryInstance.unregisterInternalMCPServer(serverId);
-        });
-        const extendedDisposable = disposable as Disposable & { serverId: string };
-        extendedDisposable.serverId = serverId;
-        return extendedDisposable;
+        }, serverId);
       },
       unregisterServer(serverId: string): void {
         mcpRegistryInstance.unregisterInternalMCPServer(serverId);
--- a/packages/main/src/plugin/mcp/register-server-result-impl.ts
+++ b/packages/main/src/plugin/mcp/register-server-result-impl.ts
@@ -1,0 +1,33 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import type { RegisterServerResult } from '@kortex-app/api';
+
+export class RegisterServerResultImpl implements RegisterServerResult {
+  serverId: string;
+
+  constructor(
+    private disposeFn: () => void,
+    serverId: string,
+  ) {
+    this.serverId = serverId;
+  }
+
+  dispose(): void {
+    this.disposeFn();
+  }
+}


packages/main/src/plugin/chunk-provider-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+      },
+    };
+  }
+
+  getChunkProviders(): ChunkProviderInfo[] {
+    return Array.from(
+      this._chunkProviders.entries().map(entry => ({
+        id: entry[0],
+        name: entry[1].name,
+      })),
```
simonrey1:
```suggestion
      this._chunkProviders.entries().map(([id, { name }]) => ({
        id,
        name,
      })),
```

packages/main/src/plugin/chunk-provider-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+  }
+
+  findProviderById(chunkerId: string): ChunkProvider | undefined {
```
simonrey1: suggestion(
```suggestion
  getProviderById(chunkerId: string): ChunkProvider | undefined {
```

packages/main/src/plugin/chunk-provider-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import type { ChunkProvider, Disposable } from '@kortex-app/api';
+import { inject, injectable } from 'inversify';
+
+import { ChunkProviderInfo } from '/@api/rag/chunk-provider-info.js';
+
+import { ApiSenderType } from './api.js';
+
+@injectable()
+export class ChunkProviderRegistry {
```
simonrey1: question: can this implement Disposable?

packages/main/src/plugin/index.ts
```diff
@@ -2701,6 +2709,29 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle(
```
simonrey1, line 2708: suggestion(non-blocking): add this new `ipcHandle` in a separate file. See `ChatManager` class for instance.

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1663,6 +1754,33 @@ export class ProviderRegistry {
...
+    this.ragProviders.set(id, ragProviderConnection);
+    this.telemetryService.track('registerRagProviderConnection', {
+      name: ragProviderConnection.name,
+      total: this.ragProviders.size,
+    });
+
+    let previousStatus = ragProviderConnection.status();
+
+    // track the status of the provider
+    const timer = setInterval(() => {
```
simonrey1, line 1769: question: what is the cost of this status check? We run it every two seconds so it can use a lot of resources

jeffmaury: That's how all providers check for update and is quite light as it's just a function call to retrieve the status and check for change

packages/main/src/plugin/chunk-provider-registry.ts
jeffmaury: Fixed by e547ad4ef10f0aabab61f2d4854970b96cdb2c55

packages/main/src/plugin/index.ts
jeffmaury: Fixed by e547ad4ef10f0aabab61f2d4854970b96cdb2c55

packages/main/src/plugin/chunk-provider-registry.ts
jeffmaury: Fixed by e547ad4ef10f0aabab61f2d4854970b96cdb2c55

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1173,12 +1232,27 @@ export class ProviderRegistry {
...
     const provider = this.getMatchingProvider(internalProviderId);
 
-    // grab the correct kubernetes connection
+    // grab the correct inference connection
     const connection = provider.inferenceConnections.find(
       connection => connection.name === providerInferenceConnectionInfo.name,
     );
     if (!connection) {
-      throw new Error(`no kubernetes connection matching provider id ${internalProviderId}`);
+      throw new Error(`no inference connection matching provider id ${internalProviderId}`);
```
simonrey1, line 1240: nitpick: thanks for correcting this typo, but this is not in the scope of this PR

packages/main/src/plugin/provider-registry.ts
```diff
@@ -1173,12 +1232,27 @@ export class ProviderRegistry {
     // grab the correct provider
     const provider = this.getMatchingProvider(internalProviderId);
 
-    // grab the correct kubernetes connection
+    // grab the correct inference connection
```
simonrey1, line 1235: nitpick: thanks for correcting this typo, but this is not in the scope of this PR

packages/preload/src/index.ts
```diff
@@ -1519,6 +1513,10 @@ export function initExposure(): void {
     return ipcInvoke('provider-registry:getProviderInfos');
   });
 
+  contextBridge.exposeInMainWorld('getChunkProviders', async (): Promise<ChunkProviderInfo[]> => {
```
simonrey1, line 1516: suggestion(non-blocking): `window.getChunkProviders` will be hard to understand, maybe add `rag` or something to help understand this method.

packages/renderer/src/lib/preferences/PreferencesProviderRendering.svelte
```diff
@@ -57,7 +57,11 @@ let providerDisplayName: string = $derived(
       ? providerInfo?.kubernetesProviderConnectionCreationDisplayName
       : providerInfo?.vmProviderConnectionCreation
         ? providerInfo?.vmProviderConnectionCreationDisplayName
-        : undefined) ?? providerInfo?.name,
+        : providerInfo?.inferenceProviderConnectionCreation
```
simonrey1, line 60: nitpick: adding `inferenceProviderConnectionCreation` here is unrelated to this PR

simonrey1 (approved):
LGTM
question: why is it needed to register a new connection for Rag?

packages/api/src/rag/chunk-provider-info.ts
```diff
@@ -0,0 +1,4 @@
+export interface ChunkProviderInfo {
```
benoitf, line 18: issue: missing headers in the file

packages/extension-api/src/extension-api.d.ts
```diff
@@ -650,16 +651,77 @@ declare module '@kortex-app/api' {
...
+  export type MCPServerDetail = components['schemas']['ServerDetail'];
+
+  export type MCPServer = {
+    server: MCPServerDetail;
+    config: MCPServerConfig;
+  };
+
+  export type RagProviderConnection = {
+    name: string;
+    mcpServer: MCPServer;
```
benoitf, line 689: question: I don't see why we would have the definition of an MCP server here

issue: I would say it should be the id of the MCP server, not its definition.

AFAIK Registering MCP is done separately

packages/extension-api/src/extension-api.d.ts
```diff
@@ -650,16 +651,77 @@ declare module '@kortex-app/api' {
...
+    credentials(): Record<string, string>;
+    index(doc: Uri, chunks: Uri[]): Promise<void>;
+    unindex(doc: Uri): Promise<void>;
```
benoitf: question: not sure about `unindex`, `deindex`

packages/extension-api/src/extension-api.d.ts
```diff
@@ -650,16 +651,77 @@ declare module '@kortex-app/api' {
...
+
+  export type MCPServer = {
+    server: MCPServerDetail;
+    config: MCPServerConfig;
+  };
+
+  export type RagProviderConnection = {
+    name: string;
+    mcpServer: MCPServer;
+    credentials(): Record<string, string>;
```
benoitf: question: how credentials would be used ?

packages/extension-api/src/extension-api.d.ts
```diff
@@ -5247,4 +5334,8 @@ declare module '@kortex-app/api' {
      */
     constructor(url: string);
   }
+
+  namespace rag {
```
benoitf: question/issue: I don't see why ChunkProvider is not part of the existing providers

like, why we have RagConnections in provider namespace but we would have chunks in rag namespace ?

packages/extension-api/src/extension-api.d.ts
jeffmaury: The idea here is that the provider gives information to the core on how to connect to the database so it gives the server definition and the configuration to connect to.

packages/extension-api/src/extension-api.d.ts
jeffmaury: index: add a document to the database, unindex: remove it

packages/extension-api/src/extension-api.d.ts
jeffmaury: Copy/paste reminiscence probably

packages/extension-api/src/extension-api.d.ts
jeffmaury: ChunkProvider will be moved in provider in a follow up PR as I think we will need the factory for them as well.

packages/extension-api/src/extension-api.d.ts
benoitf: my point was that `unindex` seems not the right word here. `deindex` or `removeIndex`, `clearIndex`

packages/extension-api/src/extension-api.d.ts
benoitf: for this, the extension should register an MCP, not embed an MCP definition inside Rag.

any extension could contribute an MCP, and they should not be forced to contribute it as a Rag definition

jeffmaury: Done in c95e5d08d8a4d302821ba2a3508ed8b8507140a9

packages/extension-api/src/extension-api.d.ts
jeffmaury: renamed to deindex

packages/extension-api/src/extension-api.d.ts
benoitf: ok thanks, but then mcpServer entry should be removed from there ?
or should it be a mcpServerId entry ?

jeffmaury: maybe the name mcpServer is wrong as it contains an id and the configuration to connect to

packages/extension-api/src/extension-api.d.ts
benoitf: ok, then it should be exported (coderabbit comment) else it can't be used

jeffmaury:
> LGTM question: why is it needed to register a new connection for Rag?

This is the extension point to register a new vector database into core

packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -1027,6 +1035,19 @@ export class ExtensionLoader implements IAsyncDisposable {
         disposables.push(registration);
         return registration;
       },
+      registerServer: (server: components['schemas']['ServerDetail']): Disposable & { serverId: string } => {
+        const serverId = analyzedExtension.id + '.' + server.name;
+        mcpRegistryInstance.registerInternalMCPServer({ ...server, serverId });
+        const disposable = Disposable.create(() => {
+          mcpRegistryInstance.unregisterInternalMCPServer(serverId);
+        });
+        const extendedDisposable = disposable as Disposable & { serverId: string };
```
simonrey1: suggestion(blocking): to avoid casting and extending manually, create an interface `RegisterServerResult`(name can be refined) to handle this disposable with serverId.

packages/main/src/plugin/mcp/register-server-result-impl.ts
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+import type { RegisterServerResult } from '@kortex-app/api';
+
+export class RegisterServerResultImpl implements RegisterServerResult {
```
simonrey1, line 20: suggestion: Implement the Disposable of Kortex api 

```ts

import type { Disposable,RegisterServerResult } from '@kortex-app/api';

export class RegisterServerResultImpl implements RegisterServerResult, Disposable {
  serverId: string;

  constructor(
    private disposeFn: () => void,
    serverId: string,
  ) {
    this.serverId = serverId;
  }

  dispose(): void {
    this.disposeFn();
  }
}
```

simonrey1: Oops, already done through RegisterServerResult.

Resolving that

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -549,6 +550,22 @@ export class MCPRegistry {
     return data;
   }
 
+  registerInternalMCPServer(server: MCPServerDetail): void {
```
simonrey1, line 553: suggestion(non-blocking): as the class is quite long (> 600 lines), can we add these methods in another class and expose an instance publicly here?

packages/main/src/plugin/index.ts
simonrey1: This comment was marked as resolved because another IPC was moved, but this code is still here because it is the same as the others `provider-registry:create[...]ProviderConnection` above, and they should all be moved in upstream project (podman-desktop).

PR #748, opened 2025-11-17
chore: use static delta in linux pr check when installing flathub binaries

benoitf:
backport from Podman Desktop https://github.com/podman-desktop/podman-desktop/pull/14909


benoitf:
adding auto-merge

serbangeorge-m (approved):

PR #750, opened 2025-11-17
chore(scheduler-extension): parse macOS plist file to extract scheduling

benoitf:
it extracts a cron expression from the scheduling part of a macOS launchagent plist file

### DIFF
--- a/extensions/scheduler/src/helper/macos/cron-plist-parser.ts
+++ b/extensions/scheduler/src/helper/macos/cron-plist-parser.ts
@@ -1,0 +1,49 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


extensions/scheduler/src/helper/macos/cron-plist-parser.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { CronParser } from '/@/helper/cron-parser';
+
+@injectable()
+export class CronPListParser {
+  @inject(CronParser)
+  private readonly cronParser: CronParser;
+
+  extractFromPlist(plistContent: string): string {
+    const minuteRegex = /<key>Minute<\/key>\s*<integer>(\d+)<\/integer>/;
```
simonrey1, line 29: nitpick: this approach with Regex works, but I would have gone with xml parsing, and accessing these properties from the resulting tree.

simonrey1 (approved):
LGTM
I left a nitpick for another approach that would use parsing of XML instead of Regex

jeffmaury (approved):

extensions/scheduler/src/helper/macos/cron-plist-parser.ts
benoitf: ok, I'll handle that in follow-ups

PR #752, opened 2025-11-17
chore(scheduler-extension): extract cron expression from schtasks xml file

benoitf:
on Windows scheduling jobs is done using schtasks. When reading scheduled jobs, need to extract the cron expression from the Windows schtask xml definition

https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/schtasks


extensions/scheduler/src/helper/windows/cron-xml-parser.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+import { CronXmlParser } from './cron-xml-parser';
+
+let cronXmlParser: CronXmlParser;
+
+beforeEach(async () => {
+  vi.resetAllMocks();
+  cronXmlParser = new CronXmlParser();
+});
+
+test('extracts interval-based trigger with 5 minute interval', () => {
```
axel7083: suggestion: those tests look likes they could be mostly test.each, the content could be in a `fixture` files

benoitf: new commit using fixtures

PR #753, opened 2025-11-17
chore(scheduler-extension): generates Windows schtasks xml file to execute jobs

benoitf:
it's generating a schtask xml file that will allow to execute a script for a given cron expression

https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/schtasks

### DIFF
--- a/extensions/scheduler/src/helper/windows/schtasks-xml-generator.spec.ts
+++ b/extensions/scheduler/src/helper/windows/schtasks-xml-generator.spec.ts
@@ -1,0 +1,167 @@
+/**********************************************************************
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import { SchtasksXmlGenerator } from './schtasks-xml-generator';
+
+let schtasksXmlGenerator: SchtasksXmlGenerator;
+
+beforeEach(async () => {
+  vi.resetAllMocks();
+  schtasksXmlGenerator = new SchtasksXmlGenerator();
+});
+
+test('generates TimeTrigger for */5 minute interval', () => {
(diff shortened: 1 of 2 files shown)


extensions/scheduler/src/helper/windows/schtasks-xml-generator.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    storageDir: 'C:\\storage',
+    command: { command: 'echo', args: ['hello'], env: {} },
+    outputFile: 'C:\\logs\\job.log',
+    cronComponents: { minute: '*/5', hour: '*', day: '*', month: '*', weekday: '*' },
+  });
+
+  expect(xml).toContain('<TimeTrigger>');
+  expect(xml).toContain('<Interval>PT5M</Interval>');
+  expect(xml).toContain('<StartBoundary>2025-01-01T00:00:00</StartBoundary>');
+  expect(xml).not.toContain('<StartCalendarInterval>');
```
axel7083, line 44: suggestion: might be a good place to use [vitest snapshots](https://vitest.dev/guide/snapshot)

You do `expect(result).toMatchSnapshot()` and it directly compare to a local file generated by vitest if ran for the first time

extensions/scheduler/src/helper/windows/schtasks-xml-generator.ts
```diff
@@ -0,0 +1,261 @@
...
+    const commandArgs = [command.command, ...command.args].map(arg => `"${arg}"`).join(' ');
+
+    // Complete arguments string for the executor
+    const allArgs = `"${this.escapePowerShellArg(storageDir)}" "${this.escapePowerShellArg(id)}" -Metadata ${escapedMetadataJson} ${commandArgs}`;
+
+    const triggers = this.buildTriggers(cronComponents);
+    const actions = this.buildActions(executorScript, allArgs, command.env);
+    const registrationInfo = this.buildRegistrationInfo(id, metadata);
+
+    const xml = `<?xml version="1.0"?>
```
axel7083, line 53: suggestion(non-blocking): since we use a lot .mustache files, does it makes sense to have it in a separate file? for readability having this inside the source code hardcoded may not be optimal

benoitf: I agree, if possible I would like to do it as a follow-up

extensions/scheduler/src/helper/windows/schtasks-xml-generator.spec.ts
benoitf: I think snapshots is not so interesting here. I mean, it's interesting if I don't need to write the code of the comparison but as auto-completion already do it, I would say it's half-baked.

extensions/scheduler/src/helper/windows/schtasks-xml-generator.ts
axel7083: np, marking conversation as resolved

benoitf: https://github.com/kortex-hub/kortex/issues/757

PR #762, opened 2025-11-18
chore(extension-scheduler): introduce inversify modules for existing code

benoitf:
bring inversify modules. Nothing should change (only bindings)

### DIFF
--- a/extensions/scheduler/src/helper/windows/_module.ts
+++ b/extensions/scheduler/src/helper/windows/_module.ts
@@ -23,7 +23,6 @@ import { SchtasksXmlGenerator } from './schtasks-xml-generator';
 
 const windowsHelpersModule = new ContainerModule(options => {
   options.bind<CronXmlParser>(CronXmlParser).toSelf().inSingletonScope();
-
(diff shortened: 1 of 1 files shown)


extensions/scheduler/src/helper/windows/_module.ts
```diff
@@ -0,0 +1,30 @@
...
+ ***********************************************************************/
+
+import { ContainerModule } from 'inversify';
+
+import { CronXmlParser } from './cron-xml-parser';
+import { SchtasksXmlGenerator } from './schtasks-xml-generator';
+
+const windowsHelpersModule = new ContainerModule(options => {
+  options.bind<CronXmlParser>(CronXmlParser).toSelf().inSingletonScope();
+
```
axel7083: question(non-blocking): why jumping a line here?

benoitf: I don't really know why, probably I like to hit enter key when thinking. I delegate all my newlines to the code formatter.

benoitf: newline removed

Issue #743, opened 2025-11-16
registering an MCP is incorrectly setting the description of the MCP as URI field (problem also when exporting it as a flow)


benoitf:
To reproduce:
1. connect to GitHub MCP
2. use chat with this GitHub MCP
3. export it as a flow

see that in the goose yaml file, the URI is 'GitHub official MCP server" rather than `https://api.githubcopilot.com/mcp/`

quick debugging shows that it gives the description instead of the url field

https://github.com/kortex-hub/kortex/blob/b5eed3e0481cc0296d4aa06351a7b4eba4d696a4/packages/main/src/plugin/mcp/mcp-registry.ts#L197

(if you look at the method definition, there are optional parameters)

maybe need to change from list or parameters (with some optionals) to an object with specific field.

note: if you restart Kortex, it'll correctly register the URL. Only on the first attempt it's failing

PR #770, opened 2025-11-19
feat: introduce native scheduler using LaunchAgents on macOS

benoitf:

to try out:

requirements
- computer using macOS
- use a non-parametrized flow
- rebase https://github.com/kortex-hub/kortex/pull/769 on top of this PR (so you've a basic UI)

go to the schedule tab of flow details
- for flow cron, use `<current minutes + 2> * * * *` (if it's 4.53PM, use `55 * * * *`

wait a couple of minutes, you should through the UI see the execution icon displaying you the result of the scheduled item.

in `~/Library/LaunchAgents` folder you should see `io.github.kortex-hub.kortex.*` jobs

`launchctl list | grep kortex` as well

### DIFF
--- a/extensions/scheduler/src/resources/kortex-scheduler-executor.sh
+++ b/extensions/scheduler/src/resources/kortex-scheduler-executor.sh
@@ -70,9 +70,12 @@ begin_json="{\"id\":\"$scheduler_id\",\"timestamp\":$start_time,\"metadata\":{$m
 echo "$begin_json" >> "$log_file"
 echo "<<<KORTEX_SCHEDULE_TASK_BEGIN_DATA>>>" >> "$log_file"
 
+# Disable exit-on-error temporarily to capture the exit code
+set +e
 "$@" >>"$log_file" 2>&1
-
 exit_code=${PIPESTATUS[0]}
+set -e
+
 end_time=$(date +%s)
 duration=$((end_time - start_time))
 echo "<<<KORTEX_SCHEDULE_TASK_END_DATA>>>" >> "$log_file"


extensions/scheduler/src/handler/launch-agent-scheduler.ts
```diff
@@ -0,0 +1,202 @@
...
+import { process } from '@kortex-app/api';
+import { inject, injectable } from 'inversify';
+
+import type { NativeScheduler } from '/@/api/native-scheduler-api';
+import { CronParser } from '/@/helper/cron-parser.js';
+import { ExecutionParser } from '/@/helper/execution-parser.js';
+import { CronPListParser } from '/@/helper/macos/cron-plist-parser.js';
+import schedulerExecutorScript from '/@/resources/kortex-scheduler-executor.sh?raw';
+
+import { PlistGenerator } from '../helper/macos/plist-generator.js';
```
MarsKubeX: Should be better `import { PlistGenerator } from '/@/helper/macos/plist-generator.js';`

benoitf: fixed

benoitf:
@jeffmaury do you have some files in `~/.local/share/kortex/scheduler/` folder and subfolders with some logs ?

benoitf:
@jeffmaury I think you hit https://github.com/kortex-hub/kortex/issues/743

I will probably need to check error case, probably the shell script abort in case of error and then we do not report the exit code, etc

benoitf:
@jeffmaury I added a new commit that will ensure to continue the execution of the script if goose is failing (then you'll get the error reported)

Issue #596, opened 2025-10-22
split ramalama extension in two parts


benoitf:
extract the RamaLama extensions part about finding container engines

it could be reused by other extensions

PR #773, opened 2025-11-20
chore(ramalama-extension): find models in running containers

benoitf:
note: it's only finding them, no registration at all

related to https://github.com/kortex-hub/kortex/issues/596


jeffmaury (approved):

MarsKubeX (approved):
LGTM

PR #779, opened 2025-11-20
Add stdio transport support for MCP servers

jameslabocki:
## Summary
- Add StdioClientTransport support for local MCP servers
- Fix UI logic bug in MCPRegistryCreateFromRegistryForm.svelte
- Maintain backward compatibility with streamable-http transport
- Enable testing of local MCP servers like podman-mcp-server

## Test plan
- [x] stdio MCP servers can be spawned and connected
- [x] Existing streamable-http servers continue to work
- [x] UI correctly detects remote vs package types
- [x] Environment variables properly inherited for child processes

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/main/src/plugin/index.ts
+++ b/packages/main/src/plugin/index.ts
@@ -2132,6 +2132,10 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle('mcp-registry:clearConfigurations', async (): Promise<void> => {
+      await mcpRegistry.clearConfigurations();
+    });
+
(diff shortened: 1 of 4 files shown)


benoitf (changes requested):
issue: it can't be merged due to hardcoded values (like the name of a mcp and then in PATH some macOS values (if you're on Windows it'll cause issues)

also there is an NPMSpawner that does the job of starting local npm commands

PR #780, opened 2025-11-21
feat: detect RamaLama models and register them

benoitf:
you may follow the README of this extension as a test

fixes https://github.com/kortex-hub/kortex/issues/403

### DIFF
--- a/extensions/ramalama/src/manager/inference-model-manager.ts
+++ b/extensions/ramalama/src/manager/inference-model-manager.ts
@@ -16,12 +16,64 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import type { Disposable } from '@kortex-app/api';
-import { injectable } from 'inversify';
+import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
+import type { Disposable, Provider } from '@kortex-app/api';
+import { inject, injectable } from 'inversify';
 
+import type { ModelInfo } from '/@/api/model-info';
+import { ModelsHandler } from '/@/handler/models-handler';
+import { RamalamaProvider } from '/@/inject/symbol';
+
+// class is responsible to manage the inference models available
 @injectable()
-export class InferenceModelManager implements Disposable {
-  async init(): Promise<void> {}
+export class InferenceModelManager {
+  @inject(ModelsHandler)
+  private modelsHandler: ModelsHandler;
 
-  dispose(): void {}
+  @inject(RamalamaProvider)
+  private ramaLamaProvider: Provider;
+
+  #modelsDisposable: Disposable[] = [];
+
+  async init(): Promise<void> {
+    this.modelsHandler.onModelsChanged(async models => {
+      await this.refreshRegistrationOfModels(models);
+    });
+
+    await this.modelsHandler.init();
+  }
+
+  async refreshRegistrationOfModels(models: readonly ModelInfo[]): Promise<void> {
+    // dispose previous registrations
+    for (const disposable of this.#modelsDisposable) {
+      disposable.dispose();
+    }
+    this.#modelsDisposable = [];
+
+    for (const modelinfo of models) {
+      const sdk = createOpenAICompatible({
+        baseURL: `http://localhost:${modelinfo.port}`,
+        name: `RamaLama/${modelinfo.port}`,
+      });
+      const disposable = this.ramaLamaProvider.registerInferenceProviderConnection({
+        name: `port/${modelinfo.port}`,
+        sdk,
+        status() {
+          return 'started';
+        },
+        models: [{ label: modelinfo.name }],
+        credentials() {
+          return {};
+        },
+      });
+      this.#modelsDisposable.push(disposable);
+    }
+  }
+
+  dispose(): void {
+    for (const disposable of this.#modelsDisposable) {
+      disposable.dispose();
+    }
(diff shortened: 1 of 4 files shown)


extensions/ramalama/src/manager/inference-model-manager.ts
```diff
@@ -16,12 +16,64 @@
...
+export class InferenceModelManager {
+  @inject(ModelsHandler)
+  private modelsHandler: ModelsHandler;
+
+  @inject(RamalamaProvider)
+  private ramaLamaProvider: Provider;
+
+  #modelsDisposable: Disposable[] = [];
+
+  async init(): Promise<void> {
```
axel7083, line 38: question: don't we want to use `@postConstructor` for init?

extensions/ramalama/src/manager/inference-model-manager.ts
```diff
@@ -16,12 +16,64 @@
...
+// class is responsible to manage the inference models available
 @injectable()
-export class InferenceModelManager implements Disposable {
-  async init(): Promise<void> {}
+export class InferenceModelManager {
+  @inject(ModelsHandler)
+  private modelsHandler: ModelsHandler;
+
+  @inject(RamalamaProvider)
+  private ramaLamaProvider: Provider;
```
axel7083, line 34: question: I thought the discussion about injecting in the properties directly was not closed especially because we had to disable typescript feature, so I am bit surprised to see it used here

extensions/ramalama/src/manager/inference-model-manager.ts
```diff
@@ -16,12 +16,64 @@
...
+        credentials() {
+          return {};
+        },
+      });
+      this.#modelsDisposable.push(disposable);
+    }
+  }
 
-  dispose(): void {}
+  dispose(): void {
```
axel7083, line 73: question: shouldn't we annotate with `@preDestroy`?

extensions/ramalama/src/manager/inference-model-manager.ts
benoitf: The only argument of not turning it off inside the core part was because you were afraid that we could miss a potential use case where a variable would be not initialized.

image

But here we have 100% of code coverage in unit tests, so there is no reason to not use it.

extensions/ramalama/src/manager/inference-model-manager.ts
benoitf: we're not forced to, there are pros/cons

extensions/ramalama/src/manager/inference-model-manager.ts
benoitf: I would say, same remark that on the postConstruct.

Usually we were using the caller (the object that creates the resource) that is calling the dispose method.

In that case, it's when we're deactiving the extension that it's calling in chain all the dispose() sequences.

if we move that to @preDestroy, it's following a different pattern. I have no preference of mixing, picking one over the other one.

jeffmaury:
> @jeffmaury could you check that on macOS your models are being launched inside containers (and not through llama-cpp binary directly)
> 
> is `ramalama ps` listing some models ?
> 
> or maybe egg and chicken problem in which order are you testing -start Kortex and then, start the RamaLama model and see it's there
> 
>     * start RamaLama model and then start Kortex and see if it's there ?

I started the ramalama model first then Kortex. Will check again the podman machine

extensions/ramalama/src/manager/inference-model-manager.ts
axel7083: The coverage is not an indicator that the code will be working at runtime, this is why we usually have both unit tests and e2e, however even with both we could be missing some edge cases (espacially because e2e are usually not testing every possible things that can be done) that would have been caught with typescript properly enforced.

The problem here is that your class constructor is not protected nor private, so I can write the following anywhere 

```ts
import { InferenceModelManager } from '/@/manager/inference-model-manager';

export class Foo {
  async bar(): Promise<void> {
    const manager = new InferenceModelManager();
    await manager.init();
  }
}
```

Let's think we ave a new feature exposing the Foo#bar method somewhere (E.g. IPC), and write the following tests

```ts
import { beforeEach, expect, test, vi } from 'vitest';
import { InferenceModelManager } from '/@/manager/inference-model-manager';
import { Foo } from '/@/manager/foo';

vi.mock(import('/@/manager/inference-model-manager'));

beforeEach(() => {
  vi.mocked(InferenceModelManager.prototype.init).mockResolvedValue(undefined);
});

test('demo foo', async () => {
  const foo = new Foo();
  await foo.bar();
  expect(foo).toBeDefined();
});
```

And we yet have 100% coverage, but runtime will crash when calling `Foo#bar`,  

Image

And let's imagine the reviewer trusted the coverage, and approved the change, so now we have a bug that could have been caught by strict property initialisation.

We could think about enforcing with eslint, or any tooling that `@injectable` class cannot be constructed, but yet all non injectable classes can now declare undefined properties and use them without anything saying that this is not initialised.

---

My point being I am against having specific config less strict for inversify, as we do a lot of copy paste in those projects, and one day this tsconfig will be used somewhere or propagated to large packages or new one without inversify, and errors could happen.

jeffmaury (commented):
LGTM. Can you rebase it so that required workflows are executed ?

jeffmaury (approved):
LGTM

PR #785, opened 2025-11-24
feat: store tools in MCPRemoteServerInfo

axel7083:
## Description

Following discussion in last daily kortex, we draw a list of steps to achieve https://github.com/kortex-hub/kortex/pull/716. See full https://github.com/kortex-hub/kortex/pull/716#issuecomment-3528448145

As suggested by @jeffmaury we can store the list of tools in the MCPRemoteServerInfo object

### DIFF
--- a/packages/main/src/plugin/mcp/mcp-manager.ts
+++ b/packages/main/src/plugin/mcp/mcp-manager.ts
@@ -133,7 +133,7 @@ export class MCPManager implements IAsyncDisposable {
       name: connectionName,
       url: url ?? '',
       description: description ?? '',
-      tools: tools,
+      tools,
     };
     this.#mcps.push(mcpRemoteServerInfo);
 


packages/main/src/plugin/mcp/mcp-manager.ts
```diff
@@ -117,12 +117,23 @@ export class MCPManager implements IAsyncDisposable {
...
+      ]),
+    );
+
     const mcpRemoteServerInfo: MCPRemoteServerInfo = {
       id: key,
       infos: { internalProviderId, remoteId: index, serverId },
       name: connectionName,
       url: url ?? '',
       description: description ?? '',
+      tools: tools,
```
benoitf: nit
```suggestion
      tools,
```

packages/renderer/src/lib/mcp/MCPDetails.svelte
```diff
@@ -29,18 +29,12 @@ async function refreshMessages(): Promise<void> {
   }
 }
 
-const mcpServer = $derived($mcpRemoteServerInfos.find(server => server.id === id));
+const mcpServer: MCPRemoteServerInfo | undefined = $derived($mcpRemoteServerInfos.find(server => server.id === id));
```
benoitf, line 32: question: why would we want to add typing there ?

benoitf (approved):

packages/renderer/src/lib/mcp/MCPDetails.svelte
axel7083: My IDE was lost and was not giving me type completion for the $derived bellow :/ I can remove it if you prefer

benoitf: I'm ok, was wondering as on my end I have correct typing in VS Code

image

axel7083: Yeahh, Webstorm is terrible with Svelte

PR #792, opened 2025-11-26
fix(ramalama-extension): track containers when initializing the handler

benoitf:
ensure we scan containers at the first launch, not only when it's coming later

related to https://github.com/kortex-hub/kortex/pull/780#issuecomment-3571732831

### DIFF
--- a/extensions/ramalama/src/handler/models-handler.ts
+++ b/extensions/ramalama/src/handler/models-handler.ts
@@ -72,6 +72,11 @@ export class ModelsHandler {
 
     // get the initial connections
     this.#connections = this.containerExtensionAPI.getEndpoints();
+
+    // refresh the models at startup
+    this.refreshModels().catch((error: unknown) => {
+      console.error('Error refreshing models:', error);
+    });
   }
 
   async refreshModels(): Promise<void> {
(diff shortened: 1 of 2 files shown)


extensions/ramalama/src/handler/models-handler.ts
```diff
@@ -72,6 +72,11 @@ export class ModelsHandler {
 
     // get the initial connections
     this.#connections = this.containerExtensionAPI.getEndpoints();
+
+    // refresh the models at startup
+    this.refreshModels().catch((error: unknown) => {
```
simonrey1, line 77: question: the enclosing method `async init(): Promise<void>` supports the `await` syntax, so should we await the `refreshModels()`?

benoitf: it's done on purpose. It's not hold the startup (do it in the background)

benoitf: when adding this new call, I thought at some point to introduce a new method to be called without being async but it would be a separate refactoring (to avoid all the call.catch.console duplicates)

simonrey1: Thanks for the clarification, I will resolve this conversation

PR #803, opened 2025-11-28
Feat/input fields section in flow creation

### LINKED ISSUE #516: Add ability to define input fields in a flow
* When creating a flow, there will be a new section in the form for “Input Fields”.   
  * An input field is defined by:  
    * A description: which explain what the input field is for  
    * An input “name”: which would be the variable we use in the flow  
    * A default value:an optional default value for the input field  
    * A value type: Text, Number, Boolean, Enum.   
    * A required value: indicating is the input is mandatory for the flow or not  
  * This section should be displaying a table with the description of the input field, the default value and show an icon for input type  
  * There should be a way to manage the input fields (add remove edit)  
  * When clicking on the add (or edit), there will be a modal displayed to provide all the information about the input field

MarsKubeX:
Adds UI components to define input parameters in the Flow Create form. Note: This PR only implements the frontend UI; parameters are not yet passed during flow generation. All parameters currently type "text" (extensible for future types)

**New Components:**

- InputFieldModal.svelte: Dialog for adding/editing input fields with validation
- InputFieldTable.svelte: Table displaying defined parameters with actions
- InputFieldNameColumn.svelte: Custom renderer for parameter name/description
- InputFieldActionsColumn.svelte: Custom renderer for edit/delete buttons
- input-field.ts: TypeScript type definition for input fields

closes #516 

https://github.com/user-attachments/assets/bf3de43c-88f5-4f33-ad62-c3dbbddff798

### DIFF
--- a/packages/renderer/src/lib/flows/components/InputFieldModal.svelte
+++ b/packages/renderer/src/lib/flows/components/InputFieldModal.svelte
@@ -38,7 +38,7 @@ const parseResult = $derived(
     name,
     description,
     format: type,
-    default: defaultValue.trim() || undefined,
+    default: defaultValue || undefined,
     required: computedRequired,
   }),
 );
@@ -49,16 +49,7 @@ function handleSave(): void {
   if (!parseResult.success) {
     return;
   }
-
-  const trimmedDefaultValue = defaultValue.trim();
-
-  onSave({
-    name: name.trim(),
-    description: description.trim(),
-    format: type,
-    default: trimmedDefaultValue || undefined,
-    required: !trimmedDefaultValue, // Required if no default value
-  });
+  onSave(parseResult.data);
 }
 </script>
 
--- a/packages/renderer/src/lib/flows/types/input-field.ts
+++ b/packages/renderer/src/lib/flows/types/input-field.ts
@@ -29,7 +29,7 @@ export const InputFieldSchema = z.object({
     ),
   description: z.string().trim().min(1, 'Description is required'),
   format: z.enum(['string']),
-  default: z.string().optional(),
+  default: z.string().trim().optional(),
   required: z.boolean(),
 });
 


packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -108,28 +155,31 @@ async function generate(): Promise<void> {
...
           {#if error}
-            <ErrorMessage {error}/>
+            <div class="px-6">
+              <ErrorMessage {error}/>
+            </div>
           {/if}
 
           <form
             novalidate
             class="p-2 space-y-7 h-fit"
```
gastoner, line 130: Thinking if we can use `space-x-6` or something like this. This way you don't have to add px-6 to every element below. It might be nice saving

packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -1,13 +1,16 @@
 <script lang="ts">
+import { faCircleInfo } from '@fortawesome/free-solid-svg-icons';
 import { Button, ErrorMessage, Input } from '@podman-desktop/ui-svelte';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { generate as generateWords } from 'random-words';
 import { onMount } from 'svelte';
+import Fa from 'svelte-fa';
```
gastoner: Can we use `Icon` from `@podman-desktop/ui-svelte`?

packages/renderer/src/lib/flows/components/InputFieldActionsColumn.svelte
```diff
@@ -0,0 +1,32 @@
...
+<div class="flex justify-end gap-1">
+  <ListItemButtonIcon
+    title="Edit parameter"
+    icon={faEdit}
+    onClick={(): void => object.onEdit(object.index, object)}
+  />
+  <ListItemButtonIcon
+    title="Delete parameter"
+    icon={faTrash}
+    onClick={(): void => object.onDelete(object.index, object)}
```
gastoner: Can we create functions for this? I think in podman-desktop repo there is some effort/code guidelines to not use inline functions in components https://github.com/podman-desktop/podman-desktop/blob/main/CODE-GUIDELINES.md#svelte

packages/renderer/src/lib/flows/components/InputFieldModal.svelte
```diff
@@ -0,0 +1,147 @@
...
+    <div class="space-y-3">
+      <!-- Field Name -->
+      <div>
+        <label for="field-name" class="block mb-1 text-sm font-medium">
+          Field Name <span class="text-[var(--pd-state-error)]">*</span>
+        </label>
+        <Input
+          id="field-name"
+          bind:value={name}
+          placeholder="repository_url"
```
gastoner: Image

Here with the field name I would change the placeholder to be at least `e.g. repository_url`, now it shows only Field Name -> repository_url like it must be some repository_url

gastoner (commented):
Left some nitpicks

Otherwise looking good on Mac 🚀

packages/renderer/src/lib/flows/FlowCreate.svelte
```diff
@@ -166,8 +216,38 @@ async function generate(): Promise<void> {
               />
             </div>
 
+            <!-- Input Fields Section -->
```
simonrey1: suggestion(non-blocking): introduce a new svelte component to render this section

simonrey1: (and this could also be moved to the new file: Image)

packages/renderer/src/lib/flows/types/input-field.ts
```diff
@@ -15,7 +15,11 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
-export interface ChunkProviderInfo {
-  id: string;
+
+export interface InputField {
```
simonrey1: suggestion(non-blocking): we could use zod to build this interface and so we could validate the form, I will submit a suggestion PR

simonrey1: I have opened PR https://github.com/MarsKubeX/kortex/pull/1 to give my suggestion

MarsKubeX: Great suggestion @simonrey1. I'll add your PR to my branch. Thanks :)

simonrey1: Thanks for taking that into account. I see that some checks are failing, and also on my PR the fields appear invalid when creating a new field, so these are issues I did not handle in my suggestion

MarsKubeX: No worries I'll take a look and solve those checks

packages/renderer/src/lib/flows/FlowCreate.svelte
MarsKubeX: The problem with this is that the Table component has already its own internal horizontal x spacing. Applying container padding would cause the table to be not aligned as expected.

simonrey1:
Hello, is this PR ready for another review? Have a good day

gastoner (approved):
LGTM, thank you for the changes

Maybe just make the text in the dialog slightly bigger? What does other think about this?

packages/renderer/src/lib/flows/components/InputFieldActionsColumn.svelte
```diff
@@ -0,0 +1,40 @@
...
+  onDelete: (index: number, field: InputField) => void;
+}
+
+interface Props {
+  object: ParameterWithCallbacks;
+}
+
+let { object }: Props = $props();
+
+function handleEdit(): void {
```
simonrey1: suggestion(non-blocking): pass responsability to build these edit and delete methods to the parent

I wrote the change in the following PR https://github.com/MarsKubeX/kortex/pull/2 if you want to take a look

packages/renderer/src/lib/flows/components/InputFieldModal.svelte
```diff
@@ -0,0 +1,146 @@
...
+}
+
+let computedRequired = $derived(!defaultValue.trim());
+
+const parseResult = $derived(
+  InputFieldSchema.safeParse({
+    name,
+    description,
+    format: type,
+    default: defaultValue.trim() || undefined,
```
simonrey1: suggestion(non-blocking): add `trim()` to the `zod` schema instead of trimming the default value here.
  
Image

packages/renderer/src/lib/flows/components/InputFieldModal.svelte
```diff
@@ -0,0 +1,146 @@
...
+  const trimmedDefaultValue = defaultValue.trim();
+
+  onSave({
+    name: name.trim(),
+    description: description.trim(),
+    format: type,
+    default: trimmedDefaultValue || undefined,
+    required: !trimmedDefaultValue, // Required if no default value
+  });
+}
```
simonrey1, line 53:
```suggestion
function handleSave(): void {
  if (!parseResult.success) {
    return;
  }

  onSave(parseResult.data);
}
```

packages/renderer/src/lib/flows/components/InputFieldActionsColumn.svelte
MarsKubeX: Hey @simonrey1. Thanks for the suggestion, I'll merge your PR.

packages/renderer/src/lib/flows/components/InputFieldModal.svelte
MarsKubeX: Good catch! changed

packages/renderer/src/lib/flows/components/InputFieldActionsColumn.svelte
simonrey1: Thanks!

simonrey1 (approved):
LGTM
I am a bit surprised that the description of the input field is compulsory though
Also on the default value column, when it's empty sometimes I get "-", sometimes nothing but it is not really important.

PR #804, opened 2025-12-01
refactor(schtasks): moved xml to mustache files

### LINKED ISSUE #757: convert schtasks generator to a mustache template
to provide better readability, use a mustache template

gastoner:
Moved xml to mustache files

Closes #757 

FYI not adding new lines to some of the `.mustache` files so the final file does not have the blank lines in it


extensions/scheduler/src/helper/windows/schtasks-xml-generator.ts
```diff
@@ -153,58 +159,35 @@ ${actions}
...
-        <DaysOfMonth>
-          ${daysXml}
-        </DaysOfMonth>
-        <Months>
-          <January /><February /><March /><April /><May /><June />
-          <July /><August /><September /><October /><November /><December />
-        </Months>
-      </ScheduleByMonth>
-    </CalendarTrigger>`);
+          const daysOfMonth = days.map(d => `<Day>${d}</Day>`).join('\n          ');
```
simonrey1, line 177: nitpick: there is a bit of xml left in this string interpolation, it could be moved to .mustache with the rest I think

simonrey1: But I guess it is complicated so I will resolve that

extensions/scheduler/types/template.d.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+declare module '*.mustache?raw' {
```
benoitf, line 19: issue: AFAIK usually it's only the type, not the way we import it (raw)

```suggestion
declare module '*.mustache' {
```

extensions/scheduler/src/helper/windows/schtasks-xml-generator.ts
```diff
@@ -18,9 +18,12 @@
 
 import type { ProviderSchedulerCommand } from '@kortex-app/api';
 import { injectable } from 'inversify';
+import mustache from 'mustache';
 
 import type { CronComponents } from '/@/helper/cron-parser';
 
+import taskTemplate from './templates/task.mustache?raw';
```
benoitf: suggestion: use absolute imports (type alias)

PR #822, opened 2025-12-05
fix: enable typecheck on ollama extension

jeffmaury:
While working on extensions, I realize that some of them errored on typecheck.
This fixes the Goose extension

`pnpm exec tsc --noEmit --project extensions/gemini`

### DIFF
--- a/extensions/ollama/src/ollama-extension.ts
+++ b/extensions/ollama/src/ollama-extension.ts
@@ -60,7 +60,12 @@ export class OllamaExtension {
         throw new Error(`HTTP error, status: ${res.status}`);
       }
       const data = await res.json();
-      models = Array.isArray(data.models) ? data.models : [];
+      models =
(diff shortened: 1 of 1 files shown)


extensions/ollama/src/ollama-extension.ts
```diff
@@ -60,7 +60,12 @@ export class OllamaExtension {
         throw new Error(`HTTP error, status: ${res.status}`);
       }
       const data = await res.json();
-      models = Array.isArray(data.models) ? data.models : [];
+      models =
+        data !== null && typeof data === 'object' && 'models' in data
```
benoitf, line 64: `null` is not something we're using, we're using `undefined`

jeffmaury: > `null` is not something we're using, we're using `undefined`

Tried it but tsc will report an error about data possibly null

PR #827, opened 2025-12-09
chore: updated mcp-registry-types dependency version

gastoner:
Updates mcp-registry-types dependency version

We are using Remote type https://github.com/kortex-hub/kortex/blob/main/packages/main/src/plugin/mcp/mcp-registry.ts#L464 but it is deprecated in newer version of openapi spec see https://raw.githubusercontent.com/modelcontextprotocol/registry/refs/tags/v1.3.10/docs/reference/api/openapi.yaml

So this PR changes Remote type to StreamableHttpTransport and SseTransport

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "@eslint/eslintrc": "^3.3.1",
     "@eslint/js": "^9.33.0",
     "@kortex-app/api": "workspace:",
-    "@kortex-hub/mcp-registry-types": "^1.2.3",
+    "@kortex-hub/mcp-registry-types": "^1.4.0-next.20251210140818.04eeb13",
     "@rollup/plugin-commonjs": "^28.0.6",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^16.0.1",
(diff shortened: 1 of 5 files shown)


packages/extension-api/src/extension-api.d.ts
```diff
@@ -41,7 +41,7 @@
 
 declare module '@kortex-app/api' {
   import type { ProviderV2 as AISDKInferenceProvider } from '@ai-sdk/provider';
-  import type { components } from '@kortex-hub/mcp-registry-types';
+  import type { components } from '@kortex-hub/mcp-registry-types' with { 'resolution-mode': 'import' };
```
benoitf: question: do we really need this `with {}` ? or it means the library is not exporting in the correct format ?

gastoner: Otherwise I get error with some resolution :( 
but probably need to update the mcp-registry-types ts-config again

gastoner: `Type-only import of an ECMAScript module from a CommonJS module must have a 'resolution-mode' attribute.`

benoitf: so I think error is on the library side yes

gastoner: @benoitf created https://github.com/kortex-hub/mcp-registry-types/pull/29 that should fix this

PR #833, opened 2025-12-12
feat(test): enabled unit tests

### LINKED ISSUE #731: unit tests check as part of PR checks
unit tests are not launched on PR checks
it's because it was failing

It would be good if failing tests are fixed/ignored to ensure new tests are validated/executed and there is no regression on new code

gastoner:
Enables unit tests in repo + fixes the tests :D 

I feel like this repo still has a lot of unnecessary code related to podman desktop, but tried to fix everything with smallest changes possible 

Closes #731

### DIFF
--- a/extensions/goose/src/test-snapshots/gemini.yaml
+++ b/extensions/goose/src/test-snapshots/gemini.yaml
@@ -13,7 +13,7 @@ stringData:
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: goose-recipe-echo
+  name: job-1
   labels:
     app: goose
   annotations:
@@ -21,10 +21,14 @@ metadata:
     kortex-hub/flow-id: demo-flow-id
 spec:
   template:
+    metadata:
+      annotations:
+        kortex-hub/version: 1.0.0
+        kortex-hub/flow-id: demo-flow-id
     spec:
       containers:
         - name: goose
-          image: quay.io/jotnar/goose:latest
+          image: quay.io/kortex/goose:2025-09-03
           command: ['/usr/local/bin/goose']
           args: ['run', '--recipe', '/home/goose/recipe.yaml']
           envFrom:
@@ -55,12 +59,12 @@ data:
     title: "Paris Explorer"
     name: "Paris Explorer"
     description: "Paris Explorer"
-
+    
     # Required for headless mode
     prompt: "Plan a 7 days trip in Paris, mention per day the planning, starting on monday."
-
+    
     instructions: |
       Your task is to:
       1. List the important monuments
       2. Plan each day of the trip
-
+          
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "kortex",
   "productName": "Kortex",
   "repository": "https://github.com/kortex-hub/kortex",
-  "homepage": "hhttps://github.com/kortex-hub/kortex",
+  "homepage": "https://github.com/kortex-hub/kortex",
   "version": "0.0.1-next",
   "license": "apache-2.0",
   "type": "module",
--- a/packages/main/src/plugin/index.spec.ts
+++ b/packages/main/src/plugin/index.spec.ts
@@ -50,6 +50,8 @@ import type { Task } from './tasks/tasks.js';
 import { Disposable } from './types/disposable.js';
 import { HttpServer } from './webview/webview-registry.js';
 
+vi.mock(import('/@/chat/chat-manager.js'));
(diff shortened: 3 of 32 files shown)


extensions/goose/src/test-snapshots/gemini.yaml
```diff
@@ -55,12 +59,12 @@ data:
     title: "Paris Explorer"
     name: "Paris Explorer"
     description: "Paris Explorer"
-
+
```
benoitf, line 62: looks unwanted new spaces ?

extensions/goose/src/test-snapshots/gemini.yaml
```diff
@@ -55,12 +59,12 @@ data:
...
     # Required for headless mode
     prompt: "Plan a 7 days trip in Paris, mention per day the planning, starting on monday."
-
+    
     instructions: |
       Your task is to:
       1. List the important monuments
       2. Plan each day of the trip
-
+
```
benoitf, line 70: formatter is not removing these trailing spaces ?

package.json
```diff
@@ -60,7 +60,7 @@
     "typecheck:extensions": "echo TODO_EXTENSIONS",
     "typecheck": "pnpm run typecheck:main && pnpm run typecheck:preload && pnpm run typecheck:renderer && pnpm run typecheck:preload-webview && pnpm run typecheck:extensions && pnpm run typecheck:extension-api",
     "prepare": "husky install",
-    "postinstall": "electron-builder install-app-deps"
+    "postinstall": "electron-builder install-app-deps && pnpm rebuild better-sqlite3"
```
benoitf: I don't see how it's related to unit tests ?

packages/main/src/plugin/index.spec.ts
```diff
@@ -50,6 +50,10 @@ import type { Task } from './tasks/tasks.js';
 import { Disposable } from './types/disposable.js';
 import { HttpServer } from './webview/webview-registry.js';
 
+vi.mock('/@/chat/db/migrate.js', () => ({
```
benoitf: syntax vi.mock(import(

also do we need to implement or just use default automocking ?

packages/renderer/vite.tests.setup.js
```diff
@@ -22,7 +22,16 @@ import 'vitest-canvas-mock';
 import typescript from 'typescript';
 import { expect } from 'vitest';
 
-global.window.matchMedia = () => {};
+global.window.matchMedia = query => ({
```
benoitf: issue: this file should not be modified

(no difference for this file between upstream and downstream)

https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/vite.tests.setup.js

if there is an enhancement, need to go first to upstream

package.json
gastoner: Without this we get 
```
 Error: The module '/Users/runner/work/kortex/kortex/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
Start at 08:41:56
was compiled against a different Node.js version using
Duration 536.97s (transform 28.67s, setup 198.47s, collect 247.94s, tests 260.55s, environment 187.23s, prepare 50.22s)
NODE_MODULE_VERSION 136. This version of Node.js requires
NODE_MODULE_VERSION 127. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
❯ bindings ../../node_modules/bindings/bindings.js:112:48
❯ new Database ../../node_modules/better-sqlite3/lib/database.js:48:64
❯ ChatManager.init src/chat/chat-manager.ts:69:20
67| await mkdir(directory, { recursive: true });
68| }
69| const sqlite = new Database(join(directory, 'chats.db')); ```
```
Running `pnpm rebuild package-name` worked for me locally will try to remove and solve this differently

extensions/goose/src/test-snapshots/gemini.yaml
gastoner: There was poor formating of white spaces that endup with the test failing :(

packages/renderer/src/Loader.spec.ts
```diff
@@ -29,6 +29,10 @@ import { beforeAll, expect, test, vi } from 'vitest';
 import Loader from './Loader.svelte';
 import { lastPage } from './stores/breadcrumb';
 
+vi.mock('./lib/chat/route/CustomChat.svelte', () => ({
```
benoitf, line 32: vi.mock(import( and let's automock ?

PR #837, opened 2025-12-12
fix: handle new Goose path format 

### LINKED ISSUE #825: Goose does not save decode path correctly
When trying out https://github.com/kortex-hub/kortex/pull/771 I've run into a issue that I cant load any flow/list any flow

In the folder I can see the flows, but in the UI:

Image

When starting the extension I can see this error: 
Image

```
console.log(dirname(decoded), this.getBasePath(), decoded)
```

On Windows it adds `\\?\` prefix to C:\.... path so the path so the check will fail 
This the code that thows the error https://github.com/kortex-hub/kortex/blob/main/extensions/goose/src/goose-recipe.ts#L112-L117

jeffmaury:
Fixes #825

### DIFF
--- a/extensions/goose/src/goose-recipe.spec.ts
+++ b/extensions/goose/src/goose-recipe.spec.ts
@@ -15,6 +15,10 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
+import { readFile } from 'node:fs/promises';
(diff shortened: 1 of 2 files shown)


axel7083 (approved):

MarsKubeX (approved):
LGTM codewise

PR #849, opened 2025-12-16
feat: added validation visual indicator to mcp server list 

### LINKED ISSUE #852: Add visual indicator to mcp server list when validation fails
Following https://github.com/kortex-hub/kortex/pull/609

This Issue should add visual indicator to list of mcp servers to those that did not pass the validation

gastoner:
Adds red visual indicator to mcp server list to those servers that does not pass validation  

image

Closes #852 

Needs rebase after https://github.com/kortex-hub/kortex/pull/609

### DIFF
--- a/packages/renderer/src/lib/images/MCPValidServerIndicatorIcon.svelte
+++ b/packages/renderer/src/lib/images/MCPValidServerIndicatorIcon.svelte
@@ -30,4 +30,4 @@ const showInvalidIndicator = $derived(object?.isValidSchema === false);
       </svg>
     </div>
   {/if}
-  </div>
+</div>


packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -549,18 +553,30 @@ export class MCPRegistry {
     // Validate the ServerList data but continue even if invalid
     this.schemaValidator.validateSchemaData(data, 'ServerList', registryURL);
 
+    // Validate each individual ServerResponse and mark invalid ones
+    const validatedServers = data.servers.map(serverResponse => {
+      const isValid = this.schemaValidator.validateSchemaData(serverResponse, 'ServerResponse', registryURL, true);
```
simonrey1: isssue(blocking): I think this is causing performance issue, the interface looks laggy at start.

For example with the https://registry.modelcontextprotocol.io/,

we call validateSchemaData:
- ServerList: 151 times
- ServerResponse: 4441

gastoner:
@simonrey1 I've changed the logic slightly, now only the ServerList is validated and the list of invalid servers are used, so no ServerResponse validation :heavy_check_mark: but I'm afraid that the lagging persists, since we are fetching the mcp servers + validating them every time we open the mcp install tab :( so we need to probably fetch the mcp servers only once during the startup ? but it would be a separate issue (also happens on the main)

packages/main/src/plugin/mcp/mcp-schema-validator.ts
```diff
@@ -51,6 +58,26 @@ export class MCPSchemaValidator {
       );
     }
 
-    return isValid;
+    // For ServerList, extract the names of invalid servers from validation errors
+    if (schemaName === 'ServerList' && validator.errors) {
```
simonrey1: suggestion: validate each ServerDetail instead of the ServerList to avoid complex regex to find which server is invalid

packages/renderer/src/lib/images/MCPIcon.svelte
simonrey1, line 1: suggestion(non-blocking): keep MCPIcon.svelte purely for rendering the svg icon, and create a new wrapper Svelte component to display the invalid indicator.

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -546,21 +550,35 @@ export class MCPRegistry {
...
+    const validatedServers = data.servers.map(serverResponse => {
+      const validationResult = this.schemaValidator.validateSchemaData(
+        serverResponse,
+        'ServerResponse',
+        registryURL,
+        false,
+      );
+      return {
+        ...serverResponse,
+        hasInvalidSchema: !validationResult.isValid,
```
simonrey1: suggestion(non-blocking): store isValid instead of inversing it

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -159,7 +159,11 @@ export class MCPRegistry {
 
       const { servers } = await this.listMCPServersFromRegistry(registry.serverUrl);
       for (const rawServer of servers) {
-        const server = this.enhanceServerDetail(rawServer.server);
+        const hasInvalidSchema =
+          'hasInvalidSchema' in rawServer && typeof rawServer.hasInvalidSchema === 'boolean'
```
simonrey1: question: how can rawServer has our property `hasInvalidSchema`?

gastoner:
@simonrey1 updated

packages/renderer/src/lib/images/MCPValidServerIndicatorIcon.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+import type { MCPServerDetail } from '/@api/mcp/mcp-server-info';
+
+import MCPIcon from './MCPIcon.svelte';
+
+interface Props {
+  object: MCPServerDetail;
```
simonrey1, line 7: nitpick: object could be 'server' for instance

packages/main/src/plugin/mcp/mcp-schema-validator.ts
```diff
@@ -32,14 +36,14 @@ export class MCPSchemaValidator {
...
-   * @returns true if valid, false if invalid
+   * @returns Validation result with isValid flag and invalidServerNames (for ServerList)
    */
   validateSchemaData(
     jsonData: unknown,
     schemaName: keyof components['schemas'],
     contextName?: string,
     suppressWarnings: boolean = false,
-  ): boolean {
+  ): SchemaValidationResult {
```
simonrey1: remark: I think this interface was introduced when the property `invalidServerNames` existed. Now it could be rollbacked to boolean I think. (Do no forget to update comment above).

simonrey1 (approved):
Image

I was able to display the list of MCP with the icon of warning:
- the invalid icon is a bit hard to see (I'm colorblind)
- left some nitpicks in the code
- it seems laggy but I think was already the case, to fix it we should implement pagination (different issue)

LGTM

vancura:
@gastoner I can't find the source for this icon. What asset do you need? Is the dot just an overlay?

gastoner:
@simonrey1 FYI I was forced to go back to `object` property from `server`  prop due to linter ^^ see comment above

vancura:
The icon update is in https://github.com/gastoner/kortex/pull/3

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -159,7 +160,7 @@ export class MCPRegistry {
 
       const { servers } = await this.listMCPServersFromRegistry(registry.serverUrl);
       for (const rawServer of servers) {
```
simonrey1, line 162: nitpick: rawServer is a bit misleading because it now has validation data included

packages/main/src/plugin/mcp/mcp-registry.ts
```diff
@@ -139,8 +140,8 @@ export class MCPRegistry {
     this.configuration = this.configurationRegistry.getConfiguration(MCP_SECTION_NAME);
   }
 
-  enhanceServerDetail(server: components['schemas']['ServerDetail']): MCPServerDetail {
-    return { ...server, serverId: encodeURI(server.name) };
+  enhanceServerDetail(server: components['schemas']['ServerDetail'], isValidSchema?: boolean): MCPServerDetail {
+    return { ...server, serverId: encodeURI(server.name), isValidSchema };
```
simonrey1: suggestion(non-blocking): move isValidSchema to the server detail in one pass, instead of first adding it to the server response, and then moving in to the server detail

See https://github.com/gastoner/kortex/pull/4 for an implementation of this suggestion

packages/renderer/src/lib/images/MCPValidServerIndicatorIcon.svelte
```diff
@@ -0,0 +1,33 @@
...
+        <path
+          d="M2.5 2.5L9.5 9.5M9.5 2.5L2.5 9.5"
+          stroke="var(--pd-content-card-bg)"
+          stroke-width="1.5"
+          stroke-linecap="round"
+          stroke-linejoin="round" />
+      </svg>
+    </div>
+  {/if}
+  </div>
```
simonrey1: nitpick: missing newline at EOF

PR #859, opened 2025-12-19
refactor: switched to use of Icon component from podman desktop

gastoner:
Changes from Fa to Icon component from podman desktop UI lib

Required for https://github.com/kortex-hub/kortex/pull/849 (causing errors => some svelte stack exceeded when loading mcp list)


simonrey1 (approved):
I don't see a diff so it looks good to me.

Should this be done upstream instead? This file also exists in Podman-Desktop.

Image

bmahabirbu (approved):
Since the changes are done in PD now is this good to go? Looks good to me codewise

simonrey1 (approved):
The last commit says "replaced content by podman desktop code" so LGTM if it was already validated upstream.

benoitf (approved):

PR #945, opened 2026-01-22
fix(goose): authenticate Octokit with GITHUB_TOKEN to avoid rate limits

serbangeorge-m:
Multiple CI runners were exhausting GitHub's unauthenticated rate limit (60 req/hr). Using GITHUB_TOKEN increases this to 5,000 req/hr

### DIFF
--- a/extensions/goose/src/extension.ts
+++ b/extensions/goose/src/extension.ts
@@ -17,7 +17,8 @@
  ***********************************************************************/
 
 import type { ExtensionContext } from '@kortex-app/api';
-import { cli, env, process, provider, version, window } from '@kortex-app/api';
+import { cli, env, process as processAPI, provider, version, window } from '@kortex-app/api';
+import type { OctokitOptions } from '@octokit/core';
 import { Octokit } from '@octokit/rest';
 
 import { GooseCLI } from './goose-cli';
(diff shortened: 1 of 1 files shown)


extensions/goose/src/extension.ts
```diff
@@ -17,20 +17,25 @@
  ***********************************************************************/
 
 import type { ExtensionContext } from '@kortex-app/api';
-import { cli, env, process, provider, version, window } from '@kortex-app/api';
+import { cli, env, process as processAPI, provider, version, window } from '@kortex-app/api';
```
jeffmaury, line 20: **question:** seems unrelated to this PR ?

serbangeorge-m: @jeffmaury I updated the import

extensions/goose/src/extension.ts
```diff
@@ -18,14 +18,19 @@
...
 
 import { GooseCLI } from './goose-cli';
 import { GooseDownloader } from './goose-downloader';
 import { GooseRecipe } from './goose-recipe';
 
 export async function activate(extensionContext: ExtensionContext): Promise<void> {
-  const octokit = new Octokit();
+  const octokitOptions: OctokitOptions = {};
+  if (globalThis.process.env.GITHUB_TOKEN) {
+    octokitOptions.auth = globalThis.process.env.GITHUB_TOKEN;
```
jeffmaury: **praise:** 
```suggestion
  if (process.env.GITHUB_TOKEN) {
    octokitOptions.auth = process.env.GITHUB_TOKEN;
```

serbangeorge-m: @jeffmaury to use process.env, I’ll have to keep the kortex process import aliased to processAPI(related to your previous comment), otherwise it hides Node’s process and env won’t exist. We still need the Kortex process API for GooseCLI

jeffmaury: > @jeffmaury to use process.env, I’ll have to keep the kortex process import aliased to processAPI(related to your previous comment), otherwise it hides Node’s process and env won’t exist. We still need the Kortex process API for GooseCLI

The go for the import rename as the rest of the codebase is not using globalThis.process.env but process.env

serbangeorge-m: updated, I am also wondering if I should include the octokit/core in package.json as dependency

jeffmaury: > updated, I am also wondering if I should include the octokit/core in package.json as dependency

Not sure if this is required but should be a follow up PR

PR #1015, opened 2026-02-19
fix: report openAI connection as closed if models cannot be retrieved

### LINKED ISSUE #768: unable to remove an invalid OpenAI connection
I registered a local OpenAI connection and then stopped Kortex

when restarting Kortex, the link is no longer valid (as I've stopped the local server)

then I see
```
openai-compatible] openai: failed to restore connection for baseURL http://127.0.0.1:60278 TypeError: fetch failed
```

and connection is not restored.

The issue is that I don't see a way in the UI to remove it, and each time I restart Kortex, I'm getting the error

possible solutions:
- restore the connection is a "disabled/invalid" state so it's visible and can be removed
- if restoration fails, delete it so I don't see the restoration message each time but then if I restart the local server, might still be better if it's still here (disabled but not active)
- other solution ?

jeffmaury:
To test:

- creates an OpenAI connection with a dummy URL and token
- Verifies the connection is listed and can be deleted

Fixes #768

### DIFF
--- a/extensions/openai-compatible/src/openAI.spec.ts
+++ b/extensions/openai-compatible/src/openAI.spec.ts
@@ -267,7 +267,7 @@ describe('restoreConnections', () => {
 });
 
 describe('listModels error handling (through factory)', () => {
-  test('non-200 response should throw', async () => {
+  test('non-200 response should return undefined', async () => {
     const openai = new OpenAI(PROVIDER_API_MOCK, SECRET_STORAGE_MOCK);
     await openai.init();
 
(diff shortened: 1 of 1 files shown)


extensions/openai-compatible/src/openAI.spec.ts
```diff
@@ -290,9 +290,9 @@ describe('listModels error handling (through factory)', () => {
 
     fetchMock.mockResolvedValueOnce({ status: 200, json: async () => ({}) });
 
-    await expect(
-      create({ 'openai.factory.apiKey': 'k', 'openai.factory.baseURL': 'http://x/v1' }),
-    ).rejects.toThrowError('malformed response from http://x/v1');
+    await expect(create({ 'openai.factory.apiKey': 'k', 'openai.factory.baseURL': 'http://x/v1' })).resolves.toBe(
```
MarsKubeX, line 293: Since the expect has change from `.rejects.toThrowError(...)` to `resolves.toBe(...)` I think the tests descriptions should be updated too, to reflects better the new behavior

PR #1049, opened 2026-03-02
feat(Skill manager): added skill manager backend

### LINKED ISSUE #1033: Create a skill manager
As users and extensions should be able to create/register skills, we need a backend component to handle and manages the storage of skills.

As an initial implementation, a skills has the following attributes:

- name (from the SKILL.md)
- path (to folder containing SKILL.md)
- description (from SKILL.md)

Reference: https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview

gastoner:
Added skills manager that stores the route to skill folders in config 

Creates skill directory in kortex home dir

Allows to: 
- registerSkill(folderPath) — Register an existing skill by pointing to a folder that contains a SKILL.md. Parses the frontmatter to extract metadata, validates it, and persists the registration.
- createSkill(options) — Create a new skill from scratch by providing name, description, and content. Generates the folder and SKILL.md file in the app's skills directory, then registers it.
- unregisterSkill(name) — Remove a skill from the registry by name.
- listSkills() — List all registered skills (returns name, description, content, path for each).
- getSkillContent(name) — Read and return the full SKILL.md content from disk for a given skill.
- listSkillFolderContent(name) — List all files/directories inside a skill's folder.

Closes #1033 

Added tests for the skill-manager

### DIFF
--- a/extensions/openai-compatible/package.json
+++ b/extensions/openai-compatible/package.json
@@ -34,7 +34,7 @@
     "watch": "vite build --watch"
   },
   "dependencies": {
-    "@ai-sdk/openai-compatible": "^2.0.33",
+    "@ai-sdk/openai-compatible": "^2.0.35",
     "@kortex-app/api": "workspace:*"
   },
   "devDependencies": {
--- a/extensions/openshift-ai/package.json
+++ b/extensions/openshift-ai/package.json
@@ -40,7 +40,7 @@
     "watch": "vite build --watch"
   },
   "dependencies": {
-    "@ai-sdk/openai-compatible": "^2.0.33",
+    "@ai-sdk/openai-compatible": "^2.0.35",
     "@kortex-app/api": "workspace:*",
     "@kubernetes/client-node": "^1.3.0"
   },
--- a/extensions/ramalama/package.json
+++ b/extensions/ramalama/package.json
@@ -21,7 +21,7 @@
     "watch": "vite build --watch"
   },
   "dependencies": {
-    "@ai-sdk/openai-compatible": "^2.0.33",
+    "@ai-sdk/openai-compatible": "^2.0.35",
     "inversify": "^7.7.1"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -563,8 +563,8 @@ importers:
   extensions/openai-compatible:
     dependencies:
       '@ai-sdk/openai-compatible':
-        specifier: ^2.0.33
-        version: 2.0.33(zod@4.3.6)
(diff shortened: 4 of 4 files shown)


packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+    const content = rawContent.slice(endIndex + DELIMITER.length + 1).trim();
+    return { metadata: parsed as SkillMetadata, content };
+  }
+
+  private validateMetadata(metadata: SkillMetadata, filePath: string): void {
+    if (typeof metadata.name !== 'string' || !metadata.name) {
+      throw new Error(`Missing or invalid 'name' in ${filePath}`);
+    }
+
+    // Requirements specifified in the Claude documentation:
```
MarsKubeX: Nitpick: "specifified" should be "specified".

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  private async loadSkillsFromConfig(): Promise<void> {
+    const stored = this.configuration.get<{ path: string }[]>(SKILL_REGISTERED) ?? [];
+    await Promise.allSettled(
+      stored.map(async (entry: { path: string }) => {
+        try {
+          const skillFilePath = join(entry.path, SKILL_FILE_NAME);
+          const metadata = await this.parseSkillFile(skillFilePath);
+          this.skills.push({ ...metadata, path: entry.path });
```
MarsKubeX: Nit: loadSkillsFromConfig mutates the array in place with `.push()`, while `registerSkill` and `createSkill` use the immutable pattern (this.skills = [...this.skills, newItem]). Would it make sense to be consistent here?

` this.skills = [...this.skills, { ...metadata, path: entry.path }];`

gastoner: Should be now consistent

packages/main/src/plugin/index.ts
```diff
@@ -2177,6 +2181,30 @@ export class PluginSystem {
       },
     );
 
+    this.ipcHandle('skill-manager:listSkills', async (): Promise<SkillInfo[]> => {
```
benoitf: all the ipc handle should be done in the skill-manager file and not in the index.ts
https://github.com/kortex-hub/kortex/blob/main/packages/main/src/plugin/flow/flow-manager.ts#L263-L283

it's to avoid the nightmare on rebasing things

gastoner: Will update

PR #1067, opened 2026-03-05
feat: agent workspaces manager

### LINKED ISSUE #1039: Create a mock component to manage agent workspaces
Create a component that will mock the future CLI calls

Agent workspaces are stored in memory and initially a limited set of workspaces are created and returned:

- one running claude workspace
- one stopped claude workspace
- one stopped cursor cli

The component should be able to support agent workspace creation, once the format of the agent workspace configuration file is defined by https://github.com/kortex-hub/kortex-cli

MarsKubeX:
Add a mock backend component to manage agent workspaces in memory, as a foundation for the Agentic Coding Workspaces feature.

The services added to the backend are:

* List — Returns all agent workspaces currently in memory
* Get — Returns a single workspace by its ID; throws if not found
* GetStatus — Returns live status (state + context usage) for a workspace
* Create — Creates a new workspace from the given options
* Start — Transitions a workspace to running status and records startedAt.
* Stop — Transitions a workspace to stopped status.
* Delete — Permanently removes a workspace from memory; throws if not found

All mutating operations emit an agent-workspace:updated event so the renderer can refresh the workspace list.

Added test unit for agent-workspace-manager.ts

Closes #1039

### DIFF
--- a/packages/api/src/agent-workspace-info.ts
+++ b/packages/api/src/agent-workspace-info.ts
@@ -18,31 +18,15 @@
 
 import type { components } from '@kortex-hub/kortex-cli-api';
 
-export type AgentType = 'claude' | 'cursor' | 'goose';
 export type AgentWorkspaceState = 'running' | 'stopped' | 'error';
 export type FileAccessLevel = 'workspace' | 'home' | 'custom' | 'full';
 
 /**
- * CLI workspace type from the `kortex workspace list` command.
+ * Workspace data from the `kortex workspace list` command.
+ * Matches the CLI contract exactly — fields will be added here
+ * as the CLI evolves and publishes them in @kortex-hub/kortex-cli-api.
  */
-export type CliWorkspace = components['schemas']['Workspace'];
-
-/**
- * Static workspace data displayed in each card of the list view.
- *
- * Extends the CLI `Workspace` schema with optional fields
- * that will be populated as the CLI evolves.
- */
-export interface AgentWorkspaceSummary extends CliWorkspace {
-  description?: string;
-  agent?: AgentType;
-  model?: string;
-  resources?: {
-    skills: string[];
-    mcpServers: string[];
-  };
-  createdAt?: string;
-}
+export type AgentWorkspaceSummary = components['schemas']['Workspace'];
 
 /**
  * Live-updating status: state and context/token usage.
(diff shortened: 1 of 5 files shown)


packages/api/src/agent-workspace-info.ts
```diff
@@ -0,0 +1,88 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { components } from '@kortex-hub/kortex-cli-api';
+
+export type AgentType = 'claude' | 'cursor' | 'goose';
```
benoitf: IMHO it can't really be coded in the UI, I would guess the cli is giving the available supported providers

packages/api/src/agent-workspace-info.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export interface AgentWorkspaceSummary extends CliWorkspace {
+  description?: string;
+  agent?: AgentType;
+  model?: string;
+  resources?: {
+    skills: string[];
+    mcpServers: string[];
+  };
+  createdAt?: string;
```
benoitf: shouldn't it go first in the spec rather than diverging from the spec ?

packages/api/src/agent-workspace-info.ts
feloy: Yes, you can have a mock for the moment, mocking `kortex-cli workspace supported-agents` or similar command (https://github.com/kortex-hub/kortex-cli/issues/35)

MarsKubeX: @benoitf Changed to have a `getSupportedAgents` function that will get supported agents from CLI

packages/api/src/agent-workspace-info.ts
MarsKubeX: @benoitf Removed from the interface definition. Now the interface matches the CLI spec.

benoitf: @MarsKubeX but it's still part of


```
export interface AgentWorkspaceInfo extends AgentWorkspaceSummary, AgentWorkspaceStatus {
  fileAccess?: FileAccessLevel;
  customPaths?: string[];
  stats?: {
    messages: number;
    toolCalls: number;
    filesModified: number;
    linesChanged: number;
  };
}
```


I would expect that UI is not defining anything that is not provided by the CLI

PR #1077, opened 2026-03-06
feat(Skills manager): added skills manager backend

### LINKED ISSUE #1033: Create a skill manager

gastoner:
Added skills manager that stores the route to skill folders in config 

Creates skill directory in kortex home dir

Allows to: 
- registerSkill(folderPath) — Register an existing skill by pointing to a folder that contains a SKILL.md. Parses the frontmatter to extract metadata, validates it, and persists the registration.
- createSkill(options) — Create a new skill from scratch by providing name, description, and content. Generates the folder and SKILL.md file in the app's skills directory, then registers it.
- unregisterSkill(name) — Remove a skill from the registry by name.
- listSkills() — List all registered skills (returns name, description, content, path for each).
- getSkillContent(name) — Read and return the full SKILL.md content from disk for a given skill.
- listSkillFolderContent(name) — List all files/directories inside a skill's folder.

Closes #1033 

Added tests for the skill-manager

### DIFF
--- a/packages/preload/src/index.ts
+++ b/packages/preload/src/index.ts
@@ -134,7 +134,7 @@ import type { PullEvent } from '/@api/pull-event';
 import type { ChunkProviderInfo } from '/@api/rag/chunk-provider-info';
 import type { ExtensionBanner, RecommendedRegistry } from '/@api/recommendations/recommendations';
 import type { ReleaseNotesInfo } from '/@api/release-notes-info';
-import type { SkillCreateOptions, SkillInfo, SkillMetadata } from '/@api/skill/skill-info';
+import type { SkillCreateOptions, SkillInfo } from '/@api/skill/skill-info';
 import type { StatusBarEntryDescriptor } from '/@api/status-bar';
 import type { PinOption } from '/@api/status-bar/pin-option';
 import type { TelemetryMessages } from '/@api/telemetry';
@@ -334,7 +334,7 @@ export function initExposure(): void {
     return ipcInvoke('flows:listSchedules');
   });
 
-  contextBridge.exposeInMainWorld('listSkills', async (): Promise<Array<SkillMetadata>> => {
+  contextBridge.exposeInMainWorld('listSkills', async (): Promise<Array<SkillInfo>> => {
     return ipcInvoke('skill-manager:listSkills');
   });
 


packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+export class SkillManager {
+  private skills: SkillInfo[] = [];
+  private configuration: Configuration | undefined;
+  private disposables: IDisposable[] = [];
+
+  constructor(
+    @inject(ApiSenderType) private apiSender: ApiSenderType,
+    @inject(IConfigurationRegistry) private configurationRegistry: IConfigurationRegistry,
+    @inject(Directories) private directories: Directories,
+    @inject(IPCHandle) private readonly ipcHandle: IPCHandle,
```
benoitf: question: should all of them be readonly ? (looks like it's coming from copy/paste)

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+    if (typeof metadata.name !== 'string' || !metadata.name) {
+      throw new Error(`Missing or invalid 'name' in ${filePath}`);
+    }
+
+    // Requirements specified in the Claude documentation:
+    // https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#skill-structure
+    if (metadata.name.length > 64) {
+      throw new Error(`'name' exceeds 64 characters in ${filePath}`);
+    }
+    if (!/^[a-z0-9:-]+$/.test(metadata.name)) {
```
benoitf: question: what about comment from coderabbit ?

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+    const skillsDir = this.directories.getSkillsDirectory();
+    const skillDir = join(skillsDir, metadata.name);
+
+    if (existsSync(skillDir)) {
+      throw new Error(`Skill directory already exists: ${skillDir}`);
+    }
+
+    await mkdir(skillDir, { recursive: true });
+
+    const fileContent = `---\nname: ${metadata.name}\ndescription: ${metadata.description}\n---\n\n${options.content}`;
```
benoitf: issue: I would defer the serialization/writing to js-yaml library

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+    this.ipcHandle('skill-manager:createSkill', async (_listener, options: SkillCreateOptions): Promise<SkillInfo> => {
+      return this.createSkill(options);
+    });
+
+    this.ipcHandle('skill-manager:getSkillContent', async (_listener, name: string): Promise<string> => {
+      return this.getSkillContent(name);
+    });
+
+    this.ipcHandle('skill-manager:listSkillFolderContent', async (_listener, name: string): Promise<string[]> => {
+      return this.listSkillFolderContent(name);
```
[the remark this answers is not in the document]
benoitf: per design it's ok

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    const skillsDir = this.directories.getSkillsDirectory();
+    const targetDir = join(skillsDir, metadata.name);
+
+    if (resolvedPath !== targetDir) {
+      if (existsSync(targetDir)) {
+        throw new Error(`Skill directory already exists: ${targetDir}`);
+      }
+      await cp(resolvedPath, targetDir, { recursive: true });
```
MarsKubeX: On a fresh install where the skills directory doesn't exist yet, registerSkill calls `cp()` without ensuring the parent skillsDir is created first. Unlike createSkill, which uses `mkdir(skillDir, { recursive: true })`, `cp({ recursive: true })` only copies the source tree. Consider a verification before cp call to be sure if the skills directory exists and if not, create it.

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+        try {
+          const metadata = await this.parseSkillFile(skillFilePath);
+          const duplicate = this.skills.find(s => s.name === metadata.name);
+          if (duplicate) {
+            return;
+          }
+
+          const isEnabled = enabledNames.has(metadata.name);
+
+          this.skills = [
```
MarsKubeX: This assignment for `this.skills` can cause a race condition. Since the `.allSettled` makes promises run concurrently, earlier writes can be overwritten by later ones. Consider using a local array to store discovered skills that can be added to `this.skills` after all promises are resolved.

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+   */
+  async unregisterSkill(name: string): Promise<void> {
+    const skill = this.findSkillByName(name);
+    await rm(skill.path, { recursive: true, force: true });
+    this.skills = this.skills.filter(s => s.name !== name);
+    this.saveSkillsToConfig();
+    this.apiSender.send('skill-manager-update');
+  }
+
+  listSkills(): SkillInfo[] {
```
MarsKubeX, line 290: Suggestion: Since SkillInfo includes content (which could be large) and it's only needed in the detail view, consider introducing a lighter interface without it for listSkills(). This way the list response stays lean, and getSkillContent() handles the full content when needed.

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -0,0 +1,387 @@
...
+    }
+  }
+
+  /**
+   * Registers a skill from an external folder by storing a reference to its
+   * path. The original folder is not copied. The skill is enabled immediately
+   * and the reference is persisted to config.
+   */
+  async registerSkill(folderPath: string): Promise<SkillInfo> {
+    const resolvedPath = resolve(homedir(), folderPath);
```
jeffmaury: **question:** why the skill should be located from the homedir ?

packages/preload/src/index.ts
```diff
@@ -333,6 +334,38 @@ export function initExposure(): void {
     return ipcInvoke('flows:listSchedules');
   });
 
+  contextBridge.exposeInMainWorld('listSkills', async (): Promise<Array<SkillMetadata>> => {
```
benoitf:
```suggestion
  contextBridge.exposeInMainWorld('listSkills', async (): Promise<Array<SkillInfo>> => {
```

?

gastoner: Yes you are correct :D

PR #1078, opened 2026-03-06
feat(Skills list): added page to display skills

### LINKED ISSUE #1034: Create a view/page to display skills
As skills are being introduced into Kortex, we need a view to display the list of skills.

See https://github.com/kortex-hub/kortex-mockups/blob/agentic-coding/skills/index.html

gastoner:
Adds page for showing list of skills

Screenshot 2026-03-06 at 10 04 36

Closes #1034 
Needs rebase after https://github.com/kortex-hub/kortex/pull/1049

Added basic tests

You can try this by going to ~/.config/kortex/skills create a folder with skill and create a valid skill.md file in here
It should be visible in the list, then you can try to:
enable/disable it (see how it changes in configuration) 
delete it (no folder in /kortex/skills and configuration record)

### DIFF
--- a/packages/renderer/src/lib/skills/SkillsList.svelte
+++ b/packages/renderer/src/lib/skills/SkillsList.svelte
@@ -21,9 +21,7 @@ $effect(() => {
   skillSearchPattern.set(searchTerm);
 });
 
-const row = new TableRow<SkillSelectable>({
-  selectable: (_): boolean => false,
-});
+const row = new TableRow<SkillSelectable>({});
 
 const nameColumn = new TableColumn<SkillSelectable>('Name', {
   width: '2fr',


packages/renderer/src/lib/skills/columns/SkillNameColumn.svelte
```diff
@@ -0,0 +1,23 @@
...
+</script>
+
+<div class="flex flex-col whitespace-nowrap max-w-full">
+  <div class="flex items-center max-w-full">
+    <div class="max-w-full">
+      <div class="flex flex-nowrap max-w-full">
+        <div
+          class="text-[var(--pd-table-body-text-highlight)] overflow-hidden text-ellipsis"
+          title={object.name}>
+          {object.name}
```
MarsKubeX: nit: I find quite a lot to have 5 nested div to display `object.name`. Maybe they can be collapsed to be less than 5.

MarsKubeX (commented):
Tested on MacOS, and working nice. The only thing I found in the code, is the comment of code rabbit in navigation-page.ts

packages/renderer/src/lib/skills/SkillsList.svelte
```diff
@@ -0,0 +1,82 @@
...
+  {/snippet}
+
+  {#snippet content()}
+    <div class="flex min-w-full h-full">
+      {#if skills.length === 0}
+        {#if searchTerm}
+          <FilteredEmptyScreen icon={NoLogIcon} kind="skills" bind:searchTerm={searchTerm} />
+        {:else}
+          <SkillEmptyScreen onclick={navigateToCreateSkill} />
+        {/if}
```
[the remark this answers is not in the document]
gastoner: True, but in SkillEmptyScreen there is the button also disabled

packages/renderer/src/lib/skills/SkillsList.svelte
```diff
@@ -0,0 +1,82 @@
...
+
+let searchTerm = $state('');
+
+$effect(() => {
+  skillSearchPattern.set(searchTerm);
+});
+
+const row = new TableRow<SkillSelectable>({
+  selectable: (_): boolean => false,
+});
```
benoitf:
```suggestion
const row = new TableRow<SkillSelectable>({});
```

avoid to put selectable function else we're seeing checkboxes that can't be selected

https://github.com/user-attachments/assets/222449a1-a840-46aa-9a2f-5c4ad7d024bb

gastoner: Screenshot 2026-03-12 at 14 37 16


Fixed @benoitf

benoitf: thanks for the change

PR #1083, opened 2026-03-10
feat(workspace-manager): added workspace manager with list option

MarsKubeX:
Add an AgentWorkspaceManager backend with a list() method that returns mock workspace data matching the @kortex-hub/kortex-cli-api Workspace schema.

Part of #1039

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -24,9 +24,7 @@ import type { AgentWorkspaceSummary } from '/@api/agent-workspace-info.js';
 import { AgentWorkspaceManager } from './agent-workspace-manager.js';
 import { mockListWorkspaces } from './agent-workspace-mock-data.js';
 
-vi.mock('./agent-workspace-mock-data.js', () => ({
-  mockListWorkspaces: vi.fn(),
-}));
+vi.mock(import('./agent-workspace-mock-data.js'));
 
 const TEST_SUMMARIES: AgentWorkspaceSummary[] = [
   {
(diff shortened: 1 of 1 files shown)


packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+
+import type { IPCHandle } from '/@/plugin/api.js';
+import type { AgentWorkspaceSummary } from '/@api/agent-workspace-info.js';
+
+import { AgentWorkspaceManager } from './agent-workspace-manager.js';
+import { mockListWorkspaces } from './agent-workspace-mock-data.js';
+
+vi.mock('./agent-workspace-mock-data.js', () => ({
```
benoitf: issue: the pattern to use is
vi.mock(import(...))

also, does the automocking works (only vi.mock(import('./agent-workspace-mock-data.js')) ?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    paths: { source: '/tmp/ws2', configuration: '/tmp/ws2/.kortex.yaml' },
+  },
+];
+
+let manager: AgentWorkspaceManager;
+
+const ipcHandle: IPCHandle = vi.fn();
+
+beforeEach(() => {
+  vi.clearAllMocks();
```
benoitf: resetAllMocks in beforeEach

packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
MarsKubeX: It does. Changed

packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
MarsKubeX: Changed

PR #1088, opened 2026-03-11
feat(Claude extension): added Claude extension

### LINKED ISSUE #1047: Create a claude extension
Extension should register skills stored in $HOME/.claude/skills

gastoner:
Adds in-built claude extension that registeres skills from claude/skills folder

Closes https://github.com/kortex-hub/kortex/issues/1047

You can test this by creating some skill in claude/skills folder 
Then enable extension (the skill should be visible in the list of skills)
Disable extension, all registered skills should now be unregistered (not visible)


benoitf (commented):
hello, I'm not sure to understand why there is a change in extension-api.d.ts to register skills while it was added at the provider level by https://github.com/kortex-hub/kortex/pull/1195


I guess the extension should reuse the existing extension point ?

extensions/claude/src/extension.ts
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import type { ExtensionContext } from '@kortex-app/api';
+
+import { ClaudeExtension } from './claude-extension';
+
+export async function activate(extensionContext: ExtensionContext): Promise<void> {
```
benoitf: IMHO new extensions should reuse the inversify pattern introduced in latest extensions
https://github.com/kortex-hub/kortex/blob/main/extensions/ramalama/src/ramalama-extension.ts

gastoner: Will update

gastoner:
@benoitf updated, just now the icon when creating the skills is not Claude one :( (but we can create a follow-up pr for this :D)

jeffmaury (commented):
I tested it and it showed the skills from the local Claude skills folder but I don't think we should enable delete in this case as it deleted the skill folder in the local .Claude skills folder which seems weird to me. @slemeur WDYT ?

benoitf (approved):
follow-up: I think we might enhance API like adding icons, having a way to prevent the removal, etc.

But I see that as follow-up

jeffmaury (approved):

gastoner:
@benoitf @benoitf created https://github.com/openkaiden/kaiden/pull/1320 (forgot to rebase this PR)

PR #1111, opened 2026-03-17
feat/workspace detail page view

### LINKED ISSUE #1044: Create a view page for displaying agent workspace details
See https://github.com/kortex-hub/kortex-mockups/blob/agentic-coding/sessions/details.html

MarsKubeX:
- Added workspace detail view (AgentWorkspaceDetails.svelte) that displays workspace configuration data fetched via a new getAgentWorkspaceConfiguration IPC channel, with loading spinner and error state handling.
- Installed @kortex-hub/kortex-workspace-configuration and exported AgentWorkspaceConfiguration type in the shared API package, following the same pattern used for CLI types.
- Added mock data (mockGetWorkspaceConfiguration) simulating future YAML config file reads, plus getConfiguration(id) method on AgentWorkspaceManager with IPC registration.
- Made workspace cards clickable to navigate to /agent-workspaces/:id/summary, with proper keyboard accessibility and event propagation handling on the remove button.
- Added route in App.svelte for the detail view.
- Added/updated unit tests for the manager, card, and detail view components.
- Closes #1044 

https://github.com/user-attachments/assets/d21f8b28-f4cb-443d-838d-7972621e899a

### DIFF
--- a/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
+++ b/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 import { ErrorMessage, Spinner, Tab } from '@podman-desktop/ui-svelte';
-import { onMount } from 'svelte';
 import { router } from 'tinro';
 
 import DetailsCell from '/@/lib/details/DetailsCell.svelte';
@@ -9,7 +8,6 @@ import DetailsTitle from '/@/lib/details/DetailsTitle.svelte';
 import DetailsPage from '/@/lib/ui/DetailsPage.svelte';
 import { getTabUrl, isTabSelected } from '/@/lib/ui/Util';
 import Route from '/@/Route.svelte';
-import type { AgentWorkspaceConfiguration } from '/@api/agent-workspace-info';
 
 interface Props {
   workspaceId: string;
@@ -17,33 +15,14 @@ interface Props {
 
 let { workspaceId }: Props = $props();
 
-let loading: boolean = $state(true);
-let configuration: AgentWorkspaceConfiguration | undefined = $state(undefined);
-let error: string | undefined = $state(undefined);
-
-onMount(() => {
-  window
-    .getAgentWorkspaceConfiguration(workspaceId)
-    .then((config: AgentWorkspaceConfiguration) => {
-      configuration = config;
-    })
-    .catch((err: unknown) => {
-      error = String(err);
-      console.error(err);
-    })
-    .finally(() => {
-      loading = false;
-    });
-});
+const configurationPromise = $derived(window.getAgentWorkspaceConfiguration(workspaceId));
 </script>
 
-{#if loading}
+{#await configurationPromise}
   <div class="flex items-center justify-center h-full">
     <Spinner />
   </div>
-{:else if error}
-  <ErrorMessage error={error} />
-{:else if configuration}
+{:then configuration}
   <DetailsPage title={configuration.name ?? ''}>
     {#snippet tabsSnippet()}
       <Tab title="Summary" selected={isTabSelected($router.path, 'summary')} url={getTabUrl($router.path, 'summary')} />
@@ -66,4 +45,6 @@ onMount(() => {
       </Route>
     {/snippet}
   </DetailsPage>
-{/if}
+{:catch error}
+  <ErrorMessage error={String(error)} />
+{/await}
--- a/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.spec.ts
+++ b/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.spec.ts
@@ -20,18 +20,22 @@ import '@testing-library/jest-dom/vitest';
 
 import { render, screen, waitFor } from '@testing-library/svelte';
 import { writable } from 'svelte/store';
+import { router } from 'tinro';
 import { beforeEach, expect, test, vi } from 'vitest';
 
 import type { AgentWorkspaceConfiguration } from '/@api/agent-workspace-info';
 
 import AgentWorkspaceDetails from './AgentWorkspaceDetails.svelte';
 
-vi.mock('tinro', () => ({
-  router: {
-    ...writable({ path: '/agent-workspaces/ws-1/summary' }),
-    goto: vi.fn(),
-  },
-}));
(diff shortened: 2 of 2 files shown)


packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
```diff
@@ -0,0 +1,69 @@
...
+import type { AgentWorkspaceConfiguration } from '/@api/agent-workspace-info';
+
+interface Props {
+  workspaceId: string;
+}
+
+let { workspaceId }: Props = $props();
+
+let loading: boolean = $state(true);
+let configuration: AgentWorkspaceConfiguration | undefined = $state(undefined);
```
gastoner: I think we can use awaited derived here

packages/main/src/plugin/agent-workspace/agent-workspace-mock-data.ts
```diff
@@ -52,6 +56,22 @@ const INITIAL_SUMMARIES: AgentWorkspaceSummary[] = [
...
+  'mock-ws-api-refactor': {
+    name: 'api-refactor',
+  },
+  'mock-ws-test-suite': {
+    name: 'test-suite-fix',
+  },
+  'mock-ws-frontend': {
+    name: 'frontend-redesign',
+  },
+};
```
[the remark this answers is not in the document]
MarsKubeX: The configuration lookup (CONFIGURATIONS) is intentionally kept separate from the workspace store because this is mock data that will be replaced by real CLI calls (reading the YAML file at paths.configuration). In the real implementation, removing a workspace deletes the config file from disk, so the configuration naturally disappears.

From a UI perspective, this path is unreachable: removing a workspace navigates back to the list view, and the detail view can only be opened from a card in the list — so getConfiguration is never called for a deleted workspace. Adding sync logic to throwaway mock code doesn't seem worth the complexity.

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCard.svelte
```diff
@@ -21,9 +44,12 @@ function handleRemove(): void {
...
 <div
-  class="flex flex-col gap-3 p-4 bg-(--pd-content-card-carousel-card-bg) hover:bg-(--pd-content-card-carousel-card-hover-bg) rounded-md"
-  role="region"
-  aria-label="workspace {workspace.name}">
+  class="flex flex-col gap-3 p-4 bg-(--pd-content-card-carousel-card-bg) hover:bg-(--pd-content-card-carousel-card-hover-bg) rounded-md cursor-pointer"
+  role="button"
+  aria-label="workspace {workspace.name}"
+  onclick={handleOpen}
+  onkeydown={handleKeydown}
+  tabindex="0">
```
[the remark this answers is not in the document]
MarsKubeX: The div[role="button"] with an inner <button> is intentional. We initially used a native <button> as the outer container, but Svelte rejects nested <button> elements (<button> cannot be a descendant of <button>). The div[role="button"][tabindex="0"] approach is the standard workaround and is already used elsewhere in the codebase (e.g., CustomPick.svelte).

The interaction is handled correctly: stopPropagation on both click and keydown on the remove button prevents it from triggering the card's navigation. The Svelte linter does not flag any a11y issues with this pattern. Making the container non-interactive would mean the whole card is no longer clickable, which is a UX regression for a common clickable-card pattern.

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
```diff
@@ -0,0 +1,69 @@
...
+      configuration = config;
+    })
+    .catch((err: unknown) => {
+      error = String(err);
+      console.error(err);
+    })
+    .finally(() => {
+      loading = false;
+    });
+});
```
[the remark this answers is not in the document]
MarsKubeX: The workspaceId prop comes from the tinro route parameter :id. When the user navigates to a different workspace, tinro destroys the old component instance and creates a new one — triggering onMount naturally. The prop never changes on a live instance.

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
MarsKubeX: Thanks for the feedback. I've changed to be a awaited derived

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
```diff
@@ -0,0 +1,50 @@
...
+import { getTabUrl, isTabSelected } from '/@/lib/ui/Util';
+import Route from '/@/Route.svelte';
+
+interface Props {
+  workspaceId: string;
+}
+
+let { workspaceId }: Props = $props();
+
+const configurationPromise = $derived(window.getAgentWorkspaceConfiguration(workspaceId));
```
gastoner, line 18: Cant we do something like 
`let configuration= $derived(await window.getAgentWorkspaceConfiguration(workspaceId));`

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
```diff
@@ -0,0 +1,50 @@
...
+              </tr>
+            {/if}
+          </DetailsTable>
+        </div>
+      </Route>
+    {/snippet}
+  </DetailsPage>
+{:catch error}
+  <ErrorMessage error={String(error)} />
+{/await}
```
gastoner, line 50: There will be then removed the await stuff

gastoner: Same as you had before maybe something like 
```
{#if cpnfiuration}
//Show details page
{:else}
// show spinner
{/if}
```

MarsKubeX: Hi @gastoner, thanks for your feedback! I compared my current approach with `$derived(await ...)` and found that it doesn't give us control over errors. If I use something like:

```
let configuration= $derived(await window.getAgentWorkspaceConfiguration(workspaceId));  
{#if cpnfiuration}
//Show details page
{:else}
// show spinner
{/if}
```
there's no way to display an error when getAgentWorkspaceConfiguration rejects. With `{#await ... {:catch}}` the error state is handled.

PR #1106, opened 2026-03-16
fix(chat): improve delete all chats button

fbricon:
- **fix(chat): reset chat view when deleting all chats**
- **fix(chat): improve delete all chats button visibility**

https://github.com/user-attachments/assets/6447a982-2670-4843-bd83-101c5e9ee476

Fixes #477

PR #1113, opened 2026-03-17
test(chat): add e2e tests for delete all chats button visibility

fbricon:
Separate e2e tests for "delete all chats" button visibility (following up on #1106), as requested by @serbangeorge-m.

~~Includes #1106 commits to keep the PR testable locally.~~ rebased against main, that already contains #1106

### DIFF
--- a/tests/playwright/src/model/pages/chat-page.ts
+++ b/tests/playwright/src/model/pages/chat-page.ts
@@ -213,10 +213,36 @@ export class ChatPage extends BasePage {
     return count;
   }
 
+  async getChatModelNames(): Promise<string[]> {
+    await this.modelDropdownSelector.click();
+    await expect(this.modelDropdownContent).toBeVisible();
+    try {
+      const count = await this.modelMenuItems.count();
+      const modelNames: string[] = [];
+      for (let i = 0; i < count; i++) {
+        const text = await this.modelMenuItems.nth(i).textContent();
+        if (text && !text.toLowerCase().includes('embed')) {
+          modelNames.push(text.trim());
+        }
+      }
+      return modelNames;
+    } finally {
+      await this.page.keyboard.press('Escape');
+    }
+  }
+
   async selectModelByIndex(index: number): Promise<void> {
     await this.modelDropdownSelector.click();
     await expect(this.modelDropdownContent).toBeVisible();
     const modelItem = this.modelMenuItems.nth(index);
+    await expect(modelItem).toBeVisible();
+    await modelItem.click();
+  }
+
+  async selectModelByName(modelName: string): Promise<void> {
+    await this.modelDropdownSelector.click();
+    await expect(this.modelDropdownContent).toBeVisible();
+    const modelItem = this.modelDropdownContent.getByRole('menuitem', { name: modelName, exact: true });
     await expect(modelItem).toBeVisible();
     await modelItem.click();
   }
--- a/tests/playwright/src/specs/provider-specs/chat-smoke.spec.ts
+++ b/tests/playwright/src/specs/provider-specs/chat-smoke.spec.ts
@@ -56,6 +56,12 @@ test.describe
 
       await chatPage.ensureChatSidebarVisible();
       let messageCount = await chatPage.getChatHistoryCount();
+      if (messageCount > 0) {
+        await expect(chatPage.deleteAllChatsButton).toBeVisible();
+        await chatPage.deleteAllChatHistoryItems();
+        await chatPage.verifyChatHistoryEmpty();
+        messageCount = 0;
+      }
 
       const chatSessions = [
         { message: 'What is Kubernetes?', expectedIndex: 1 },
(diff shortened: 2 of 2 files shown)


tests/playwright/src/specs/provider-specs/chat-smoke.spec.ts
```diff
@@ -252,4 +293,35 @@ test.describe
...
+            timeout: TIMEOUTS.MODEL_RESPONSE,
+            message: 'Model should respond with the expected response',
+          })
+          .toBeTruthy();
+      }
+
+      // Wait for all chats to be created
+      await expect
+        .poll(async () => await chatPage.getChatHistoryCount(), { timeout: TIMEOUTS.MODEL_RESPONSE })
+        .toBeGreaterThanOrEqual(chatsToCreate);
```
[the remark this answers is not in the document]
serbangeorge-m: I actually agree with this comment. This test might be very flaky on ci. Could you skip it for linux for now?

tests/playwright/src/model/pages/chat-page.ts
```diff
@@ -210,6 +213,24 @@ export class ChatPage extends BasePage {
     return count;
   }
 
+  async getChatModelIndices(): Promise<number[]> {
```
serbangeorge-m: I do not understand why we need this. What issue is trying to solve? we have already a method handling the dropdown used in multiple tests

tests/playwright/src/model/pages/chat-page.ts
```diff
@@ -174,15 +177,15 @@ export class ChatPage extends BasePage {
     await handleDialogIfPresent(this.page);
   }
 
-  async waitForChatHistoryCount(expectedCount: number, timeout = 10_000): Promise<void> {
+  async waitForChatHistoryCount(expectedCount: number, timeout: number = 10_000): Promise<void> {
```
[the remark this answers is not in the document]
fbricon: fixed

tests/playwright/src/model/pages/chat-page.ts
fbricon: because on my Ollama machine, I have a bunch of embedding models, that do not support chat and fail when called. Of course this doesn't happen on the CI, but I need this to be able to run it locally. This wouldn't be an issue if model selection was deterministic. If a given set of models is a prerequisite for running the tests, then we could select those and only those in the tests

tests/playwright/src/specs/provider-specs/chat-smoke.spec.ts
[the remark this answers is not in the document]
fbricon: these tests have been passing on Linux so far, why not keep them until they start showing flakiness?

tests/playwright/src/model/pages/chat-page.ts
serbangeorge-m: the tests are design to detect ollama models only when OLLAMA_ENABLED is set true

fbricon: yes I run them locally with `OLLAMA_ENABLED=true pnpm run test:e2e --grep "CHAT-"`,  and my embedding model is sometimes picked up

serbangeorge-m: This is a local env issue, not a test issue. The tests are designed for CI where only the correct model(granite3.2:2b) is provisioned

fbricon: code like this shows the tests adapt to the environment they're running on:

https://github.com/kortex-hub/kortex/blob/b695a9f369975f97c71905e29de809034d13f6da/tests/playwright/src/specs/provider-specs/chat-smoke.spec.ts#L107-L112 

If we can't run the e2e tests locally, then there's no incentive to write them, as the feedback loop waiting for CI to run is way too long.

benoitf: @serbangeorge-m I think it would be great if tests, especially smoke tests, can be more robust when running locally.

We're separating configuration files for testing but here it's more about the ollama instance. From the test POV I would say we don't care if there are already existing models or not, we just want to be sure it's using our models

I know that for containers, for example, sometimes we don't care on all containers that we might have already on a podman instance, we check that if we pull an image, this image is there and that if we start the container, it's there.


so what would you recommend to make the test more robust on running within an existing ollama instance (which of courses can contains things)

serbangeorge-m: @benoitf The tests are running ok locally. Rather than tweaking the tests to skip embedding models(which is not a problem at all in the test infrastructure) it would be better to fix this at the product level by filtering the model dropdown based on capabilities. Chat interface should only show models that actually support chat, no?

fbricon: Ideally yes, but there's no surefire way to do it, except [querying](https://ollama.readthedocs.io/en/api/#show-model-information) all models one by one, to see if they maybe contain a chat template, so it's gonna slow things down for sure. Even the Ollama client doesn't do it.

Screenshot 2026-03-19 at 17 36 21

Checking for `embed` in the model name is a heuristic good enough for tests, but I wouldn't trust it for user facing selection.

We can certainly look into it but I think this goes beyond this PR

serbangeorge-m: @fbricon I am ok with the embed filtering in this case but we should really avoid future test adaptations for different custom local environments. Also I think it would be better to return the model names instead of indices to avoid a potential race condition between the dropdown interactions, could you add a selectModelByName() method and use that instead?

fbricon: done

PR #1134, opened 2026-03-18
fix(chat): prevent scroll jump when hovering edit button

fbricon:
if there's a very long conversation, larger than the viewport, then when the mouse hovers over the edit button on top of the convo, the page automatically scrolls down to the end of the page.

Fix involves removing `attributes` from MutationObserver config so CSS class changes
(like hover opacity transitions) no longer trigger auto-scroll to bottom.


MarsKubeX (approved):
LGTM tested on mac.

PR #1140, opened 2026-03-20
feat: start stop button in workspace

### LINKED ISSUE #1123: add button to start/stop workspace in the UI
Add a button to start/stop workspace in the cards of workspace list view and also in the workspace detail view,

MarsKubeX:
Added start and stop buttons ford workspaces in workspace list view and in workspace detail view

https://github.com/user-attachments/assets/b9daa007-0388-459a-b605-483ae24e113d

Closes #1123

### DIFF
--- a/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
+++ b/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetails.svelte
@@ -46,7 +46,6 @@ function handleStartStop(): void {
         title={isRunning ? 'Stop Workspace' : 'Start Workspace'}
         onClick={handleStartStop}
         icon={isRunning ? faStop : faPlay}
-        detailed={true}
         inProgress={inProgress} />
     {/snippet}
     {#snippet tabsSnippet()}


packages/renderer/src/stores/agent-workspaces.ts
```diff
@@ -17,17 +17,57 @@
...
 import type { Writable } from 'svelte/store';
-import { writable } from 'svelte/store';
+import { get, writable } from 'svelte/store';
 
 import type { AgentWorkspaceSummary } from '/@api/agent-workspace-info';
 
+export type AgentWorkspaceStatus = 'stopped' | 'running' | 'starting' | 'stopping';
+
 export const agentWorkspaces: Writable<AgentWorkspaceSummary[]> = writable([]);
+export const agentWorkspaceStatuses: Writable<Map<string, AgentWorkspaceStatus>> = writable(new Map());
```
gastoner, line 27: (NIT: not req) Thinking if e.g. usage of SvelteMap here would be better? + renaming the file or just a Map 
WDYT @benoitf

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCard.svelte
```diff
@@ -63,10 +88,20 @@ function handleRemove(): void {
       <span class="truncate">{workspace.paths.configuration}</span>
     </div>
   </div>
-  <div class="flex justify-end">
+  <div class="flex justify-end gap-1">
+    <button
```
jeffmaury, line 92: **question:** why not using Button component ?

MarsKubeX: For the onclick function I need the MouseEvent to stopPropagation since the action icon lives inside the clickable card and the Button component only exposes onclick?: () => void (no MouseEvent).

jeffmaury: **suggestion:** OK I understand but if we can't use the Button component in this context, shouldn't we have an issue opened upstream ?

MarsKubeX: Hey @jeffmaury great idea. I've create [this issue](https://github.com/podman-desktop/podman-desktop/issues/16745) in the upstream to have this.

packages/renderer/src/stores/agent-workspaces.ts
MarsKubeX: Changed the style for the start/stop button to match the one in the card. I've also created issue #1146 to check the SvelteMap.

PR #1157, opened 2026-03-23
fix(chat): implement stop generation button functionality

### LINKED ISSUE #535: The "Stop Generation" button is unresponsive and does not cancel the AI's response stream
During the sending state of a message, the send button correctly transforms into a "Stop Generation" icon. However, clicking this icon has no effect. The AI continues to generate its response until it completes naturally, ignoring the request to stop.

Image

fbricon:
- Wire abort signal through IPC layer to cancel AI response streams
- Make inferenceStreamText return onDataId synchronously for immediate abort listener registration
- Store AbortController per stream with chatId tracking
- Abort active streams when chat is deleted
- Add E2E test [CHAT-17] for stop generation

https://github.com/user-attachments/assets/a21e31f2-d6d7-4002-b5a7-20008680a062

Fixes #535

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/packages/main/src/chat/chat-manager.ts
+++ b/packages/main/src/chat/chat-manager.ts
@@ -55,6 +55,7 @@ import { ParameterExtractor } from './parameter-extraction.js';
 export class ChatManager {
   private chatQueries!: ChatQueries;
   private userId!: string;
+  private activeStreams = new Map<number, { controller: AbortController; chatId: string }>();
 
   constructor(
     @inject(ProviderRegistry)
@@ -102,6 +103,7 @@ export class ChatManager {
     this.userId = await getOrCreateUserId();
 
     this.ipcHandle('inference:streamText', (_, params) => this.streamText(params));
+    this.ipcHandle('inference:stopStream', (_, onDataId: number) => this.stopStream(onDataId));
     this.ipcHandle('inference:generate', (_, params) => this.generate(params));
     this.ipcHandle('inference:generateFlowParams', (_, params) => this.generateFlowParams(params));
     this.ipcHandle('inference:detectFlowFields', (_, params) => this.detectFlowFields(params));
@@ -145,10 +147,16 @@ export class ChatManager {
   }
 
   private async deleteChat(id: string): Promise<undefined> {
+    this.stopStreamsByChat(id);
     await this.chatQueries.deleteChatById({ id });
   }
 
   private async deleteAllChats(): Promise<undefined> {
+    // Abort all active streams
+    for (const [onDataId, stream] of this.activeStreams.entries()) {
+      stream.controller.abort();
+      this.activeStreams.delete(onDataId);
+    }
     await this.chatQueries.deleteAllChatsForUser({ userId: this.userId });
   }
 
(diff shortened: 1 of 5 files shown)


packages/main/src/chat/chat-manager.ts
```diff
@@ -315,20 +329,40 @@ export class ChatManager {
...
+        const { done, value } = await reader.read();
+        if (done) {
+          break;
+        }
+        this.webContents.send('inference:streamText-onChunk', params.onDataId, value);
       }
-      this.webContents.send('inference:streamText-onChunk', params.onDataId, value);
+    } finally {
+      this.activeStreams.delete(params.onDataId);
+      this.webContents.send('inference:streamText-onEnd', params.onDataId);
```
bmahabirbu, line 373: activeStreams.delete() in stopStream and the one in the finally block can both run so its doing a double delete? should be fine but just a question

fbricon: So I asked Claude, why not just rely on the code in the finally block?

---
The Real Issue: Event Loop Timing

  The timing window is about async execution order in the event loop:
```typescript
  private stopStream(onDataId: number): void {
    const stream = this.activeStreams.get(onDataId);
    if (stream) {
      stream.controller.abort();
      // No delete here - relies on finally
    }
  }
```
  What actually happens:
  1. stopStream() called (synchronous)
  2. controller.abort() triggers abort signal (synchronous)
  3. stopStream() returns (synchronous)
  4. Event loop continues - other code can execute
  5. Eventually, the stream's reader.read() Promise rejects/resolves
  6. finally block runs

  Between steps 3-6, the map still has the entry even though we've "stopped" it.

  Realistic Problematic Scenario

  // Renderer calls: stop stream
  await window.inferenceStopStream(123);
  // ^ This returns immediately after stopStream() executes

  // Immediately after (same tick or next tick):
  await window.deleteChat(chatId);
    → stopStreamsByChat(chatId) runs
    → iterates activeStreams
    → finds entry 123 (STILL THERE - finally hasn't run yet!)
    → aborts the already-aborted controller
    → deletes entry

  // Later (when microtask queue processes):
  // The finally block from the original stream runs
    → tries to delete (already deleted)
    → sends onEnd event

  The issue is async ordering. The finally block runs "eventually" when the Promise settles, but synchronous code keeps executing.

  Why Immediate Delete Matters

  With delete in stopStream():
```typescript
  private stopStream(onDataId: number): void {
    const stream = this.activeStreams.get(onDataId);
    if (stream) {
      stream.controller.abort();
      this.activeStreams.delete(onDataId);  // Synchronous cleanup
    }
  }
```
  Now:
  window.inferenceStopStream(123);
  // Map entry deleted IMMEDIATELY (synchronous)

  window.deleteChat(chatId);
    → stopStreamsByChat(chatId)
    → iterates activeStreams
    → entry 123 is GONE
    → nothing to process

  The Real Benefits

  1. Synchronous state consistency: The map is correct immediately after stopStream() returns, not "eventually"
  2. Prevents redundant operations: stopStreamsByChat() won't re-abort already-stopped streams
  3. Predictable state: Caller can trust the map state without waiting for Promises to settle
  4. finally as safety net: Catches cases we didn't explicitly stop (natural completion, errors)

packages/preload/src/index.ts
```diff
@@ -1396,18 +1396,29 @@ export function initExposure(): void {
...
+      const id = onDataCallbacksStreamTextId;
+      onDataCallbacksStreamText.set(id, { onChunk, onError, onEnd });
+      ipcInvoke('inference:streamText', { ...params, onDataId: id }).catch((err: unknown) => {
+        const callback = onDataCallbacksStreamText.get(id);
+        if (callback) {
+          callback.onError(String(err));
+        }
+      });
+      return id;
     },
```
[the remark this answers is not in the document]
fbricon: fixed

packages/main/src/chat/chat-manager.ts
bmahabirbu: That looks fine to me!

PR #1171, opened 2026-03-24
chore(agent-workspaces): make store reactive via event-driven updates

### LINKED ISSUE #1170: Make agent workspace store reactive via event-driven updates
The agent workspace store (packages/renderer/src/stores/agent-workspaces.svelte.ts) currently only fetches workspace data once on system-ready. After mutations (remove, start, stop), callers must manually invoke fetchAgentWorkspaces() to keep the UI in sync.

MarsKubeX:
Emit 'agent-workspace-update' from AgentWorkspaceManager after mutations and subscribe in the renderer store, removing manual fetchAgentWorkspaces calls from components.

Fixes #1170

### DIFF
--- a/packages/renderer/src/stores/agent-workspaces.svelte.spec.ts
+++ b/packages/renderer/src/stores/agent-workspaces.svelte.spec.ts
@@ -16,62 +16,13 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { get } from 'svelte/store';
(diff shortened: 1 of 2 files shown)


MarsKubeX:
> LGTM but seems to me there was an abstraction for classes that was handling events and make more concise code

You're right @jeffmaury. I've made a small change to use the EventStore, the abstraction for handling store events.

PR #1173, opened 2026-03-25
fix: show reasoning before response and improve streaming indicators

### LINKED ISSUE #1168: Confusing “Reasoning” UI
**Description:**
The current “Reasoning” UI is confusing because it suggests the model is still reasoning while the response is already being streamed. From a user’s perspective, reasoning (i.e., thinking) should happen *before* the answer is generated, not concurrently with it.

Right now, the interface shows a “Reasoning…” indicator while the answer is being produced:

Image

and only afterward displays a “Reasoned for a few seconds” summary:
Image

This creates a mismatch between the expected mental model (reasoning then answer) and what the UI communicates (reasoning happening during answer generation).

**Expected behavior:**
Reasoning should be clearly presented as a prior step:

* Show “Reasoning…” before the response starts
* Display the reasoning summary before the start of the answer
* Use a different label while answering (e.g. “Generating response…”)

**Actual behavior:**

* The answer starts streaming while “Reasoning…” is still shown
* The reasoning summary appears only after the response
* This makes it unclear whether reasoning is happening before, during, or after the response

fbricon:
fix: show reasoning before response and improve streaming indicators

Previously, the "Reasoning..." indicator displayed during the entire
message streaming process, even after reasoning was complete. This was
confusing as users expect reasoning to happen before the answer.

Changes:
- "Reasoning..." shows only during reasoning phase (before text starts)
- Once text starts streaming, shows "Reasoned for X" with actual duration
  using humanize-duration library (e.g., "2 seconds", "1 minute, 30 seconds")
- Reasoning component always renders before text content
- Single global spinner shows at end of message during both reasoning
  and text generation (simplified from having separate spinners)
- Only the last assistant message shows a spinner during generation
- Chevron to expand reasoning rotates: points right when collapsed,
  down when expanded, with smooth transition animation
- Users can expand reasoning during streaming to view content in real-time

https://github.com/user-attachments/assets/881c6f8f-0b42-434b-8cec-a00774e0071a

Fixes #1168

### DIFF
--- a/packages/renderer/src/lib/chat/components/messages/preview-message-test.svelte
+++ b/packages/renderer/src/lib/chat/components/messages/preview-message-test.svelte
@@ -1,0 +1,23 @@
+<script lang="ts">
+import type { UIMessage } from '@ai-sdk/svelte';
+
+import { TooltipProvider } from '/@/lib/chat/components/ui/tooltip';
+
+import PreviewMessage from './preview-message.svelte';
+
+let {
+  message,
+  messages,
+  readonly,
+  loading,
+}: {
+  message: UIMessage;
+  messages: UIMessage[];
+  readonly: boolean;
+  loading: boolean;
+} = $props();
+</script>
+
+<TooltipProvider>
+  <PreviewMessage {message} {messages} {readonly} {loading} />
+</TooltipProvider>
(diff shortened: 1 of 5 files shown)


packages/renderer/src/lib/chat/components/messages/preview-message-test.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
+import type { UIMessage } from '@ai-sdk/svelte';
+
+import { TooltipProvider } from '../ui/tooltip';
```
benoitf: prefer to use /@/ alias import
https://github.com/kortex-hub/kortex/blob/main/.vscode/settings.json

fbricon: fixed

PR #1181, opened 2026-03-25
fix: enable safeStorage on linux and enable tests

serbangeorge-m:
Fixes https://github.com/kortex-hub/kortex/issues/651

### DIFF
--- a/packages/main/src/index.ts
+++ b/packages/main/src/index.ts
@@ -33,10 +33,9 @@ import { AnimatedTray } from './tray-animate-icon.js';
 import { TrayMenu } from './tray-menu.js';
 import { isMac, isWindows, stoppedExtensions } from './util.js';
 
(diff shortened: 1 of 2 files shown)


serbangeorge-m:
> I want to avoid that any testing code goes into production bits
> 
> > It does nothing unless both CI and ELECTRON_PLAINTEXT_SAFE_STORAGE are set
> > there is no reason to have that code be part of the binaries delivered to end users
> 
> This is why I was thinking on including a file, that does nothing but when it's built for tests, we replace that empty file by the 'do that in CI mode' stuff
> 
> so, when building bits for testing purpose, code is there and when building bits for releases, code is never there so you can specify CI or ELECTRON_PLAINTEXT_SAFE_STORAGE it will never do anything

To keep it out of production, I would need to have a file for release builds and one for test builds? swap them at build time? It adds some build complexity I think but if it is a better approach I can try this

PR #1191, opened 2026-03-27
fix(chat): invert syntax highlighting in user message code blocks

### LINKED ISSUE #1190: Syntax highlighting in user messages makes snippets hard to read
In the chat view, ask a model this question:

> 
> can you explain this code:
> \`\`\`java
> public class Fibonacci {
> 
>   public static long fibIterative(int n) {
>       if (n < 0) throw new IllegalArgumentException("n must be >= 0");
>       if (n == 0) return 0;
>       long a = 0, b = 1;               // F(0), F(1)
>       for (int i = 2; i <= n; i++) {
>           long tmp = a + b;
>           a = b;
>           b = tmp;
>       }
>       return b;                         // F(n)
>   }
> \`\`\`

Syntax highlighting is applied to the user message, but is actually hard to read in both Light and Dark modes

Image
Image

fbricon:
- **fix(chat): invert syntax highlighting in user message code blocks**
User message bubbles use bg-primary which inverts the app theme, so
code snippet highlighting colors need to be swapped accordingly to
remain readable in both light and dark modes.

Hex values are defined once as --hljs-light-* / --hljs-dark-* palette
variables and referenced via var() in all theme blocks. These colors
follow the GitHub syntax highlighting scheme and don't map to the
project's tailwind color palette, so they are kept self-contained in
syntax-highlighting.css rather than registered in the ColorRegistry.

- **fix(chat): wrap long lines in code blocks instead of overflowing**

Screenshot 2026-03-27 at 11 48 18
Screenshot 2026-03-27 at 11 48 45

Fixes #1190

### DIFF
--- a/packages/renderer/src/lib/markdown/syntax-highlighting.css
+++ b/packages/renderer/src/lib/markdown/syntax-highlighting.css
@@ -19,46 +19,106 @@
 /* Syntax highlighting styles using GitHub's color scheme */
 /* Uses CSS custom properties for easy light/dark mode switching */
 
-/* Light mode syntax highlighting */
+/* Define light and dark color palettes once (GitHub color scheme) */
 :root {
-  --hljs-comment: #6a737d;
-  --hljs-keyword: #d73a49;
-  --hljs-string: #032f62;
-  --hljs-title: #6f42c1;
-  --hljs-function: #6f42c1;
-  --hljs-number: #005cc5;
-  --hljs-literal: #005cc5;
-  --hljs-built-in: #005cc5;
-  --hljs-type: #d73a49;
-  --hljs-variable: #e36209;
-  --hljs-attribute: #e36209;
-  --hljs-meta: #6a737d;
-  --hljs-tag: #22863a;
-  --hljs-name: #22863a;
-  --hljs-operator: #d73a49;
-  --hljs-bg: transparent;
-  --hljs-fg: #24292e;
+  --hljs-light-comment: #6a737d;
+  --hljs-light-keyword: #d73a49;
+  --hljs-light-string: #032f62;
+  --hljs-light-title: #6f42c1;
+  --hljs-light-number: #005cc5;
+  --hljs-light-variable: #e36209;
+  --hljs-light-tag: #22863a;
+  --hljs-light-fg: #24292e;
+
+  --hljs-dark-comment: #8b949e;
+  --hljs-dark-keyword: #ff7b72;
+  --hljs-dark-string: #a5d6ff;
+  --hljs-dark-title: #d2a8ff;
+  --hljs-dark-number: #79c0ff;
+  --hljs-dark-variable: #ffa657;
+  --hljs-dark-tag: #7ee787;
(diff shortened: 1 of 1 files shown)


packages/renderer/src/lib/markdown/syntax-highlighting.css
```diff
@@ -61,6 +61,47 @@
...
+  --hljs-literal: #005cc5;
+  --hljs-built-in: #005cc5;
+  --hljs-type: #d73a49;
+  --hljs-variable: #e36209;
+  --hljs-attribute: #e36209;
+  --hljs-meta: #6a737d;
+  --hljs-tag: #22863a;
+  --hljs-name: #22863a;
+  --hljs-operator: #d73a49;
+  --hljs-fg: #24292e;
```
benoitf: do you think it can use css variables defined 

usually we don't hardcode any color in the code, colors are coming from color-registry (defining variables) that rely on tailwind-color-palette stuff

fbricon: I've tweaked the css variables a bit, but moving to the color registry doesn't seem to be a good idea according to claude:


> me: could we add the hljs values in tailwind-color-palette.json and then benefit from the color
> registry?
> 
> claude Opus: Not cleanly, because of the user message inversion requirement.
> 
>   The ColorRegistry provides one value per color based on the active theme — e.g., --pd-hljs-comment
>    would be #6a737d in light mode or #8b949e in dark mode. But for user messages, we need both light
>    and dark values available simultaneously so we can swap them. The registry doesn't expose the
>   "other theme's" values.
> 
>   To work around it, we'd have to register each color twice:
>   this.registerColor('hljs-light-comment', { dark: '#6a737d', light: '#6a737d' });
>   this.registerColor('hljs-dark-comment', { dark: '#8b949e', light: '#8b949e' });
> 
>   That misuses the registry — the light/dark fields would have identical values since we're
>   registering palette entries, not themed colors. We'd still need the CSS inversion logic in
>   syntax-highlighting.css to swap them for [data-role='user'].
> 
>   So the current approach (self-contained CSS variables) is the better fit here.

packages/renderer/src/lib/chat/components/messages/preview-message.svelte
```diff
@@ -118,7 +118,7 @@ const isGeneratingResponse = $derived(loading && message.role === 'assistant' &&
             </Tooltip>
           {/if}
           <div
-            class={cn('flex flex-col gap-4', {
+            class={cn('flex flex-col gap-4 overflow-hidden', {
```
gastoner, line 121: Question: Is this really required? Not sure how is the overflow related to syntax highlighting

fbricon: yeah that's why it's in a different commit, it's just something I caught while testing it. Since it's a tiny fix I added it to the PR

gastoner (approved):
LGTM on Mac

bmahabirbu (approved):

PR #1195, opened 2026-03-27
feat: allow to register skills by extensions

### LINKED ISSUE #1031: Update extension to allow extensions to register skills
Extensions should be able to register skills that may be provided by systems the extension is handling connection to

benoitf:
fixes #1031


benoitf:
> Seems to me the implementation of the extension API is missing ?

AFAIK it's there. It's adding a new entry inside a provider so it's modifying ProviderImpl and reusing existing methods where we could add a folder and gets a Disposable

PR #1207, opened 2026-03-31
fix(chat): fix message buttons bar LnF

fbricon:
- **fix: vertically center input button bar with equal padding**
- **fix: use color registry variables for chat input styling** : fixes attach button background on hover
 
Screenshot 2026-03-31 at 12 06 00

### DIFF
--- a/packages/renderer/src/lib/chat/components/multimodal-input.svelte
+++ b/packages/renderer/src/lib/chat/components/multimodal-input.svelte
@@ -235,7 +235,7 @@ $effect((): (() => void) | void => {
   {/if}
 
   <div class={cn(
-    'bg-muted flex flex-col rounded-2xl border dark:border-zinc-700',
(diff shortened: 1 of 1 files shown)


fbricon:
the screenshot was taken from the wrong branch :-/ I still need to fix styling

packages/renderer/src/lib/chat/components/multimodal-input.svelte
```diff
@@ -235,7 +235,7 @@ $effect((): (() => void) | void => {
   {/if}
 
   <div class={cn(
-    'bg-muted flex flex-col rounded-2xl border dark:border-zinc-700',
+    'bg-muted flex flex-col rounded-2xl border border-[var(--pd-input-field-stroke)]',
```
benoitf, line 238: thanks for using a variable now

PR #1242, opened 2026-04-04
feat(skill-manager): prevent removing contributed skills

gastoner:
This PR hides the delete button for skills registered by an extension e.g. Claude

### DIFF
--- a/packages/api/src/skill/skill-info.ts
+++ b/packages/api/src/skill/skill-info.ts
@@ -40,4 +40,5 @@ export interface SkillFileContent extends SkillMetadata {
 export const SKILL_SECTION = 'skills';
 export const SKILL_ENABLED = 'enabled';
 export const SKILL_REGISTERED = 'registered';
+export const SKILL_REMOVED = 'removed';
 export const SKILL_FILE_NAME = 'SKILL.md';
--- a/packages/api/src/skill/skill-info.ts
+++ b/packages/api/src/skill/skill-info.ts
@@ -24,6 +24,7 @@ export interface SkillMetadata {
 export interface SkillInfo extends SkillMetadata {
   path: string;
   enabled: boolean;
+  managed: boolean;
 }
 
 export interface SkillFolderInfo {
@@ -40,5 +41,4 @@ export interface SkillFileContent extends SkillMetadata {
 export const SKILL_SECTION = 'skills';
 export const SKILL_ENABLED = 'enabled';
 export const SKILL_REGISTERED = 'registered';
-export const SKILL_REMOVED = 'removed';
 export const SKILL_FILE_NAME = 'SKILL.md';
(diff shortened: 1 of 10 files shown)


MarsKubeX (approved):
LGTM

gastoner:
> Honestly I don't get why we need a config key for that seems erroneous to me. Can you describe the use case as for me it could me managed through a skill type ?

@jeffmaury 
We need some config key to store the information about which external skill have we deleted (e.g. from ~/.claude/skills). In another start of the app we dont want to loaded previously 'removed' skill (for kaiden it is fine because the skill is really removed, but for external skills we want to just "hide" it). 
By skill type do you mean like external/in-built?

jeffmaury (commented):
Sorry to insist but why the skill is not simply removed from the existing lists

On another topic, not sure using configuration is the right choice

gastoner:
> Sorry to insist but why the skill is not simply removed from the existing lists
> 
> On another topic, not sure using configuration is the right choice

It should be removed, but to not load all the skills again (even the removed ones), after next startup I've added those removed skills to the config so it is not added again 

AFAIK the configuration is only place where this information will persist the restart or am I worong? 

I used a JSON file for RAG environment
What I don't get is that if you store the removed skills in a config key, it will grow indefinitely and why the skills are not removed from the list ? I'm afraid I don't understand how the skills are stored in the config key ?

jeffmaury:
I don't know what issue this PR is trying to solve but according to https://github.com/openkaiden/mockups/blob/main/skills/index.html

Skills can be enabled/disabled and removed.
A disabled skill means that it will not be by default created in the agent workspace configuration when creating an agent
A contributed skill (by an extension) cannot be deleted.

So I think the only information that needs to be stored in the list of enabled skills.

gastoner:
> I don't know what issue this PR is trying to solve but according to https://github.com/openkaiden/mockups/blob/main/skills/index.html
> 
> Skills can be enabled/disabled and removed. A disabled skill means that it will not be by default created in the agent workspace configuration when creating an agent A contributed skill (by an extension) cannot be deleted.
> 
> So I think the only information that needs to be stored in the list of enabled skills.

Without this PR you can delete the contributed skill. So if I understand correctly if you have a contributed skill, you wont see probably the delete button at all. My idea was to keep the delete button for all skills so we can remove the skill from list e.g. when having multiple of them (this is what this PR does). 

Can you please confirm that e.g. we want to hide the delete button for contributed skills ? So they cant be deleted? 
cc @jeffmaury

jeffmaury:
Yes see the mockup

gastoner:
@MarsKubeX @jeffmaury I've updated the PR

jeffmaury (approved):

MarsKubeX (approved):
LGTM. Just update the description of the PR to remove the ```SKILL_REMOVED``` reference.

PR #1270, opened 2026-04-08
chore: rename from kortex to kaiden

benoitf:
step 1 of https://github.com/openkaiden/kaiden/issues/1259

need to rename workspace packages

### DIFF
--- a/extensions/goose/src/kube-template.ts
+++ b/extensions/goose/src/kube-template.ts
@@ -23,7 +23,7 @@ export interface KubeTemplateOptions {
   job: {
     name: string;
   };
-  kortex: {
+  kaiden: {
     version: string;
   };
   recipe: {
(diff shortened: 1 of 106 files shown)


extensions/goose/src/kube-template.ts
```diff
@@ -44,8 +44,8 @@ export interface KubeTemplateOptions {
 // https://github.com/packit/ai-workflows/tree/main/goose-container
 export const GOOSE_IMAGE = 'quay.io/kortex/goose:2025-09-03'; //'quay.io/jotnar/goose:latest';
```
serbangeorge-m, line 45: this should not be renamed?

benoitf: hello @serbangeorge-m so before the rename I need to push the images to another org on quay

PR #1308, opened 2026-04-10
chore(deps): switch the mcp types dependency from kortex to kaiden

benoitf:
related to https://github.com/openkaiden/kaiden/issues/1259

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "@openkaiden/kdn-api": "^0.1.11",
     "@openkaiden/workspace-configuration": "^0.1.11",
     "@openkaiden/api": "workspace:",
-    "@kortex-hub/mcp-registry-types": "^1.4.0-next.20251210140818.04eeb13",
+    "@openkaiden/mcp-registry-types": "^1.3.10",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^16.0.3",
     "@sveltejs/opencode": "^0.1.6",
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -86,7 +86,7 @@
     "@openkaiden/kdn-api": "^0.1.11",
     "@openkaiden/workspace-configuration": "^0.1.11",
     "@openkaiden/api": "workspace:",
-    "@kortex-hub/mcp-registry-types": "^1.4.0-next.20251210140818.04eeb13",
+    "@openkaiden/mcp-registry-types": "^1.3.10",
```
gastoner, line 89: Looks to me like we are downgrading the version?

gastoner (commented):
Otherwise LGTM

package.json
benoitf: in fact the tag was not published, you used the first version from the next build but we should use tag so I published a good one

benoitf: I've fix the issue (adding correct token) so creating a tag publishes a npmjs package/release

packages/main/src/plugin/extension/extension-loader.ts
```diff
@@ -19,9 +19,9 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 
-import { components } from '@kortex-hub/mcp-registry-types';
 import type * as containerDesktopAPI from '@openkaiden/api';
 import { ChunkProvider, RegisterServerResult } from '@openkaiden/api';
+import { components } from '@openkaiden/mcp-registry-types';
```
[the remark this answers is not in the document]
benoitf: I'm only changing the name of the import, not its type

jeffmaury (approved):

PR #1324, opened 2026-04-13
fix: align chat view background with the rest of the app

fbricon:
The chat dark mode background was near-black instead of dark gray like
all other views. Update --background to use --pd-content-bg, and fix
the input box toolbar having a different shade than the textarea.

Before:
Screenshot 2026-04-13 at 10 34 30

After:
Screenshot 2026-04-13 at 10 32 42

### DIFF
--- a/packages/renderer/src/app.css
+++ b/packages/renderer/src/app.css
@@ -60,7 +60,7 @@ a {
 }
 
 .dark {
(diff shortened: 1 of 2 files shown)


MarsKubeX (approved):
LGTM

PR #1365, opened 2026-04-16
fix: skills import dialog error handling and validation

### LINKED ISSUE #1352: Importing a Skill from a .md file requires adding skill instructions
### Bug description

When trying to import a new skill through the Create Skill dialog, you can't continue without filling the 'Content' field. Isn't this content what I'm providing in the selected .md file? Is this input overwriting my .md file?

Image

### Operating system

Windows 10

### Installation Method

### Version

0.0.3

### Steps to reproduce

- Go to Skills
- Click on New skill
- Choose an .md file
- Fill the 'Name' and 'Description' fields
- You should be able to continue but until you fill something in 'Content' you can't

### Relevant log output

```shell

```

### Additional context

_No response_

serbangeorge-m:
- Handle files without YAML frontmatter gracefully
- Show errors to user instead of silently failing
- Fix Create button validation for file imports
- Fix drag-and-drop passing invalid sourcePath
- Clean up skill directory on creation failure

Fixes https://github.com/openkaiden/kaiden/issues/1352


.agents/skills/ui-components/SKILL.md
```diff
@@ -1,3 +1,8 @@
+---
+name: ui-components
+description: UI Component Development Guide for Kortex
```
jeffmaury: **praise:** 
```suggestion
description: UI Component Development Guide for Kaiden
```

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -362,9 +371,15 @@ export class SkillManager {
    */
   async getSkillFileContent(filePath: string): Promise<SkillFileContent> {
     const rawContent = (await readFile(filePath, 'utf-8')).trimStart();
-    const metadata = this.extractFrontmatter(rawContent, filePath);
-    const body = this.stripFrontmatter(rawContent);
-    return { name: metadata.name, description: metadata.description, content: body };
+    try {
+      const metadata = this.extractFrontmatter(rawContent, filePath);
+      const body = this.stripFrontmatter(rawContent);
+      return { name: metadata.name ?? '', description: metadata.description ?? '', content: body };
```
jeffmaury: **question:** name and description are mandatory so why do we need to default them

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -362,9 +371,15 @@ export class SkillManager {
...
+    } catch {
+      // File has no valid frontmatter — return raw content so the user
+      // can fill in name/description manually in the dialog.
+      return { name: '', description: '', content: rawContent };
```
jeffmaury: **issue:** should be handled on the called side as otherwise there will be no way to detect errors

jeffmaury (changes requested):
Should be split in 2 PRs and the update the skill seeems invalid

serbangeorge-m:
> Should be split in 2 PRs and the update the skill seeems invalid

currently the importing is failing if the skills do not have the yaml frontmatter, that's why I updated the ui-components/SKILL.md

For the manual creation of the skill the backend generates the yaml frontmatter. Should we accept/import a skill without yaml frontmatter?

packages/main/src/plugin/skill/skill-manager.ts
serbangeorge-m: agree, defaults are unnecessary

packages/main/src/plugin/skill/skill-manager.ts
serbangeorge-m: reverted to throw on missing frontmatter, handled on the caller side. Both browse and drag and drop show same error message for consistency. Also added scroll into view so the error is visible otherwise will be hidden if you do not scroll all the way down.

Screenshot 2026-04-16 at 11 50 51 AM

fbricon:
This PR fixes the SKL-IMPORT e2e tests failing on all other PRs

fbricon (approved):
LGTM, but would be good to fix coderabbit's stuff

jeffmaury:
> > Should be split in 2 PRs and the update the skill seeems invalid
> 
> currently the importing is failing if the skills do not have the yaml frontmatter, that's why I updated the ui-components/SKILL.md
> 
> For the manual creation of the skill the backend generates the yaml frontmatter. Should we accept/import a skill without yaml frontmatter?

No it is mandatory so if it's used by E2E tests that's ok then

jeffmaury (approved):

PR #1401, opened 2026-04-20
feat: add recursive resource tree to skill details page

### LINKED ISSUE #1371: Update Skill Details Resources to show recursive tree
### Task content

Kaiden should now show a recursive tree for resources

bmahabirbu:
## Summary
- Replace the flat file list in the skill details Resources tab with a recursive tree view
- Folders are expandable/collapsible with chevron icons, children indented per depth level
- New `SkillResourceEntry` type with recursive `children` field replaces flat `string[]` API
- Backend `readDirectoryTree` recurses into subdirectories automatically

Fixes: https://github.com/openkaiden/kaiden/issues/1371

## Test plan
- [ ] Navigate to a skill with nested resource directories and verify the tree renders correctly
- [ ] Click folders to expand/collapse and verify children show/hide

### DIFF
--- a/packages/main/src/plugin/skill/skill-manager.ts
+++ b/packages/main/src/plugin/skill/skill-manager.ts
@@ -39,6 +39,7 @@ import {
   type SkillFolderInfo,
   type SkillInfo,
   type SkillMetadata,
+  type SkillResourceEntry,
 } from '/@api/skill/skill-info.js';
 
 const RESERVED_WORDS = ['anthropic', 'claude'];
@@ -128,9 +129,12 @@ export class SkillManager {
       return this.getSkillContent(name);
     });
 
-    this.ipcHandle('skill-manager:listSkillFolderContent', async (_listener, name: string): Promise<string[]> => {
-      return this.listSkillFolderContent(name);
-    });
+    this.ipcHandle(
+      'skill-manager:listSkillFolderContent',
+      async (_listener, name: string, relativePath?: string): Promise<SkillResourceEntry[]> => {
+        return this.listSkillFolderContent(name, relativePath);
+      },
+    );
 
     this.ipcHandle(
       'skill-manager:getSkillFileContent',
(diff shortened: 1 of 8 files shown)


packages/renderer/src/lib/skills/SkillResourceTree.svelte
```diff
@@ -0,0 +1,49 @@
...
+      </div>
+      <span class="text-sm font-medium text-[var(--pd-content-text)] font-mono">{entry.name}/</span>
+    </button>
+    {#if expanded[entry.name] && entry.children?.length}
+      <SkillResourceTree entries={entry.children} depth={depth + 1} />
+    {/if}
+  {:else}
+    <div
+      class="flex items-center gap-3 px-5 py-3 border-b border-[var(--pd-content-card-border)] last:border-b-0 hover:bg-[var(--pd-content-card-hover-bg)]"
+      style="padding-left: {20 + 16 + depth * 20}px;">
```
gastoner: Nit: I think we should be able to use some getLeftPaddingClass() in `class` prop instead of style ? maybe we can use derived directly ?

gastoner (commented):
Left nit

gastoner:
Cant expand the folder :(

packages/renderer/src/lib/skills/SkillResourceTree.svelte
bmahabirbu: Good call

bmahabirbu:
@gastoner try importing a new skill and see if it works? Maybe its a os thing ill give it another shot on my linux machine and test it on windows

packages/main/src/plugin/skill/skill-manager.ts
```diff
@@ -377,10 +381,23 @@ export class SkillManager {
...
   /** Lists all file/directory names inside the skill's folder. Directories have a trailing slash. */
-  async listSkillFolderContent(name: string): Promise<string[]> {
+  async listSkillFolderContent(name: string): Promise<SkillResourceEntry[]> {
     const skill = this.findSkillByName(name);
-    const entries = await readdir(skill.path, { withFileTypes: true });
-    return entries.map(entry => (entry.isDirectory() ? `${entry.name}/` : entry.name));
+    return this.readDirectoryTree(skill.path);
+  }
+
+  private async readDirectoryTree(dirPath: string): Promise<SkillResourceEntry[]> {
```
jeffmaury: **issue:** if the hierarchy of the skill is deep, this will cause a lot of recursing as its fully loaded. Can't we load through lazy loading (level by level) ?

packages/renderer/src/lib/skills/SkillDetails.svelte
```diff
@@ -50,6 +49,17 @@ $effect(() => {
     .catch((err: unknown) => console.error('Unexpected error loading skill details:', err));
 });
 
+function countEntries(entries: SkillResourceEntry[]): number {
```
jeffmaury: **suggestion:** should rather be in a separate helper ts file

packages/renderer/src/lib/skills/SkillResourceTree.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
```
jeffmaury, line 1: **issue:** missing unit test for this file

jeffmaury (approved):
LGTM.

Tested on Win11Pro

packages/main/src/plugin/skill/skill-manager.ts
bmahabirbu: Thats what I was thinking as well and I ran into this issue testing a what if scenario of having a skill be embedded in a deep directory. Kaiden became somewhat unresponsive 

I checked and most well written skills would not have such a deep directory but maybe its better to be safe than sorry or have a depth limit

gastoner (commented):
Sorry was probably worng branch that I tested on :D works fine on Mac 
Personally I would keep the folders expanded by default? WDYT ?

bmahabirbu:
@gastoner going to implement lazy load so i think that means by default we want the inner folders closed

bmahabirbu:
@coderabbit resolve

PR #1426, opened 2026-04-22
feat: add llmMetadata to inference connection

### LINKED ISSUE #1425: Add provider information to inference connection

Some agents need the notion of remote provider.

Expose it as part of an, optional field of the inference connection

_No response_

### Additional context

_No response_

jeffmaury:
Fixes #1425

### DIFF
--- a/extensions/gemini/src/gemini.spec.ts
+++ b/extensions/gemini/src/gemini.spec.ts
@@ -197,6 +197,7 @@ describe('factory', () => {
     expect(PROVIDER_MOCK.registerInferenceProviderConnection).toHaveBeenCalledWith({
       name: 'dum*****',
       type: 'cloud',
+      providerName: 'gemini',
       status: expect.any(Function),
       lifecycle: {
         delete: expect.any(Function),
--- a/extensions/gemini/src/gemini.spec.ts
+++ b/extensions/gemini/src/gemini.spec.ts
@@ -197,7 +197,7 @@ describe('factory', () => {
     expect(PROVIDER_MOCK.registerInferenceProviderConnection).toHaveBeenCalledWith({
       name: 'dum*****',
       type: 'cloud',
-      providerName: 'gemini',
+      llmMetadata: { name: 'gemini' },
       status: expect.any(Function),
       lifecycle: {
         delete: expect.any(Function),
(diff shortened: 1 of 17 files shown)


benoitf (commented):
I was wondering why the extension would need to provide the name of the provider as we're calling a method on the provider


I think from client side/extension side we don't change the register method, but then it's when we grab the connection object that we can have this extra field coming from its parent the provider

Because here I could call registerConnection with a different provider name than the one we have

benoitf:
I think it's like for https://github.com/openkaiden/kaiden/blob/fc186b4b4bf4034d5cea180a7b8876e63d355f89/packages/main/src/plugin/provider-registry.ts#L1781

but we add providerName in addition of providerId (I would assume that you might still find the provider's name by it's providerId as well no) ?


for inferenceConnections it would be there:
https://github.com/openkaiden/kaiden/blob/fc186b4b4bf4034d5cea180a7b8876e63d355f89/packages/main/src/plugin/provider-registry.ts#L1909

jeffmaury:
> I think it's like for
> 
> https://github.com/openkaiden/kaiden/blob/fc186b4b4bf4034d5cea180a7b8876e63d355f89/packages/main/src/plugin/provider-registry.ts#L1781
> 
> but we add providerName in addition of providerId (I would assume that you might still find the provider's name by it's providerId as well no) ?
> 
> for inferenceConnections it would be there:
> 
> https://github.com/openkaiden/kaiden/blob/fc186b4b4bf4034d5cea180a7b8876e63d355f89/packages/main/src/plugin/provider-registry.ts#L1909

Yes but here we're dealing with other notion of providers: there are the LLM providers so the ramalama extension might specify openai as the providerName there as it's likely how it will be seen from the agent POV
Maybe I should rename it to llmProvider or modelProvider as it's used to land into modelId

benoitf:
yes please use another name then.

Also I think it might be a free-form entry because you may want to add many fields

maybe llmMetadata

and llmMetadata can have name, version, whatever

benoitf (approved):

MarsKubeX (approved):

jeffmaury:
@coderabbitai resolve

PR #1431, opened 2026-04-23
feat/onboarding agent step

### LINKED ISSUE #1316: Add default agent selection as part of the onboarding

The default agent should be set during the onboarding flow

It is now the first step of the onboarding

From the mockup page: [openkaiden/kortex-mockups@sprint-290-demo-flow/onboarding/index.html](https://github.com/openkaiden/kortex-mockups/blob/sprint-290-demo-flow/onboarding/index.html?rgh-link-date=2026-04-10T17%3A35%3A40.000Z)

Image

_No response_

### Additional context

_No response_

MarsKubeX:
## Summary

- Add the coding agent selection step (step 1) to the onboarding wizard with an OpenCode agent tile and local runtime probe (Ollama/Ramalama detection)
- Introduce a two-layer architecture: an `agent-registry.ts` for tile metadata and separate panel components per agent, so adding a new agent is just one registry entry + an optional panel
- Filter available agents at runtime via `window.getCliInfo()` (backed by `kdn info --output json`), falling back to all registry entries when the CLI is unavailable or returns no agents
- Persist the selected default agent to `settings.json` via `ConfigurationRegistry` when the wizard completes, so workspace creation can read it later
- Add `required` prop to `CardSelector` to prevent deselection in single-choice contexts

> **Note:** Providers that require API keys are not included in this step until the credentials system redesign is complete.

https://github.com/user-attachments/assets/b1f12e5a-2e6e-4079-ba9c-403149b1f403

## Test plan

- [ ] Open the app for the first time (or reset the welcome flag / delete `settings.json`) so the onboarding wizard shows
- [ ] Verify OpenCode is selected by default with the "Recommended" badge
- [ ] Select OpenCode — confirm the local runtime probe runs (shows "Checking…", then "detected" or "not found" depending on whether Ollama/Ramalama is running)
- [ ] Click "Check again" after starting Ollama — confirm status updates to "detected"
- [ ] Complete the wizard — confirm `onboarding.defaultAgent` is written to `settings.json`
- [ ] Complete the wizard with no kdn installed — confirm it still closes gracefully (fallback shows all agents, persistence error is caught)

Closes #1316

### DIFF
--- a/packages/main/src/plugin/onboarding/onboarding-init.ts
+++ b/packages/main/src/plugin/onboarding/onboarding-init.ts
@@ -35,7 +35,7 @@ export class OnboardingInit {
         [`${OnboardingSettings.SectionName}.${OnboardingSettings.DefaultAgent}`]: {
           description: 'Default coding agent selected during onboarding',
           type: 'string',
-          default: 'opencode',
+          default: '',
           hidden: true,
         },
       },
--- a/packages/main/src/plugin/onboarding/onboarding-init.ts
+++ b/packages/main/src/plugin/onboarding/onboarding-init.ts
@@ -1,0 +1,46 @@
+/**********************************************************************
+ * Copyright (C) 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { inject, injectable } from 'inversify';
+
+import { type IConfigurationNode, IConfigurationRegistry } from '/@api/configuration/models.js';
+
+import { OnboardingSettings } from './onboarding-settings.js';
+
+@injectable()
+export class OnboardingInit {
+  constructor(@inject(IConfigurationRegistry) private configurationRegistry: IConfigurationRegistry) {}
+
+  init(): void {
+    const onboardingConfiguration: IConfigurationNode = {
+      id: 'preferences.onboarding',
+      title: 'Onboarding',
+      type: 'object',
+      properties: {
+        [`${OnboardingSettings.SectionName}.${OnboardingSettings.DefaultAgent}`]: {
+          description: 'Default coding agent selected during onboarding',
+          type: 'string',
+          default: 'opencode',
+          hidden: true,
+        },
+      },
+    };
+
+    this.configurationRegistry.registerConfigurations([onboardingConfiguration]);
+  }
+}
--- a/packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
+++ b/packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
@@ -1,0 +1,149 @@
+/**********************************************************************
+ * Copyright (C) 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import '@testing-library/jest-dom/vitest';
+
+import { fireEvent, render, screen, waitFor } from '@testing-library/svelte';
+import { beforeEach, expect, test, vi } from 'vitest';
+
+import CodingAgentStep from './CodingAgentStep.svelte';
+import type { OnboardingState } from './guided-setup-steps';
+import { createDefaultOnboardingState } from './guided-setup-steps';
+
+let onboarding: OnboardingState;
+
+function stubLocalInference(hasConnection: boolean): void {
+  const providers = hasConnection
+    ? [{ id: 'ollama', inferenceConnections: [{ type: 'local', status: 'started' }] }]
+    : [{ id: 'ollama', inferenceConnections: [] }];
+  (window as unknown as Record<string, unknown>).getProviderInfos = vi.fn().mockResolvedValue(providers);
+}
+
+beforeEach(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+  vi.resetAllMocks();
+  onboarding = createDefaultOnboardingState();
+  vi.stubGlobal(
+    'getCliInfo',
+    vi.fn().mockResolvedValue({ version: '0.1.0', agents: ['opencode'], runtimes: ['podman'] }),
+  );
+  stubLocalInference(false);
+});
+
+function renderStep(overrides: Partial<OnboardingState> = {}): void {
+  Object.assign(onboarding, overrides);
+  render(CodingAgentStep, {
+    stepId: 'coding-agent',
+    title: 'Choose your coding agent',
+    description: 'Pick one runtime for kdn.',
+    onboarding,
+  });
+}
+
+test('renders the OpenCode agent tile', () => {
+  renderStep();
+
+  expect(screen.getByRole('button', { name: 'OpenCode' })).toBeInTheDocument();
+});
+
+test('OpenCode is selected by default', () => {
+  renderStep();
+
+  const tile = screen.getByRole('button', { name: 'OpenCode' });
+  expect(tile.className).toContain('border-[var(--pd-content-card-border-selected)]');
+});
+
+test('shows Recommended badge on OpenCode', () => {
+  renderStep();
+
+  expect(screen.getByText('Recommended')).toBeInTheDocument();
+});
+
+test('shows local runtime panel when OpenCode is selected', () => {
+  renderStep();
+
+  expect(screen.getByTestId('opencode-panel')).toBeInTheDocument();
+  expect(screen.getByText('Local Runtime')).toBeInTheDocument();
+});
+
+test('shows probe checking state on mount with OpenCode selected', async () => {
+  (window as unknown as Record<string, unknown>).getProviderInfos = vi.fn().mockReturnValue(new Promise(() => {}));
+
+  renderStep();
+
+  await waitFor(() => {
+    expect(screen.getByTestId('probe-checking')).toBeInTheDocument();
+  });
+});
+
+test('shows detected state when a local inference connection exists', async () => {
+  stubLocalInference(true);
+
+  renderStep();
+
+  await waitFor(() => {
+    expect(screen.getByTestId('probe-detected')).toBeInTheDocument();
+  });
+});
+
+test('shows not-found state when no local inference connections exist', async () => {
+  renderStep();
+
+  await waitFor(() => {
+    expect(screen.getByTestId('probe-not-found')).toBeInTheDocument();
+  });
+});
+
+test('Check again button retries the probe', async () => {
+  renderStep();
+
+  await waitFor(() => {
+    expect(screen.getByTestId('probe-not-found')).toBeInTheDocument();
+  });
+
+  stubLocalInference(true);
+
+  await fireEvent.click(screen.getByRole('button', { name: 'Check again' }));
+
+  await waitFor(() => {
+    expect(screen.getByTestId('probe-detected')).toBeInTheDocument();
+  });
+});
+
+test('step title and description are rendered', () => {
+  renderStep();
+
+  expect(screen.getByText('Choose your coding agent')).toBeInTheDocument();
+  expect(screen.getByText('Pick one runtime for kdn.')).toBeInTheDocument();
+});
+
+test('card selector region has correct aria-label', () => {
+  renderStep();
+
+  expect(screen.getByRole('region', { name: 'Coding agent' })).toBeInTheDocument();
+});
+
+test('falls back to all registry entries when getCliInfo fails', async () => {
+  vi.stubGlobal('getCliInfo', vi.fn().mockRejectedValue(new Error('kdn not found')));
+
+  renderStep();
+
+  await waitFor(() => {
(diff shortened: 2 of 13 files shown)


packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+beforeEach(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+  vi.resetAllMocks();
+  onboarding = createDefaultOnboardingState();
+  vi.stubGlobal(
+    'getCliInfo',
+    vi.fn().mockResolvedValue({ version: '0.1.0', agents: ['opencode'], runtimes: ['podman'] }),
+  );
+  stubLocalInference(false);
+});
```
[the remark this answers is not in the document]
MarsKubeX: vi.stubGlobal for getProviderInfos: Fixed. Replaced direct window assignments with vi.stubGlobal for consistency and proper test isolation.

packages/renderer/src/lib/guided-setup/CodingAgentStep.svelte
```diff
@@ -0,0 +1,66 @@
...
+let selectedAgent = $state(untrack(() => onboarding.agent));
+let activeDefinition: AgentDefinition | undefined = $derived(definitionsByAgent.get(selectedAgent));
+
+$effect(() => {
+  if (definitionsByAgent.has(selectedAgent)) {
+    onboarding.agent = selectedAgent as CliAgent;
+  } else if (filteredDefinitions.length > 0) {
+    selectedAgent = filteredDefinitions[0]!.cliName;
+  }
+});
```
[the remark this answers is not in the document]
MarsKubeX: Fixed. filteredDefinitions now falls back to the full registry when the filter produces an empty result, preventing a blank card area if the CLI advertises agents unknown to the UI.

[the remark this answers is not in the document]
MarsKubeX: The cast is safe by construction — selectedAgent can only hold a value from agentDefinitions[].cliName, which is typed as CliAgent. Deriving CliAgent from the registry would invert the dependency: CliAgent represents CLI agent names (used by kdn init --agent), so the registry must conform to it, not the other way around. The registry is always a subset of what the CLI supports.

packages/renderer/src/lib/guided-setup/GuidedSetup.spec.ts
```diff
@@ -184,8 +189,57 @@ test('stepper progress bar has correct number of connector lines', () => {
...
+  await fireEvent.click(screen.getByRole('button', { name: /Continue/ }));
+  await fireEvent.click(screen.getByRole('button', { name: /Go to Dashboard/ }));
+  await vi.advanceTimersByTimeAsync(0);
+
+  expect(consoleErrorSpy).toHaveBeenCalledWith('Failed to persist onboarding defaults', persistError);
+  expect(closeMock).toHaveBeenCalled();
+
+  consoleErrorSpy.mockRestore();
+  vi.stubGlobal('updateConfigurationValue', vi.fn().mockResolvedValue(undefined));
+});
```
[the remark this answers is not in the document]
MarsKubeX: Fixed. Removed the unnecessary consoleErrorSpy.mockRestore() and vi.stubGlobal re-stub at the end of the test — beforeEach with vi.resetAllMocks() already handles both.

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,102 @@
...
+
+function openRamalamaLink(): void {
+  window.openExternal('https://github.com/containers/ramalama').catch(() => {});
+}
+
+function handleRetryProbe(): void {
+  probeLocalRuntime().catch((err: unknown) => {
+    console.error('Local runtime probe failed', err);
+  });
+}
```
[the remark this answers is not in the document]
MarsKubeX: The .catch handlers are intentional defensive programming for fire-and-forget async calls in Svelte event handlers and effects. While hasLocalInferenceConnection currently catches internally, removing the outer .catch would create unhandled promise rejections if that function is ever refactored to propagate errors. The cost is negligible and the safety net is worth keeping.

bmahabirbu (approved):
LGTM!!

jeffmaury (commented):
Maybe I missed something ?:
- why only opencode is listed ?
- UI is not readable in light mode

Image

MarsKubeX:
> Maybe I missed something ?:
> 
> * why only opencode is listed ?
> * UI is not readable in light mode
> 
> Image

Thanks for the feedback @jeffmaury. 
- For the UI not readable in light mode, is fixed:
Captura de pantalla 2026-04-24 a las 12 23 13

- For the other agent providers, since they need API Keys, I prefer to develop that in other issue because of the credentials system redesign and how we should retrieve models in step 2. #1439

packages/main/src/plugin/onboarding/onboarding-init.ts
```diff
@@ -0,0 +1,46 @@
...
+import { type IConfigurationNode, IConfigurationRegistry } from '/@api/configuration/models.js';
+
+import { OnboardingSettings } from './onboarding-settings.js';
+
+@injectable()
+export class OnboardingInit {
+  constructor(@inject(IConfigurationRegistry) private configurationRegistry: IConfigurationRegistry) {}
+
+  init(): void {
+    const onboardingConfiguration: IConfigurationNode = {
```
gastoner, line 30: NIT: Thinking about usage of preDestroy (will be empty) and postInit? WDYT

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,102 @@
...
+
+$effect(() => {
+  if (probeStatus === 'idle') {
+    probeLocalRuntime().catch((err: unknown) => {
+      console.error('Local runtime probe failed', err);
+    });
+  }
+});
```
gastoner: Thinking if we can convert this to derived or derivedBy

packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  vi.stubGlobal('getProviderInfos', vi.fn().mockResolvedValue(providers));
+}
+
+beforeEach(() => {
+  vi.useFakeTimers({ shouldAdvanceTime: true });
+  vi.resetAllMocks();
+  onboarding = createDefaultOnboardingState();
+  vi.stubGlobal(
+    'getCliInfo',
+    vi.fn().mockResolvedValue({ version: '0.1.0', agents: ['opencode'], runtimes: ['podman'] }),
```
gastoner, line 55: Nit: I think we can use vi.mocked(window.getCliInfo).mockRe.... I think it is a prefered way?

packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  expect(screen.getByText('Choose your coding agent')).toBeInTheDocument();
+  expect(screen.getByText('Pick one runtime for kdn.')).toBeInTheDocument();
+});
+
+test('card selector region has correct aria-label', () => {
+  renderStep();
+
+  expect(screen.getByRole('region', { name: 'Coding agent' })).toBeInTheDocument();
+});
```
gastoner, line 167: Personally, I would connect all tests that do the pure rendering and then expect

packages/renderer/src/lib/guided-setup/CodingAgentStep.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+let filteredDefinitions = $derived.by(() => {
+  if (!cliAgents) return agentDefinitions;
+  const filtered = agentDefinitions.filter(d => cliAgents!.includes(d.cliName));
+  return filtered.length > 0 ? filtered : agentDefinitions;
+});
+
+const definitionsByAgent = $derived(new Map<string, AgentDefinition>(filteredDefinitions.map(d => [d.cliName, d])));
```
gastoner, line 31: I think we can use SvelteMap here? insteda of having Map in derived

packages/renderer/src/lib/guided-setup/GuidedSetup.spec.ts
```diff
@@ -133,29 +137,30 @@ test('skipped steps are not marked completed', async () => {
...
-  await fireEvent.click(screen.getByRole('button', { name: 'Continue' }));
+  await fireEvent.click(screen.getByRole('button', { name: /Continue/ }));
+  await fireEvent.click(screen.getByRole('button', { name: /Continue/ }));
 
-  const dashboardButton = screen.getByRole('button', { name: 'Go to Dashboard' });
+  const dashboardButton = screen.getByRole('button', { name: /Go to Dashboard/ });
   await fireEvent.click(dashboardButton);
+  await vi.advanceTimersByTimeAsync(0);
 
   expect(closeMock).toHaveBeenCalled();
```
gastoner: I think having this is vi.waitfor should be enough, so we dont need the adviceTimersByTime...

gastoner (commented):
Left some nitpicks, otherwirse tried on Mac and working fine

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,102 @@
...
+import { faCircleCheck, faTriangleExclamation } from '@fortawesome/free-solid-svg-icons';
+import { Button, Link, Spinner } from '@podman-desktop/ui-svelte';
+import { Icon } from '@podman-desktop/ui-svelte/icons';
+
+type ProbeStatus = 'idle' | 'checking' | 'detected' | 'not-found';
+let probeStatus: ProbeStatus = $state('idle');
+
+async function hasLocalInferenceConnection(): Promise<boolean> {
+  try {
+    const providers = await window.getProviderInfos();
```
jeffmaury: **suggestion:** I think there is a store for this. Look at model-utils.ts

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,102 @@
...
+    return false;
+  }
+}
+
+async function probeLocalRuntime(): Promise<void> {
+  probeStatus = 'checking';
+  probeStatus = (await hasLocalInferenceConnection()) ? 'detected' : 'not-found';
+}
+
+function openOllamaLink(): void {
```
jeffmaury, line 14: **question:** questioning the non genericity of this should be have them from extension or provider ?

packages/main/src/plugin/onboarding/onboarding-init.ts
MarsKubeX: OnboardingInit only registers a static config node, there's nothing to clean up in preDestroy, and init() has no dependencies on other services being ready, so postInit isn't needed either. I prefer to keep it lean without and empty preDestroy()

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
MarsKubeX: I think $derived cannot contain async side-effects, and here there is the async ``probeLocalRuntime()``

packages/renderer/src/lib/guided-setup/CodingAgentStep.svelte
MarsKubeX: Since definitionsByAgent is recomputed by $derived whenever filteredDefinitions changes (never mutated in place), a plain Map is correct and more efficient. I think SvelteMap is more useful when you need fine-grained reactivity on mutations to the map itself, which doesn't apply here.

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
MarsKubeX: Good idea. The only problem is that Ollama and Ramalama extensions are nort passing links to the ``createProvider``, so the link array is empty. I've created issue #1482 to add them.

packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts
MarsKubeX: All the rendering test should be under the same describe now.

packages/renderer/src/lib/guided-setup/GuidedSetup.spec.ts
MarsKubeX: Should use vi.waitfor now.

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
MarsKubeX: Yes, updated to use model.utils.

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,90 @@
...
+function handleRetryProbe(): void {
+  checking = true;
+  fetchProviders()
+    .catch((err: unknown) => {
+      console.error('Local runtime probe failed', err);
+    })
+    .finally(() => {
+      checking = false;
+    });
+}
```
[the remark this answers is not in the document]
MarsKubeX: The panel now derives hasLocalInference reactively from the providerInfos store (via EventStore), which is populated on system-ready during app startup -- well before the user can reach the onboarding wizard. There's no window where the store is empty, so an initial probe on mount would be a redundant fetchProviders() call. The "Checking..." spinner is intentionally reserved for explicit "Check again" retries.

packages/renderer/src/lib/guided-setup/panels/OpenCodePanel.svelte
```diff
@@ -0,0 +1,90 @@
...
+  $providerInfos.some(p => p.inferenceConnections.some(c => c.type === 'local' && c.status === 'started')),
+);
+
+function openOllamaLink(): void {
+  window.openExternal('https://ollama.com').catch(() => {});
+}
+
+function openRamalamaLink(): void {
+  window.openExternal('https://github.com/containers/ramalama').catch(() => {});
+}
```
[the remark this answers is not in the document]
MarsKubeX: The Link component from @podman-desktop/ui-svelte doesn't accept href -- on:click with window.openExternal() is the standard pattern used across the codebase (e.g., ProviderLinks.svelte, WelcomePage.svelte). There's no setupSecurityRestrictionsOnLinks in the renderer.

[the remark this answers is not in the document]
MarsKubeX: @coderabbitai, the Link component from podman-desktop/ui-svelte doesn't accept href -- on:click with window.openExternal() is the standard pattern used across the codebase (e.g., ProviderLinks.svelte, WelcomePage.svelte). There's no setupSecurityRestrictionsOnLinks in the renderer.

gastoner (approved):
Thank you for the changes

PR #1494, opened 2026-04-29
refactor(loader): replace animated character with static logo SVG

### LINKED ISSUE #1491: Kaiden-themed loader

Now that we have the new Kaiden icon/logo, we also need the new loader.

—

_No response_

### Additional context

_No response_

vancura:
Swap the animated loader (rotating dots, floating head PNG assets, blinking eyes) for a static inline SVG of the Kaiden logo at 50% opacity. Remove all animation infrastructure that is no longer needed.

`Loader.svelte`:
- remove toggle state and `loadingSequence` interval (drove the old animation tick)
- remove all `clearInterval` calls for `loadingSequence`
- remove "Initializing..." heading from the loading screen template

`LoaderAnimation.svelte`:
- replace full animated component (script, PNG imports, CSS keyframes, SVG defs/clipPaths) with a plain inline SVG at 300x339px, opacity 0.5

`LoaderAnimation.spec.ts`:
- drop animation-specific tests (dots, keyframes, blinking, size props)
- replace with five focused tests: SVG renders, correct viewBox, role=status, sr-only text, aria-hidden

---

- Resolves: #1491

---

Light theme:

Image

Dark theme:

Image

(Same SVG for both)

Co-Authored-By: Claude <noreply@anthropic.com>


benoitf (approved):

packages/renderer/src/Loader.svelte
```diff
@@ -92,7 +80,6 @@ window.events.receive('starting-extensions', (value: unknown) => {
   <main class="flex flex-row w-screen h-screen justify-center" style="-webkit-app-region: drag;">
     <div class="flex flex-col justify-center">
       <LoaderAnimation />
-      <h1 class="text-center text-xl">Initializing...</h1>
```
benoitf, line 95: I think it's better to say that something is happening and not remove it

MarsKubeX (approved):
LGTM. Tested on MacOS

packages/renderer/src/Loader.svelte
vancura: The `Loader` is on screen for less than half a second, so nobody reads it really. A headline, even a short one, may cause FOMO: "What did I just miss? Was it important?"

More importantly, "Initializing..." implies a technical process that could fail - it creates mild anxiety. A logo is neutral and brand-affirming: users immediately understand it as a splash screen, not a status message.

The act of showing a splash screen at all is already the signal that something is happening - text saying "Initializing..." adds no new information, it just repeats the obvious.

If there is just the picture, the user will know it was only an illustration, nothing they need to worry about. Illustrations are much easier and faster to read than any text.

benoitf: in rare case the screen is displayed forever (so at least you know it was to do something)

vancura: In that case, it would be better to show the headline "Initializing..." after a short while, not instantly. Reduces mental weight - and flicker.

vancura: CleanShot 2026-04-29 at 13 43 15@2x

This is now displayed after two seconds and stays then.

Commit: fe1a743688345fc988f3c6a8cbd3e1b6875fe9b1

packages/renderer/src/Loader.svelte
```diff
@@ -8,17 +8,16 @@ import ColorsStyle from './lib/style/ColorsStyle.svelte';
 import { lastPage } from './stores/breadcrumb';
 
 let systemReady = $state(false);
+let showTitle = $state(false);
 
-let toggle = $state(false);
-
-let loadingSequence: NodeJS.Timeout;
-
+let titleTimer: NodeJS.Timeout;
```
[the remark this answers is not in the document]
vancura: @coderabbitai: Finding 2a (type change): Not applied. The codebase already uses NodeJS.Timeout for extensionsStarterChecker on line 14 under the same conditions, so this is consistent with the existing pattern. Changing titleTimer alone would create an inconsistency; changing both is out of scope for this PR.

Finding 2b (null guards): Not applied. clearTimeout(undefined) is a safe no-op in browsers, and onDestroy always runs after onMount, so titleTimer is guaranteed to be assigned by the time it is cleared. The guard on extensionsStarterChecker is a pre-existing style choice, not a correctness requirement.

benoitf:
typecheck failure is unrelated, rebasing to include the fix

PR #1610, opened 2026-05-05
fix(container): activate podman socket on Linux when not running

bmahabirbu:
## Summary
- On Linux, the podman API socket is not always active — it requires systemd socket activation or Podman Desktop to be running
- Modified `PodmanSocketLinuxFinder` to spawn `podman system service --time=0` when the rootless socket is missing, matching Podman Desktop's behavior
- The spawned process is killed on extension deactivation via `@preDestroy()` dispose
- Includes Flatpak support (`flatpak-spawn --host`)

## Test plan
- [x] Unit tests updated and passing (11 tests)
- [ ] On Linux without Podman Desktop running, verify Kaiden discovers podman containers
- [ ] Verify socket process is cleaned up when Kaiden exits
- [ ] Verify no duplicate process spawned on repeated socket finder polls

### DIFF
--- a/extensions/container/packages/extension/src/helper/socket-finder/podman/podman-linux-finder.ts
+++ b/extensions/container/packages/extension/src/helper/socket-finder/podman/podman-linux-finder.ts
@@ -16,15 +16,22 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { ChildProcess } from 'node:child_process';
+import { spawn } from 'node:child_process';
 import { existsSync } from 'node:fs';
 import { resolve } from 'node:path';
 
-import { injectable } from 'inversify';
+import { injectable, preDestroy } from 'inversify';
 
 import type { SocketFinder } from '/@/api/socket-finder';
 
 @injectable()
 export class PodmanSocketLinuxFinder implements SocketFinder {
+  static readonly SOCKET_WAIT_ATTEMPTS = 50;
+  static readonly SOCKET_WAIT_INTERVAL_MS = 100;
+
+  #podmanProcess: ChildProcess | undefined;
+
   async findPaths(): Promise<string[]> {
     const paths: string[] = [];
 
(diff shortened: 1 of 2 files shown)


bmahabirbu:
This was needed because on linux it couldnt find the ramalama services because of this

bmahabirbu:
relates to https://github.com/openkaiden/kaiden/issues/1602

extensions/container/packages/extension/src/helper/socket-finder/podman/podman-linux-finder.ts
```diff
@@ -47,4 +57,42 @@ export class PodmanSocketLinuxFinder implements SocketFinder {
...
+  protected async startPodmanService(socketPath: string): Promise<void> {
+    if (this.#podmanProcess) {
+      return;
+    }
+
+    let command = 'podman';
+    let args = ['system', 'service', '--time=0'];
+    if (process.env.FLATPAK_ID) {
+      command = 'flatpak-spawn';
+      args = ['--host', 'podman', ...args];
```
jeffmaury, line 70: **praise:** should use exec from the extension API

jeffmaury (changes requested):
I'm questioning the rationale as we don't have this in Podman Desktop as ask user to run it so might have permissions issues. @benoitf might help

bmahabirbu:
@jeffmaury I noticed that in podman-desktop the socket opens and closes automatically on linux so thought id just try to add the functionality in kaiden too

benoitf:
so I know in Podman Desktop we start the service but users are complaining that sometimes they have the service already started so there are like 2 servers then

benoitf (commented):
overall I'm ok as it simplifies the default experience but as @jeffmaury said, it needs to use the exec API

MarsKubeX (commented):
Looks good in general but same comment as the others, use exec from the api extension

benoitf:
@bmahabirbu are you planning to fix this ? or we put the label 'on hold'

benoitf:
setting to draft as I didn't get a reply

bmahabirbu:
@benoitf I think we should close this since we are moving to vm driver WDYT?

benoitf:
for knowledge base we still need containers

bmahabirbu:
Right, okay ill update this one soon

PR #1686, opened 2026-05-06
chore: avoid to report catalog content is not there due to internet issue

benoitf:
the catalog is empty for now so there is no issue

follow-up of https://github.com/openkaiden/kaiden/issues/1589


fbricon (approved):

fbricon:
@coderabbit resolve

PR #1719, opened 2026-05-07
feat: rewrite host-native service URLs for WSL2 container access

### LINKED ISSUE #1718: Kaiden should expose services if local or host for windows
### Task content

WSL2 needs 2 ip addresses the host one and the vm one to target either container to agent container connction or host to agent container. 

We can assume connections are container -> container first as the default. If a service is running on windows host we need to add that somehow

bmahabirbu:
## Summary
- Add `hostNative` field to `InferenceProviderConnection` so extensions can declare they run on the OS host (not in a container)
- Ollama extension sets `hostNative: true`
- `getModelId()` rewrites `localhost` → `native-host.internal` for host-native services when building workspace model IDs
- Paired with [kdn#fix/wsl2-host-containers-internal] which injects `native-host.internal` → Windows gateway IP inside containers

Fixes: https://github.com/openkaiden/kaiden/issues/1718

On WSL2 with pasta networking, `host.containers.internal` resolves to the podman VM (169.254.1.2), not the Windows host. Host-native services like Ollama are unreachable through that address. `native-host.internal` is a sentinel hostname that kdn resolves to the correct IP per platform. On Mac/Linux this is a no-op since both resolve to the same place.

## Test plan
- [x] Unit tests for `rewriteHostNativeEndpoint()` and `getModelId()` with host-native vs container models
- [x] Unit test verifying Ollama extension registers with `hostNative: true`
- [ ] Manual: create workspace with Ollama model → verify `kdn init --model` receives `native-host.internal` in endpoint
- [ ] Manual: create workspace with RamaLama model → verify endpoint stays `localhost`

### DIFF
--- a/extensions/ollama/src/ollama-extension.spec.ts
+++ b/extensions/ollama/src/ollama-extension.spec.ts
@@ -70,6 +70,9 @@ describe('OllamaExtension', () => {
     expect(vi.mocked(provider.createProvider)).toHaveBeenCalled();
     expect(extensionContext.subscriptions).toContain(ollamaProvider);
     expect(vi.mocked(ollamaProvider.registerInferenceProviderConnection)).toHaveBeenCalled();
+    expect(vi.mocked(ollamaProvider.registerInferenceProviderConnection)).toHaveBeenCalledWith(
+      expect.objectContaining({ hostNative: true }),
+    );
     expect(vi.mocked(ollamaProvider.updateStatus)).toHaveBeenCalledWith('started');
   });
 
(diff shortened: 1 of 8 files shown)


bmahabirbu:
We cant do this in kdn because kdn doesnt know what services are local or host so we need a small addition here

bmahabirbu:
depends on https://github.com/openkaiden/kdn/pull/423 as well

jeffmaury (changes requested):
That's a hacky patch as it does cover only resources from Kaiden and not one specified manually through workspace.json.
Would prefer ollama to report IP based

bmahabirbu:
hmm so should i hardcode ollama to be native-host.internal on the kdn side then? and close this pr

bmahabirbu:
The real problem is we dont know if a service is container or windows host but you are right handle on the kdn side

maybe a simple probe is all we need closing for now

jeffmaury:
Closing at not needed because of #1773

PR #1749, opened 2026-05-07
feat(agent-workspace): add filesystem mount configuration

### LINKED ISSUE #1741: Path mounted should have an optional target destination
### Bug description

Path should have a optional destination and a ro/rw flag (ro by default)

### Operating system

Win11Pro

### Installation Method

None

### Version

0.0.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

### LINKED ISSUE #1744: Paths specified in file system are not wired into workspace configuration
### Bug description

We should allow to mount them in workspace

See #1741 

### Operating system

Win11Pro

### Installation Method

None

### Version

0.0.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

gastoner:
Replace the single-path custom access model with structured mounts (host, target, read-only). Preset modes (workspace, home, full, custom) map to concrete mount entries passed to kdn init via workspace.json.

Screenshot 2026-05-07 at 13 13 16

Closes https://github.com/openkaiden/kaiden/issues/1741
Closes #1744

### DIFF
--- a/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreateStepFileSystem.svelte
+++ b/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreateStepFileSystem.svelte
@@ -12,24 +12,30 @@ export interface FileAccessOption {
   badge?: string;
 }
 
+export interface CustomMount {
+  host: string;
+  target: string;
+  ro: boolean;
+}
+
 interface Props {
   fileAccessOptions: FileAccessOption[];
   selectedFileAccess: string;
-  customPaths: string[];
+  customMounts: CustomMount[];
   onBrowseCustomPath: (index: number) => Promise<void>;
-  onAddCustomPath: () => void;
-  onRemoveCustomPath: (index: number) => void;
-  onUpdateCustomPath: (index: number, value: string) => void;
+  onAddCustomMount: () => void;
+  onRemoveCustomMount: (index: number) => void;
+  onUpdateCustomMount: (index: number, field: keyof CustomMount, value: string | boolean) => void;
 }
 
 let {
   fileAccessOptions,
   selectedFileAccess = $bindable(),
-  customPaths,
+  customMounts,
   onBrowseCustomPath,
-  onAddCustomPath,
-  onRemoveCustomPath,
-  onUpdateCustomPath,
+  onAddCustomMount,
+  onRemoveCustomMount,
+  onUpdateCustomMount,
 }: Props = $props();
 
 function selectOption(value: string): void {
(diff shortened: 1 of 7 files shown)


jeffmaury (commented):
LGTM.

Not sure we might align with Podman Desktop:

Image

@slemeur @vancura WDYT ?

slemeur:
I actually think it's clearer the way it's done on Kaiden - let's go with @gastoner solution, we can still iterate later on.

slemeur (approved):

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreateStepFileSystem.svelte
```diff
@@ -85,21 +91,47 @@ function selectOption(value: string): void {
...
-        <Button onclick={(): Promise<void> => onBrowseCustomPath(index)} aria-label="Browse for directory" icon={faFolderOpen} />
-        {#if customPaths.length > 1}
-          <Button class="text-red-400" onclick={(): void => onRemoveCustomPath(index)}>Remove</Button>
-        {/if}
+    <p class="text-xs text-[var(--pd-content-card-text)] opacity-70 mb-3">Mount host directories into the workspace. Target is optional (defaults to the host path).</p>
+    {#each customMounts as mount, index (index)}
+      <div class="flex flex-col gap-2 mb-3 p-3 rounded-lg bg-[var(--pd-content-card-inset-bg)] border border-[var(--pd-content-card-border)]">
+        <div class="flex gap-3 items-center">
+          <div class="flex-1 min-w-0">
+            <span class="text-[11px] text-[var(--pd-content-card-text)] opacity-60 mb-1 block">Host path</span>
```
vancura: Both span labels (also on line 114) render at `text-[11px]` with `opacity-60`. At 11pt (below the 18pt / 14pt-bold thresholds) WCAG AA requires 4.5:1.

Don't use opacity. This is hard to track. Hardcoded opacities are a road to hell, and totally outside of `color-registry`.

vancura: "Host path" and "Target path in workspace" (below) are `<span>` elements visually acting as labels. The inputs themselves carry `aria-label`, so they're     not unlabeled, but the visible text (`span`) and the programmatic name (`aria-label`) are not linked.

The `aria-label` reads "Host path 1" / "Host path 2" etc., while the visible `span` just says "Host path".

If the label text ever changes from the `aria-label` string, it will silently misrepresent itself. Using `<label>` with for wired to the input's id would be cleaner and self-correcting.

vancura (changes requested):

gastoner:
rebased

benoitf:
e2e tests are not passing

PR #1770, opened 2026-05-07
fix(ramalama): use ramalama as provider id instead of openai

fbricon:
Mainly a cosmetic fix, it's notably displayed as the provider name in opencode on the lower right

Before:

Screenshot 2026-05-07 at 23 06 27

After:
Screenshot 2026-05-07 at 23 42 01

### DIFF
--- a/extensions/ramalama/src/manager/inference-model-manager.spec.ts
+++ b/extensions/ramalama/src/manager/inference-model-manager.spec.ts
@@ -70,7 +70,7 @@ describe('InferenceModelManager', () => {
       expect.objectContaining({
         name: `port/${modelPort}`,
         type: 'local',
-        llmMetadata: { name: 'openai' },
+        llmMetadata: { name: 'ramalama' },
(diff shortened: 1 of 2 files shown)


MarsKubeX (approved):
LGTM

jeffmaury (commented):
For me it breaks the concept of providerID as the original idea is to identify the API used to contact the LLM and adding new providerid increase the complexity

fbricon:
the ollama and ramalama providers allow users to simplify passing modelids on CLI, and infer base urls based on heuristics

jeffmaury:
> the ollama and ramalama providers allow users to simplify passing modelids on CLI, and infer base urls based on heuristics

Adding new providers augment the agent vs provider matrix which is already having a lot of holes

bmahabirbu:
Agree with Jeff here because all ramalama is doing is spawning a llama.cpp endpoint in a container that uses the openai spec

Same as if you would use VLLM to spawn it as well

fbricon:
Ok I yield

PR #1773, opened 2026-05-08
fix(ollama): proxy endpoint on Windows for container accessibility

### LINKED ISSUE #1756: Update ollama endpoints
### Task content

Change endpoints reported by ollama extension from localhost to local IP

bmahabirbu:
## Summary
- On Windows, Ollama binds to `127.0.0.1:11434` which is unreachable from containers running inside the Podman VM (WSL). Previously `localhost` was rewritten to `host.containers.internal` which routes to the WSL VM, not the Windows host where Ollama runs.
- Adds a lightweight HTTP reverse proxy (Windows only) in the Electron process that forwards requests to Ollama on loopback, and exposes them on the host's LAN IP so containers can route to it.
- Mac/Linux behavior is unchanged — containers can reach the host directly via `127.0.0.1`.

Fixes #1756

## Test plan
- [x] Unit tests pass (`pnpm test:unit extensions/ollama/src/ollama-extension.spec.ts`)
- [ ] Verify on Windows: endpoint registered with host LAN IP and proxy port
- [ ] Verify on Mac/Linux: endpoint remains `http://127.0.0.1:11434/v1`
- [ ] Verify agent workspace creation with Ollama model on Windows can reach Ollama from container

### DIFF
--- a/extensions/ollama/src/ollama-extension.ts
+++ b/extensions/ollama/src/ollama-extension.ts
@@ -15,20 +15,43 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { IncomingMessage, Server, ServerResponse } from 'node:http';
+import { createServer, request } from 'node:http';
+import { networkInterfaces } from 'node:os';
+
 import { type Disposable, type ExtensionContext, type Provider, provider } from '@openkaiden/api';
 import { createOllama } from 'ollama-ai-provider-v2';
+
+function getLocalIP(): string {
+  const nets = networkInterfaces();
+  for (const entries of Object.values(nets)) {
+    for (const entry of entries ?? []) {
+      if (entry.family === 'IPv4' && !entry.internal) {
+        return entry.address;
+      }
+    }
+  }
+  return '127.0.0.1';
+}
+
+const OLLAMA_HOST = '127.0.0.1';
+const OLLAMA_PORT = 11434;
 
 export class OllamaExtension {
   #extensionContext: ExtensionContext;
   #currentModels: string[] = [];
   #connectionDisposable: Disposable | undefined;
   #interval: NodeJS.Timeout | undefined;
+  #proxyServer: Server | undefined;
+  #proxyPort: number | undefined;
 
   constructor(extensionContext: ExtensionContext) {
     this.#extensionContext = extensionContext;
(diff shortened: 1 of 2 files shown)


extensions/ollama/src/ollama-extension.ts
```diff
@@ -15,20 +15,43 @@
...
   #interval: NodeJS.Timeout | undefined;
+  #proxyServer: Server | undefined;
+  #proxyPort: number | undefined;
 
   constructor(extensionContext: ExtensionContext) {
     this.#extensionContext = extensionContext;
   }
 
   async activate(): Promise<void> {
+    await this.startProxy();
```
jeffmaury, line 53: **issue:** should be done only on Windows (and if possible WSL)

jeffmaury (changes requested):

extensions/ollama/src/ollama-extension.ts
bmahabirbu: Agreed thats an easy fix still wip testing now

bmahabirbu: Actually we should have it for all os's no? because the defualt ollama is not using 0.0.0.0 so the container wont be able to bind to it anyways

bmahabirbu:
So the proxy can take care of routing 0.0.0.0 to 127.0.0.1 so we dont need users to set the ollama as 0.0.0.0 so thats a little magic. BUT we still need the container to allow the ability to connect to the windows host ip to 0.0.0.0 so we still need Freds patch

extensions/ollama/src/ollama-extension.ts
jeffmaury: Ok good point

jeffmaury (approved):

jeffmaury:
> So the proxy can take care of routing 0.0.0.0 to 127.0.0.1 so we dont need users to set the ollama as 0.0.0.0 so thats a little magic. BUT we still need the container to allow the ability to connect to the windows host ip to 0.0.0.0 so we still need Freds patch

No I don't think we needs Fred patch but we need to allow all local IPs to be allowed at the network level

benoitf (commented):
I don't think we should any reverse proxy running on the host. There are usually too many related issues

PR #1791, opened 2026-05-09
fix: ask for user consent before update existing workspace file

### LINKED ISSUE #1790: Workspace file may be updated without user consent
### Bug description

if a workspace file is already existing, it may be replace without user getting knowledge

### Operating system

All

### Installation Method

None

### Version

0.0.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

jeffmaury:
consent

Fixes #1790


benoitf:
I was expecting the dialog when selecting a folder telling me that there is an existing file ( as soon as possible)
And then I could just say reuse my workspace configuration and not have to fill all the fields

benoitf (approved):

PR #1807, opened 2026-05-11
fix: kill PTY terminals when stopping a workspace

### LINKED ISSUE #1806: Workspaces: Terminal UI fails to render correctly after stop/start cycle
### Bug description

When a user initially starts a Workspace and navigates to the Terminal tab, the interface renders as expected. However, if the workspace is stopped and then restarted, returning to the Terminal tab results in a rendering glitch where the terminal session appears in a small, restricted window or fails to fill the available container space.

Image

### Operating system

macOS

### Installation Method

### Version

0.2.x

### Steps to reproduce

1. Create a workspace
2. Navigate to the Terminal tab and verify it renders correctly.
3. Stop the workspace from main page.
4. Start the workspace again.
5. Navigate back to the Terminal tab.

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
## Summary
- PTY terminal processes were never killed when stopping a workspace, leaving zombie terminals that could not receive data or end events
- Fixed by tracking workspace ID per terminal and killing all associated terminals in `stop()`
- Fixed race condition where EventStore refresh overwrote local `stopping` state back to `running`, causing the terminal component to remount and reconnect to a dead PTY
- Fixed stale terminal content flashing on workspace restart by only restoring stored content when the terminal session is still alive

Fixes: https://github.com/openkaiden/kaiden/issues/1806

## Test plan
- [ ] Start a workspace, open the terminal tab, type some commands
- [ ] Stop the workspace — terminal should close cleanly (no hanging terminal)
- [ ] Start the workspace again — terminal should open fresh with no flash of old content
- [ ] While workspace is running, navigate away from terminal tab and back — terminal content should be preserved
- [ ] Unit tests pass (`npx vitest run packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts packages/renderer/src/lib/agent-workspaces/AgentWorkspaceTerminal.spec.ts packages/renderer/src/stores/agent-workspaces.svelte.spec.ts`)

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -261,26 +261,8 @@ export class AgentWorkspaceManager implements Disposable {
 
   async stop(id: string): Promise<AgentWorkspaceId> {
     const result = await this.kdnCli.stopWorkspace(id);
-    this.killTerminalsForWorkspace(id);
     this.apiSender.send('agent-workspace-update');
     return result;
-  }
-
-  private killTerminalsForWorkspace(workspaceId: string): void {
-    for (const [onDataId, wsId] of [...this.terminalWorkspaceIds.entries()]) {
-      if (wsId !== workspaceId) continue;
-      const proc = this.terminalProcesses.get(onDataId);
-      if (proc) {
-        try {
-          proc.kill();
-        } catch {
-          /* already exited */
-        }
-      }
-      this.terminalProcesses.delete(onDataId);
-      this.terminalCallbacks.delete(onDataId);
-      this.terminalWorkspaceIds.delete(onDataId);
-    }
   }
 
   shellInAgentWorkspace(
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -57,6 +57,7 @@ export class AgentWorkspaceManager implements Disposable {
     { write: (param: string) => void; resize: (w: number, h: number) => void }
   >();
   private readonly terminalProcesses = new Map<number, IPty>();
+  private readonly terminalWorkspaceIds = new Map<number, string>();
 
   constructor(
     @inject(ApiSenderType)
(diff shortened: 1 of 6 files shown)


serbangeorge-m (approved):

serbangeorge-m:
@bmahabirbu checked on macOS and seems to work ok

gastoner (approved):
Also checked on MacOS, maybe @jeffmaury for Windows as well?

jeffmaury:
I don't experience this on Windows and I do feel it should be handled on the CLI side. Can you clarify where you have this and what is the behavior on CLI ?

gastoner:
FYI https://github.com/openkaiden/kaiden/pull/1801 is merged, maybe rather rebase just in case

bmahabirbu:
@coderabbit resolve

bmahabirbu:
@jeffmaury I believe what was happening was when we stopped a workplace from the ui kdn properly cleaned up its session but the ui kept its terminal state via pyt so when starting it back the ui didnt restart the terminal and loaded the old terminal state

since kdn no longer had that terminal state running it caused the terminal in the ui to freeze 

I experiences the issue described in Windows as the terminal state being frozen after a stop and start

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -260,10 +261,28 @@ export class AgentWorkspaceManager implements Disposable {
 
   async stop(id: string): Promise<AgentWorkspaceId> {
     const result = await this.kdnCli.stopWorkspace(id);
+    this.killTerminalsForWorkspace(id);
     this.apiSender.send('agent-workspace-update');
     return result;
   }
 
+  private killTerminalsForWorkspace(workspaceId: string): void {
```
jeffmaury: **issue:** this part should be removed as kdn terminal process exit when the pod is stopped. And it would not fix the case when the workspace is removed outside kaiden. So if you need to reset a state, it should be when we are detecting that the workspace has been removed (in Details.svelte and we redirect to the workspace page

jeffmaury (changes requested):

bmahabirbu:
@coderabbit resolve

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: updated thanks check the new commits

jeffmaury (approved):

PR #1880, opened 2026-05-14
refactor(kdn): decouple binary download from electron-builder

bmahabirbu:
## Summary
- Move KDN binary downloading out of electron-builder's `beforePack` hook into the extension's own tooling
- Add dedicated `download` and `download:all` scripts to the KDN extension that fetch binaries into `extensions/kdn/assets/<platform>-<arch>/`
- Electron-builder now only declares pre-existing asset files as `extraResources` — no script execution
- Add `"type": "module"` to the KDN extension and rename `build.js` to `build.cjs`
- Update CI workflows to run `pnpm download:kdn:all` before the compile step

Fixes: https://github.com/openkaiden/kaiden/issues/1805

## Test plan
- [x] `pnpm download:kdn` downloads KDN binary for host platform to `extensions/kdn/assets/`
- [ ] `pnpm download:kdn:all` downloads all 5 platform/arch combos
- [ ] `pnpm compile` packages the app with KDN binary in `resources/kdn/`
- [ ] CI workflows pass with the new download step

To test this, do `pnpm download:kdn` to download kdn 
Separately, you can do  pnpm compile to test the packaging 

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1891, opened 2026-05-15
feat(kdn): add version upgrade/downgrade support via CLI Tools page

### LINKED ISSUE #1469: kdn should download the binary automatically
### Bug description

right now it just checks for it in the extension storage missing the download

### Operating system

all OS

### Installation Method

None

### Version

0.0.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
## Summary
- Add a version updater to the CLI Tools page that lets users select and install any available kdn release when kdn is installed via extension/bundled resources
- Filter selectable versions by platform/arch so only compatible binaries are shown
- Use digest-based checksum verification for all downloads (via `getReleaseByTag`)
- No auto-download fallback — kdn must be present via bundled resources, extension storage, or system PATH

Fixes: https://github.com/openkaiden/kaiden/issues/1469

## Test plan
- [ ] CLI Tools page shows kdn with an update option when installed via extension (`installationSource: 'extension'`)
- [ ] Version picker lists available releases filtered to the current platform/arch
- [ ] Selecting a version downloads and installs it, updating the displayed version
- [ ] Externally installed kdn (PATH) does not show the updater
- [ ] When kdn is not found anywhere, activation throws with a clear error message
- [ ] `pnpm -F kdn test` passes

### DIFF
--- a/extensions/kdn/src/kdn-download.ts
+++ b/extensions/kdn/src/kdn-download.ts
@@ -34,7 +34,7 @@ interface ReleaseInfo {
   digests: Map<string, string>;
 }
 
-export async function getLatestRelease(): Promise<ReleaseInfo> {
+export async function getLatestRelease(signal?: AbortSignal): Promise<ReleaseInfo> {
   const headers: Record<string, string> = { Accept: 'application/vnd.github.v3+json' };
   const token = process.env['GITHUB_TOKEN'];
   if (token) {
@@ -43,6 +43,7 @@ export async function getLatestRelease(): Promise<ReleaseInfo> {
   const res = await fetch(`https://api.github.com/repos/${KDN_REPO}/releases/latest`, {
     headers,
     redirect: 'follow',
+    signal,
   });
   if (!res.ok) {
     throw new Error(`failed to fetch latest kdn release: ${res.status} ${res.statusText}`);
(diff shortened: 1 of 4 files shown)


extensions/kdn/package.json
```diff
@@ -4,12 +4,13 @@
...
   "version": "0.2.0-next",
   "icon": "icon.png",
+  "type": "module",
   "publisher": "kaiden",
   "license": "Apache-2.0",
   "engines": {
     "kaiden": "^0.0.1"
   },
-  "main": "./dist/extension.js",
+  "main": "./dist/extension.cjs",
```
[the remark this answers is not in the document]
bmahabirbu: We need type module and extension.js here because Done. Turns out "type": "module" is needed — download.ts uses import.meta.url and top-level await, both ESM-only
  features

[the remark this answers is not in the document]
bmahabirbu: https://github.com/openkaiden/kaiden/pull/1880

this is the pr that does this its also in review

jeffmaury (changes requested):
Looks like this PR also contains #1880

jeffmaury (changes requested):
Extension does not start for me:

Image

extensions/kdn/src/kdn-download.ts
MarsKubeX, line 263: Now that kdn-extension.ts imports this module directly, the guard fires in the Electron app (where VITEST is unset), calling parseArgs(process.argv.slice(2)) which fails because Electron's argv doesn't contain --output, --platform, --arch. I think this is making the extension to fail at start.
Replace the VITEST guard with a proper "am I the main module" check, or (better) split the CLI entrypoint into a separate file (e.g. kdn-download-cli.ts) and keep kdn-download.ts as a pure library with no top-level side effects.

extensions/kdn/src/kdn-extension.ts
```diff
@@ -14,14 +14,17 @@
  * limitations under the License.
  *
  * SPDX-License-Identifier: Apache-2.0
- ***********************************************************************/
+ **********************************************************************/
```
MarsKubeX, line 17: Nit: keep the same number of '*' to avoid unnecessary noise in the PR.

extensions/kdn/src/kdn-download.ts
```diff
@@ -58,11 +59,84 @@ export async function getLatestRelease(): Promise<ReleaseInfo> {
...
+    headers['Authorization'] = `Bearer ${token}`;
+  }
+  const res = await fetch(`https://api.github.com/repos/${KDN_REPO}/releases/tags/v${version}`, {
+    headers,
+    redirect: 'follow',
+    signal,
+  });
+  if (!res.ok) {
+    throw new Error(`failed to fetch kdn release v${version}: ${res.status} ${res.statusText}`);
+  }
```
MarsKubeX, line 75: I see that this code is the same from ``getLatestRelease``. Could it be in a helper to avoid duplication?

extensions/kdn/src/kdn-download.ts
bmahabirbu: https://github.com/openkaiden/kaiden/pull/1880

so this will fix that issue! So ill put this in draft until my other pr goes in then i can continue

bmahabirbu:
closing no longer working on kdn

PR #1905, opened 2026-05-18
feat(claude): add custom baseURL support for Anthropic provider

### LINKED ISSUE #1855: Feature Request: Support Custom baseURL for Anthropic/Claude Provider

Description:
Currently, the Anthropic provider in Kaiden only allows for the configuration of the ANTHROPIC_API_KEY. While the OpenAI provider supports a baseURL field, the Anthropic provider is hardcoded to the default Anthropic API endpoints.

This request is to add a baseURL configuration option for the Anthropic provider to allow compatibility with local inference engines and proxy services.

Use Case:
Tools like llama.cpp (llama-server) now support Anthropic-compatible API endpoints alongside OpenAI and Ollama. Allowing a custom baseURL for the Claude provider enables Kaiden to interface with locally-hosted models (via llama-server) or alternative API gateways while maintaining the specific request/response formatting required for Claude-style interactions.

Proposed Changes:

UI/Configuration: Update the "Create Claude" configuration screen to include an optional baseURL field, mirroring the "Create OpenAI" implementation.

Provider Logic: Pass the custom baseURL to the Anthropic client initialization. If the field is blank, it should default to https://api.anthropic.com.

Example Workflow:

Provider: Claude

baseURL: http://localhost:8080 (or relevant local endpoint)

apiKey: sk-no-key-required (for local servers) or valid Anthropic key (for proxies)

Technical Implementation Context

When implementing this, ensure the client initialization handles the base URL correctly. If Kaiden uses the official Anthropic SDK, the implementation typically follows this pattern:

TypeScript
const anthropic = new Anthropic({
  apiKey: 'my_api_key',
  baseURL: 'http://localhost:8080', // This should be the new configurable field
});
For llama-server specifically, using the Claude-compatible endpoint often provides better results for models fine-tuned on Anthropic's prompt formats compared to forcing them through an OpenAI-style shim.

Using the llama-server's /v1 endpoint, but Claude Code agent workspace does not see the OpenAI model provider.

### Additional context

_No response_

MarsKubeX:
Allow users to configure a custom base URL when creating a Claude provider connection, enabling local inference engines (e.g. llama-server) and API proxies. Connections with a custom URL are registered as self-hosted with the endpoint exposed in the model ID.

Captura de pantalla 2026-05-18 a las 13 30 47

Not merge until https://github.com/openkaiden/kdn/pull/543 is merged.
Close #1855

### DIFF
--- a/extensions/claude/package.json
+++ b/extensions/claude/package.json
@@ -20,6 +20,11 @@
           "scope": "InferenceProviderConnectionFactory",
           "markdownDescription": "Provide a Claude API key\n\n:button[Grab a key]{href=\"https://console.anthropic.com/settings/keys\"}"
         },
+        "claude.factory.baseURL": {
+          "type": "string",
+          "scope": "InferenceProviderConnectionFactory",
+          "description": "Custom base URL (leave empty for https://api.anthropic.com)"
+        },
         "claude.factory.apiKey": {
           "type": "string",
           "scope": "InferenceProviderConnectionFactory",
--- a/extensions/claude/src/manager/claude-inference-manager.spec.ts
+++ b/extensions/claude/src/manager/claude-inference-manager.spec.ts
@@ -25,7 +25,7 @@ import { assert, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { ClaudeProviderSymbol, SecretStorageSymbol } from '/@/inject/symbol';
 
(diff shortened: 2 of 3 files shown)


jeffmaury (changes requested):
Blocking for now as we need to decide if we should be backward compatible

MarsKubeX:
closed in favor of #2094

PR #1949, opened 2026-05-20
fix: redact secret values from kdn CLI error logs

### LINKED ISSUE #1944: 'kdn secret create' fails on duplicate entry and exposes unmasked secret value in console logs
### Bug description

When creating a new workspace using a configuration name or identifier that already exists, the application attempts to regenerate the secret via kdn secret create. This causes a terminal execution failure (secret already exists). Furthermore, the console error logs the entire command string in plain text, exposing the sensitive --value (API key/secret) without masking it.

✖ main ↪ kdn failed: /Applications/Kaiden.app/Contents/Resources/kdn/kdn secret create test-vertex-ai-openai --type other --value `{secretValue}` --host api.openai.com --header Authorization --headerTemplate Bearer ${value} --output json - failed to create secret: secret "test-vertex-ai-openai": secret already exists

### Operating system

macOS

### Installation Method

None

### Version

0.2.x

### Steps to reproduce

1.Create a workspace with a specific configuration (e.g., name it test-project-openai).
2.Attempt to create a second workspace using the exact same configuration name.
3.Open the console logs and observe the red error output.

### Relevant log output

```shell

```

### Additional context

_No response_

serbangeorge-m:
## Summary

- Redacts `--value` (API key/secret) from `console.error` output when `kdn secret create` fails, replacing it with `***`
- Handles the "secret already exists" case silently in `KdnCli.createSecret` — returns `{ name }` instead of logging an error and forcing callers to catch-and-ignore
- Simplifies `AgentWorkspaceManager.ensureModelSecret` by removing its now-unnecessary try/catch for duplicate secrets

## Test plan

- [x] New tests: `returns secret name silently when secret already exists`, `does not log error when secret already exists`, `rethrows non-"already exists" errors`, `redacts secret value from error logs on non-"already exists" failure`
- [x] All 72 `kdn-cli.spec.ts` tests pass
- [x] All 89 `agent-workspace-manager.spec.ts` tests pass
- [x] All 16 `secret-manager.spec.ts` tests pass
- [x] Typecheck passes

Closes #1944

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -640,13 +640,13 @@ describe('ensureModelSecret', () => {
     expect(secretManager.create).not.toHaveBeenCalled();
   });
 
-  test('tolerates "already exists" error from secret creation', async () => {
+  test('proceeds normally when secret already exists (handled by KdnCli)', async () => {
     vi.mocked(providerRegistry.getInferenceConnectionCredentials).mockReturnValue({
       credentials: { 'claude:tokens': 'sk-ant-secret' },
       llmMetadataName: 'anthropic',
       endpoint: undefined,
     });
-    vi.mocked(secretManager.create).mockRejectedValue(new Error('secret already exists: my-workspace-anthropic'));
+    vi.mocked(secretManager.create).mockResolvedValue({ name: 'my-workspace-anthropic' });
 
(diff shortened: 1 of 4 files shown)


fbricon (approved):
I would probably make redactSensitiveArgs more generic and pass the sensitive flags as args, and have a redactSecretArgs function wrap it for ['--value'], but that's unnecessary for now as it's only used for secrets, so maybe overkill for now. We can always refactor it if/when needed.

PR #1950, opened 2026-05-20
feat(openshell): add extension to download and bundle OpenShell CLI

### LINKED ISSUE #1924: Create an OpenShell extension

We need user to be able to locate the OpenShell executable

Similar to the kdn extension should locate the OpenShell executable the same way

_No response_

### Additional context

_No response_

bmahabirbu:
## Summary
- Adds a new `openshell` extension that downloads the OpenShell CLI binary from `NVIDIA/OpenShell` GitHub releases at build time and bundles it into the Electron app via `extraResources`
- Follows the same pipeline pattern as the `kdn` extension: download during build phase, verify SHA256 via GitHub digest API, package into `.cdix`, and include in electron-builder `beforePack`
- Supports `darwin-arm64`, `linux-x64`, and `linux-arm64` targets (no Windows binary available upstream — Windows builds skip gracefully)

Fixes: https://github.com/openkaiden/kaiden/issues/1924

## Test plan
- [ ] Run `pnpm --filter openshell download:all` and verify all 3 platform binaries download with checksum verification
- [ ] Run `pnpm run build:extensions:openshell` and verify the extension builds and produces `.cdix`
- [ ] Verify CI builds pass on macOS and Linux runners
- [ ] Verify Windows CI build skips openshell packaging without error


benoitf (commented):
note: we should have somewhere the version of OpenShell to use. To have a build that can be reproduced over time (avoid the usage of latest)

extensions/openshell/scripts/download.ts
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ **********************************************************************/
+
+import { resolve, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseArgs } from 'node:util';
+
+import { getLatestRelease, downloadOpenshell } from '../src/openshell-download';
```
benoitf: I don't think we can import relatively like that from scripts to src folder

extensions/openshell/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "openshell",
+  "displayName": "openshell",
```
benoitf: it's OpenShell, not openshell

extensions/openshell/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "publisher": "kaiden",
+  "license": "Apache-2.0",
+  "engines": {
+    "kaiden": "^0.0.1"
+  },
+  "main": "./dist/extension.js",
+  "source": "./src/extension.ts",
+  "contributes": {
+    "configuration": {
+      "title": "openshell",
```
benoitf: OpenShell

extensions/openshell/package.json
```diff
@@ -0,0 +1,47 @@
...
+      "properties": {
+        "openshell.binary.path": {
+          "type": "string",
+          "format": "file",
+          "scope": "DEFAULT",
+          "default": "",
+          "description": "Custom path to openshell binary (Default is blank)"
```
benoitf: OpenShell

benoitf (commented):
icon for OpenShell looks not expected (I see Kaiden image)

.electron-builder.config.cjs
```diff
@@ -174,6 +174,30 @@ const config = {
     // download & bundle kdn CLI binary
     await downloadKdn(context);
 
+    // include pre-downloaded openshell CLI binary (non-Windows only)
+    if (context.electronPlatformName !== 'win32') {
```
benoitf: hello, why do we have such complex things there ?

if we include files, if it's not there on Windows, it'll just be excluded (not included)

if I look at 
https://github.com/podman-desktop/podman-desktop/blob/main/.electron-builder.config.cjs#L159-L165

it looks way more simpler

why we don't download files at the right place rather than having many x64/arm64 includes ?

.electron-builder.config.cjs
```diff
@@ -174,6 +174,25 @@ const config = {
     // download & bundle kdn CLI binary
     await downloadKdn(context);
 
+    // include pre-downloaded OpenShell CLI binary
+    const openshellArchMap = { [Arch.x64]: 'x64', [Arch.arm64]: 'arm64' };
+    const openshellArch = openshellArchMap[context.arch];
+    if (openshellArch) {
```
benoitf: I'm not sure to follow why we would need this if condition

it's used to append a suffix, so couldn't we just add the suffix ?



and as noted before, while do we need to have an existsync condition as it would not include things if it's not existing anyway

jeffmaury: Also bunding into executable is not part of the original issue

jeffmaury (changes requested):
Remove the  bundling part

.gitignore
```diff
@@ -14,6 +14,7 @@ dist
 /storybook/storybook-static/
 yarn.lock
 extensions/podman/assets/
+extensions/openshell/assets/
```
jeffmaury: **issue:** unrelated to this PR

extensions/openshell/types/template.d.ts
jeffmaury, line 1: **praise:** Why this change ?

jeffmaury (approved):

benoitf (commented):
I was wondering why this extension is not using inversify ?

bmahabirbu:
@coderabbit resolve

benoitf (approved):

PR #1972, opened 2026-05-25
fix: stream lifecycle cleanup and Ollama abort handling

### LINKED ISSUE #1971: Ollama memory not released after workspace deletion or chat navigation
### Bug description

When using Ollama as an inference provider, models stay loaded in RAM after a workspace is stopped or removed, and chat streams are never cancelled when navigating away. This causes Ollama memory to grow continuously.

Two issues:

- Stopping or deleting a workspace does not tell Ollama to unload the model. The container is removed but Ollama keeps the model in memory.
- Navigating away from an active chat stream does not cancel the request. Ollama continues generating tokens for abandoned streams, keeping the model active and preventing it from unloading.

### Operating system

macOS

### Installation Method

None

### Version

0.2.x

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

serbangeorge-m:
## Description

Streaming resources are not properly released on abort or error, and the Ollama extension does not cancel in-flight requests on deactivation.

This PR fixes stream lifecycle cleanup across the chat pipeline:
- **ChatManager**: cancel the stream reader on error, propagate abort signal to background title generation
- **IPC transport**: add `cancel()` callbacks to ReadableStreams so consumer cancellation stops the underlying IPC stream
- **Preload**: tighten stream state matching by onDataId, release buffered chunks early on completion/error, add TTL cleanup on error path
- **Ollama extension**: add AbortController to cancel in-flight fetches on deactivation, clean up connection disposable on deactivate
- Remove leftover `console.log` calls from chat transport

Closes https://github.com/openkaiden/kaiden/issues/1971

## Steps to test

1. Open chat, send a prompt, abort mid-stream — app should remain responsive with no console errors
2. Send another prompt and let it complete — response should render correctly
3. Navigate away from chat and back while a stream is in progress — reconnection should replay buffered chunks
4. Disable/re-enable the Ollama extension — no lingering fetch errors should appear in the console


MarsKubeX (approved):
LGTM

benoitf (dismissed):
I'm not sure to understand why we need checks for deactivated ?
Either an eventlistener is not removed but I don't see why we should have check for deactivated. As if deactivated it should mean the code should never be invoked again

serbangeorge-m:
> I'm not sure to understand why we need checks for deactivated ? Either an eventlistener is not removed but I don't see why we should have check for deactivated. As if deactivated it should mean the code should never be invoked again

This was flagged by @coderabbitai - the ones marked as resolved.

jeffmaury (changes requested):
Looks like a bad fix for a real problem: it is linking core to a specific extension.

I think that should be done at the E2E level rather than here or maybe at the ollama extension through a keep alive parameter

jeffmaury (changes requested):
This PR is linking workspaces and ollama model management that is invalid to me. If a user has loaded an ollama model and use it against a workspace when this one is stopped it will unload it which is wrong to me

So I have the feeling your are trying to fix an E2E issue so it should be fixed here where you can control the ollama models lifecycle

bmahabirbu (commented):
Agreed that this should be a e2e specific fix for managing ollama lifecycle

the issue is we dont want to be managing the lifecycle of ollama models for users if they expect the model to still be working after a workspace is closed

https://docs.ollama.com/faq#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately

Maybe we could clear the cache for the model if a workspace gets deleted to preserve the models usable state that could be an option to investigate

serbangeorge-m:
> This PR is linking workspaces and ollama model management that is invalid to me. If a user has loaded an ollama model and use it against a workspace when this one is stopped it will unload it which is wrong to me
> 
> So I have the feeling your are trying to fix an E2E issue so it should be fixed here where you can control the ollama models lifecycle

@jeffmaury I have reverted the workspace related changes, the remaining changes are scoped to stream lifecycle cleanup only.

jeffmaury (approved):

PR #1982, opened 2026-05-27
feat: switch knowledges UI to new extension point

### LINKED ISSUE #1976: Update knowledges UI to use the new extension point

Use now the new type of resouces					

Required as Docling extension now implement this new extension point

_No response_

### Additional context

_No response_

jeffmaury:
To test:
- remove existing knowleges (as the format for the storage is not backward compatible)
- create a milvus resource
- create a docling resource
- create a knowledge using those resource
- index documents into that knowledge
- remove the knowledge

new-knowledge

Fixes #1976


packages/main/src/plugin/rag-environment-registry.ts
```diff
@@ -77,15 +74,15 @@ export class RagEnvironmentRegistry {
         _listener,
         name: string,
         ragConnection: { name: string; providerId: string },
-        chunkerId: string,
+        chunkerConnection: { id: string; providerId: string },
```
gastoner, line 77: Cant this be `{ name: string; providerId: string }` similarly as `ragConnection` has? => we can create an interface for this then also etc

gastoner (commented):
Left one nit, 
works for me with .md and .pdf file, but the .txt or .log files that can be selected. 

Also would be nice to add a scrolling mechanism for the uploaded files

packages/main/src/plugin/rag-environment-registry.ts
jeffmaury: > Cant this be `{ name: string; providerId: string }` similarly as `ragConnection` has? => we can create an interface for this then also etc

No because we added an id for those new connections in order to add the identification semantic and the name is more for display purpose. We probably should add id to ragConnection (as we are doing also for inference connections)

jeffmaury:
> Left one nit, works for me with .md and .pdf file, but the .txt or .log files that can be selected.
> 
> Also would be nice to add a scrolling mechanism for the uploaded files

Not related to this PR so probably will open a follow up issue

bmahabirbu:
Ya I think simple we can remove the supported (txt etc) message for users since it doesnt work at the moment so not related to this pr

bmahabirbu (approved):

PR #1985, opened 2026-05-28
fix: set HOME env var for child processes on Windows

bmahabirbu:
## Summary
- On Windows, `$HOME` is not a standard environment variable (`USERPROFILE` is used instead). The `kdn` CLI relies on `$HOME` to expand mount paths in `workspace.json` (e.g. when "Home Directory" file access is selected), causing workspace creation to fail with `crun: executable file not found in $PATH` errors.
- Ensures `HOME` is set to `os.homedir()` in the child process environment on Windows, matching the existing pattern for `PATH` adjustments.

Fixes: https://github.com/openkaiden/kdn/issues/528

## Test plan
- [x] Existing `exec.spec.ts` tests pass (23/23)
- [x] Existing `kdn-cli.spec.ts` tests pass (72/72)
- [x] New test verifies `HOME` is injected on Windows when not already set
- [ ] Manual: On Windows, create a workspace with "Home Directory" file access and verify the terminal launches correctly

### DIFF
--- a/packages/main/src/plugin/util/exec.spec.ts
+++ b/packages/main/src/plugin/util/exec.spec.ts
@@ -559,6 +559,40 @@ describe('exec', () => {
     expect(stdout).toContain('Hello, World!');
     expect(setEncodingMock).toBeCalledWith('utf16le');
   });
+
+  test('should set HOME to homedir on Windows when HOME is not set', async () => {
+    const command = 'echo';
+    const args = ['test'];
+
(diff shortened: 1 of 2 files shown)


jeffmaury (changes requested):
Should be done upstream

bmahabirbu:
See https://github.com/podman-desktop/podman-desktop/pull/17655

PR #1986, opened 2026-05-28
fix(ci): remove playwright postinstall and fix e2e browser setup

serbangeorge-m:
## Summary
- Remove `postinstall: playwright install chromium` from `tests/playwright/package.json` — stops all non-e2e jobs (typecheck, lint, unit tests, builds) from downloading a browser they never use
- For e2e jobs:
  - Bump `pnpm/action-setup` v4.1.0 → v6.0.8 (Node.js 20 → 24) to fix `playwright install` hanging on some runners
  - Cache Playwright browsers (`~/.cache/ms-playwright`) so downloads are skipped on cache hit
  - Install Chromium as a dedicated step, only on cache miss

## Problem
`pnpm install` was hanging for 40+ minutes on some CI jobs because the `postinstall` hook triggered `playwright install chromium` on **every** job. The Chromium zip extraction would stall on runners with I/O contention (noisy neighbor), causing non-deterministic cancellations. The old `pnpm/action-setup` v4.1.0 running on deprecated Node.js 20 exacerbated the issue.

## Test plan
- [ ] Verify `typecheck` and `linter` jobs complete without hanging at `Execute pnpm`
- [ ] Verify `smoke-e2e-tests` install Chromium and run successfully
- [ ] Verify Playwright browser cache is populated on first run and reused on subsequent runs

### DIFF
--- a/.github/workflows/e2e-test-job.yaml
+++ b/.github/workflows/e2e-test-job.yaml
@@ -95,6 +95,32 @@ jobs:
       - name: Execute pnpm
         run: pnpm install
 
+      - name: Get Playwright version and cache path
+        id: playwright-info
+        shell: bash
+        run: |
+          echo "version=$(npx --yes playwright --version | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')" >> "$GITHUB_OUTPUT"
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            echo "path=$USERPROFILE/AppData/Local/ms-playwright" >> "$GITHUB_OUTPUT"
+          elif [[ "$RUNNER_OS" == "macOS" ]]; then
+            echo "path=$HOME/Library/Caches/ms-playwright" >> "$GITHUB_OUTPUT"
+          else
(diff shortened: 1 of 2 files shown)


benoitf (changes requested):
if done, should be turned on on dependabot side

serbangeorge-m:
> if done, should be turned on on dependabot side

everything related to CI action versions?

benoitf:
yes, for actions update

jeffmaury (commented):
Wondering if this is aligned with upstream ? Shouldn't we cherry pick then ?

serbangeorge-m:
> Wondering if this is aligned with upstream ? Shouldn't we cherry pick then ?

The e2e workflow is different — kaiden uses a reusable e2e-test-job.yaml, nothing to cherry-pick I think. For the action version bumps, I'll create a separate PR to enable dependabot for github-actions so they stay in sync automatically.

benoitf (approved):
please update title/body before merging

jeffmaury (approved):

PR #2020, opened 2026-06-01
feat(openshell): add OpenshellCli wrapper for sandbox and gateway operations

### LINKED ISSUE #1923: Create a backend openshell-cli component similar to kdn-cli

First step to adopt OpenShell

In the first iteration implement the same interface 

_No response_

### Additional context

_No response_

gastoner:
Introduces a low-level CLI wrapper that maps Kaiden sandbox/gateway operations to openshell commands (create, list, start, stop, delete, connect) and registers it as a singleton in the plugin DI container. 

Closes https://github.com/openkaiden/kaiden/issues/1923

### DIFF
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -1,0 +1,74 @@
+/**********************************************************************
+ * Copyright (C) 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import z from 'zod';
+
+export const GatewayInfoSchema = z.object({
+  name: z.string(),
+  endpoint: z.string(),
+  active: z.boolean().optional(),
+  auth: z.string().optional(),
+  type: z.string().optional(),
+});
+
+export type GatewayInfo = z.output<typeof GatewayInfoSchema>;
+
+export const SandboxInfoSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  phase: z.string(),
+  created_at: z.string().optional(),
+  current_policy_version: z.number().optional(),
+  labels: z.record(z.string(), z.string()).optional(),
+  resource_version: z.number().optional(),
+});
+
+export type SandboxInfo = z.output<typeof SandboxInfoSchema>;
+
+export interface CreateSandboxOptions {
+  name?: string;
+  gateway?: string;
+  from?: string;
+  gpu?: boolean;
+  gpuDevice?: string;
+  cpu?: string;
+  memory?: string;
+  providers?: string[];
+  labels?: Record<string, string>;
+  command?: string[];
+}
+
+export interface GatewayAddOptions {
+  endpoint: string;
+  name?: string;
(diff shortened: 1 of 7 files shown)


benoitf (commented):
I think we're missing the part that manage the lifecycle of the gateway (and how to start the gateway)

I guess it's a separate file as it's not for the CLI but without the gateway part, it won't be possible to start /list etc stuff

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * The gateway is an external service — this class does not start or stop it.
+ * Instead it wraps the client-side CLI commands that register, select, and
+ * query gateways so that sandbox operations have a configured endpoint.
+ *
+ * Commands:
+ *   - `openshell gateway add <endpoint>`
+ *   - `openshell gateway remove [name]`
+ *   - `openshell gateway select [name]`
+ *   - `openshell gateway list`
```
benoitf: it's a separate binary

it's named openshell-gateway

https://github.com/NVIDIA/OpenShell/releases/download/v0.0.52/openshell-gateway-aarch64-apple-darwin.tar.gz


and for Windows, it should be run through a podman/docker container

gastoner: Do we want to add new config for specifying a custom binary path for openshell-gateway too?

benoitf: I think that yes it would be interesting as well.

Kaiden should come out of the box (at least on Linux and macOS) with a binary
for Windows for now it'll require a docker/podman engine to start the gateway

benoitf: all the gateway operation of the CLI need to go to openshell-cli.ts (the registration of gateway, etc)

while here (this file) it should be about starting a new gateway (or stopping it)

bmahabirbu: So we will need to bundle the cli + gateway + sandbox binaries then!

jeffmaury (commented):
After discussion with @benoitf, I don't think we need to handle gateways in the extension.

On the backend component, we need to:
- discover existing gateways
- for each gateway list the sandboxes
- if we have found no gateways, we should start one

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -0,0 +1,258 @@
...
+    } catch (err: unknown) {
+      const detail = this.extractCliError(err);
+      console.error(`openshell failed: ${cliPath} --version — ${detail}`);
+      throw new Error(detail);
+    }
+  }
+
+  // ── sandbox commands ──────────────────────────────────────────────
+
+  async createSandbox(options: { name?: string; agent: string }): Promise<void> {
```
benoitf: I think we need a CreateParams interface

as we won't use only the agent

we need to provide some providers, and other stuff

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -0,0 +1,258 @@
...
+    const args = ['sandbox', 'create'];
+    if (options.name) {
+      args.push('--name', options.name);
+    }
+    args.push('--', options.agent);
+    await this.runCli(args);
+  }
+
+  async listSandboxes(): Promise<unknown> {
```
benoitf: I would expect it's providing something not unknown but with a valid data

use zod to validate the json + having the interface we have and returning the data

packages/api/src/openshell-gateway-info.ts
```diff
@@ -0,0 +1,46 @@
...
+  [key: string]: unknown;
+}
+
+export interface OpenshellGatewayStartOptions {
+  port?: number;
+  bindAddress?: string;
+  disableTls?: boolean;
+}
+
+export interface GatewaySandboxes {
```
benoitf, line 71: please use zod as we will need to ensure the JSON is valid as well (as some versions may not be compliant with what we expect)

benoitf (approved):

gastoner:
Rebased

jeffmaury (approved):

PR #2022, opened 2026-06-01
refactor(renderer): extract reusable Card component with gradient background

fbricon:
Centralize card styling (background, border, shadow, gradient) into a single Card component so all surfaces share the same visual treatment and dark/light theming is defined once.

Makes the Agent workspaces page look closer to the mockups:

Screenshot 2026-05-29 at 12 31 10
Screenshot 2026-05-29 at 12 31 34

### DIFF
--- a/packages/renderer/src/lib/components/Card.svelte
+++ b/packages/renderer/src/lib/components/Card.svelte
@@ -1,0 +1,15 @@
+<script lang="ts">
+import type { Snippet } from 'svelte';
+import type { HTMLAttributes } from 'svelte/elements';
+
+interface Props extends HTMLAttributes<HTMLDivElement> {
+  class?: string;
+  children: Snippet;
+}
+
+let { class: className, children, ...restProps }: Props = $props();
+</script>
+
+<div class="content-card {className}" {...restProps}>
+  {@render children()}
+</div>
--- a/packages/renderer/src/lib/models/ProviderConnectionTiles.svelte
+++ b/packages/renderer/src/lib/models/ProviderConnectionTiles.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 import { router } from 'tinro';
 
+import Card from '/@/lib/components/Card.svelte';
 import type { InferenceConnectionSummary } from '/@/lib/models/models-utils';
 
 interface Props {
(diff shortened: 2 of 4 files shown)


packages/renderer/src/lib/components/Card.svelte
```diff
@@ -0,0 +1,31 @@
...
+    box-shadow:
+      0 1px 2px rgba(13, 27, 46, 0.06),
+      0 8px 24px rgba(13, 27, 46, 0.08);
+  }
+
+  :global(.dark .content-card) {
+    background: linear-gradient(160deg, var(--pd-content-card-bg) 0%, var(--pd-content-bg) 100%);
+    box-shadow: none;
+  }
+</style>
```
gastoner: Can we avoid using styles?

packages/renderer/src/lib/models/ProviderConnectionTiles.svelte
```diff
@@ -102,19 +102,18 @@ function getSecondBadge(status: string): BadgeInfo | undefined {
...
-      0 8px 24px color-mix(in srgb, var(--pd-content-bg) 7%, transparent);
+      0 8px 24px color-mix(in srgb, var(--pd-content-bg) 7%, transparent) !important;
   }
 
-  :global(.dark) .provider-tile {
-    background: linear-gradient(145deg, var(--pd-content-card-bg) 0%, var(--pd-content-bg) 100%);
-    box-shadow: none;
+  :global(.dark .provider-tile) {
+    box-shadow: none !important;
   }
```
gastoner: IMHO I dont think that the `important` is right here (the style section should be also converted to class but thats a different issue)

jeffmaury (approved):
LGTM.

Just wondering if styles shouldn't be declared centrally ?

packages/renderer/src/lib/models/ProviderConnectionTiles.svelte
fbricon: fixed

packages/renderer/src/lib/components/Card.svelte
fbricon: styles converted to a central class

fbricon:
> LGTM.
> 
> Just wondering if styles shouldn't be declared centrally ?

done

PR #2024, opened 2026-06-01
feat(copilot): add copilot agent extension

### LINKED ISSUE #1919: Create copilot (github ?) extension
### Task content

So that it registers Copilot agent

fbricon:
Register a GitHub Copilot agent extension with build scripts, unit tests,
and integration into the monorepo build pipeline. Includes dual-theme icons
(dark/light variants) and full test/typecheck coverage.

Screenshot 2026-06-01 at 15 57 53
Screenshot 2026-06-01 at 15 58 20

Note the Agent Workspace wizard does NOT show the copilot agent, as it still uses a list of hardcoded agents, instead of discovering them dynamically, I believe that's out-of-scope of this feature, will need to be handled in a subsequent PR.

Fixes #1919


extensions/copilot/src/extension.ts
```diff
@@ -0,0 +1,50 @@
...
+        light: './icon_light.png',
+        dark: './icon_dark.png',
+      },
+      logo: {
+        dark: './icon_dark.png',
+        light: './icon_light.png',
+      },
+    },
+    tags: [],
+    isSupportedRuntime(): boolean {
```
jeffmaury: **praise:** should be removed as absence is similar and it's likely to be removed from the API

jeffmaury (approved):

MarsKubeX (approved):
LGTM

PR #2025, opened 2026-06-01
feat(vertex-ai): persist connection IDs in secret storage

### LINKED ISSUE #1942: Review VertexAI extension storage format
### Task content

See #1917 for the new format

MarsKubeX:
Replace the transient counter-based IDs for vertex extension with stable UUIDs stored alongside connection configs in the existing JSON array. Legacy entries without an id field are auto-migrated on first read.

Closes #1942

### DIFF
--- a/extensions/vertex-ai/src/vertex-ai.spec.ts
+++ b/extensions/vertex-ai/src/vertex-ai.spec.ts
@@ -33,13 +33,7 @@ import {
   type VertexAiConnectionConfig,
 } from './vertex-ai';
 
-vi.mock(import('node:crypto'), async importOriginal => {
-  const actual = await importOriginal();
-  return {
-    ...actual,
-    randomUUID: vi.fn().mockReturnValue('fake-uuid-1'),
-  };
-});
+vi.mock(import('node:crypto'));
 
 vi.mock(import('node:fs/promises'));
 vi.mock(import('node:os'));
@@ -501,6 +495,16 @@ describe('factory', () => {
       'vertex-ai.factory.region': 'us-east5',
       'vertex-ai.factory.credentialsFile': '/home/user/.config/gcloud/application_default_credentials.json',
     });
+
+    const stored: StoredConnection[] = [
+      {
+        id: 'fake-uuid-1',
+        projectId: 'my-project',
+        region: 'us-east5',
+        credentialsFile: '/home/user/.config/gcloud/application_default_credentials.json',
+      },
+    ];
+    vi.mocked(SECRET_STORAGE_MOCK.get).mockResolvedValue(JSON.stringify(stored));
 
     await expect(
       create({
--- a/extensions/vertex-ai/src/vertex-ai.spec.ts
+++ b/extensions/vertex-ai/src/vertex-ai.spec.ts
@@ -16,6 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import { randomUUID } from 'node:crypto';
 import { access, readFile } from 'node:fs/promises';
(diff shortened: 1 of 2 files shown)


extensions/vertex-ai/src/vertex-ai.ts
```diff
@@ -142,22 +145,28 @@ export class VertexAi implements Disposable {
...
   }
 
-  private async removeConnectionConfig(config: VertexAiConnectionConfig): Promise<void> {
-    const configs = await this.getConnectionConfigs();
+  private async removeConnection(config: VertexAiConnectionConfig): Promise<void> {
+    const stored = await this.getStoredConnections();
     const key = this.getConfigHash(config);
-    const filtered = configs.filter(c => this.getConfigHash(c) !== key);
+    const filtered = stored.filter(c => this.getConfigHash(c) !== key);
     await this.secrets.store(CONNECTIONS_KEY, JSON.stringify(filtered));
```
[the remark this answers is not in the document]
MarsKubeX: Dismissing as out of scope. The race scenario can't occur — factory() has an explicit duplicate guard at line 423 that rejects a second call with the same config before any save happens, and registerInferenceProviderConnection has its own duplicate check at line 290. The old code (removeConnectionConfig) also removed by config hash — this behavior is unchanged. Switching to ID-based removal is a design improvement beyond the scope of #1942.

bmahabirbu (approved):

extensions/vertex-ai/src/vertex-ai.spec.ts
```diff
@@ -24,7 +25,21 @@ import { createVertexAnthropic } from '@ai-sdk/google-vertex/anthropic';
...
+  type VertexAiConnectionConfig,
+} from './vertex-ai';
+
+vi.mock(import('node:crypto'), async importOriginal => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    randomUUID: vi.fn().mockReturnValue('fake-uuid-1'),
+  };
+});
```
gastoner: Cant we do just `vi.mock(import('node:crypto'));`?

gastoner (approved):
Codewise LGTM, left one nit

extensions/vertex-ai/src/vertex-ai.ts
```diff
@@ -142,22 +145,28 @@ export class VertexAi implements Disposable {
...
-    await this.secrets.store(CONNECTIONS_KEY, JSON.stringify(configs));
+  private async saveConnection(connection: StoredConnection): Promise<void> {
+    const stored = await this.getStoredConnections();
+    stored.push(connection);
+    await this.secrets.store(CONNECTIONS_KEY, JSON.stringify(stored));
   }
 
-  private async removeConnectionConfig(config: VertexAiConnectionConfig): Promise<void> {
-    const configs = await this.getConnectionConfigs();
+  private async removeConnection(config: VertexAiConnectionConfig): Promise<void> {
```
jeffmaury: **question:** we should use the connection id to remove here and also in the connections array

jeffmaury (changes requested):

extensions/vertex-ai/src/vertex-ai.spec.ts
MarsKubeX: I think no since it also use createHash from node:crypto for config hashing and need the real implementation. The factory preserves createHash while only overriding randomUUID.

extensions/vertex-ai/src/vertex-ai.ts
MarsKubeX: Should be fixed

jeffmaury (approved):

PR #2040, opened 2026-06-02
fix(claude): use connection id for removal and map keying

MarsKubeX:
RemoveConnection now filters by persisted id, and the connections Map is keyed by id with a separate Set for duplicate token detection.

### DIFF
--- a/extensions/claude/src/manager/claude-inference-manager.ts
+++ b/extensions/claude/src/manager/claude-inference-manager.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { createHash, randomUUID } from 'node:crypto';
+import { randomUUID } from 'node:crypto';
 
 import { createAnthropic } from '@ai-sdk/anthropic';
 import AnthropicClient from '@anthropic-ai/sdk';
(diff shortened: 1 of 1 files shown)


extensions/claude/src/manager/claude-inference-manager.ts
```diff
@@ -88,17 +89,17 @@ export class ClaudeInferenceManager {
...
+    const filtered = stored.filter(entry => entry.id !== id);
     await this.secrets.store(TOKENS_KEY, JSON.stringify(filtered));
   }
 
   private async registerInferenceProviderConnection({ id, token }: { id: string; token: string }): Promise<void> {
     const key = this.maskKey(token);
     const tokenHash = this.getTokenHash(token);
 
-    if (this.connections.has(tokenHash)) {
+    if (this.activeTokenHashes.has(tokenHash)) {
```
jeffmaury: **suggestion:** Not needed just change connections to be a map keyed by id

jeffmaury (changes requested):

extensions/claude/src/manager/claude-inference-manager.ts
MarsKubeX: Should be updated.

gastoner (approved):
Codewise LGTM

jeffmaury (approved):

PR #2057, opened 2026-06-03
feat(projects): add create project wizard with local and git source flows

### LINKED ISSUE #1928: Add a new create project workflow

Users should be able to create project

Create a workflow/page to allow project creation.

From the mockup:

Image

Image

_No response_

### Additional context

_No response_

MarsKubeX:
Introduces a two-step wizard (Source → Review) that supports creating projects from either a local directory or a Git repository URL, with path validation, tilde expansion, and inline error display.

https://github.com/user-attachments/assets/6be79577-6ad7-401d-a030-2aeccbb40a59

https://github.com/user-attachments/assets/a87da055-afb3-4a60-8c6c-f1c36e8f7e38

Closes #1928

### DIFF
--- a/packages/api/src/workspace-project-info.ts
+++ b/packages/api/src/workspace-project-info.ts
@@ -26,6 +26,7 @@ export interface FilesystemConfiguration {
 export interface WorkspaceProjectInfo {
   id: string;
   name: string;
+  description?: string;
   folder: string;
   skills: string[];
   mcpServers: string[];
@@ -38,3 +39,11 @@ export interface WorkspaceProjectInfo {
 export type WorkspaceProjectCreateOptions = Omit<WorkspaceProjectInfo, 'id'>;
 
 export type WorkspaceProjectUpdateOptions = Partial<Omit<WorkspaceProjectInfo, 'id'>>;
+
+export interface WorkspaceProjectAnalysis {
+  name: string;
+  description: string;
+  folder: string;
+  gitRepository?: string;
+  gitBranch?: string;
+}
--- a/packages/main/src/plugin/util/resolve-home-path.ts
+++ b/packages/main/src/plugin/util/resolve-home-path.ts
@@ -16,24 +16,12 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-function normalizeGitRepoPath(url: string): string {
-  return url
-    .trim()
-    .replace(/\/+$/, '')
-    .replace(/\.git$/, '');
-}
+import { homedir } from 'node:os';
+import { join } from 'node:path';
 
-export function formatGitUrl(url: string): string {
-  return normalizeGitRepoPath(url).replace(/^https?:\/\//, '');
-}
-
-export function extractRepoName(url: string): string {
-  const cleaned = normalizeGitRepoPath(url);
-  const lastSegment = cleaned.split('/').at(-1) ?? '';
-  return lastSegment.charAt(0).toUpperCase() + lastSegment.slice(1);
-}
-
-export function extractRepoSlug(url: string): string {
-  const cleaned = normalizeGitRepoPath(url);
-  return cleaned.split('/').at(-1) ?? 'project';
+export function resolveHomePath(inputPath: string): string {
+  if (inputPath.startsWith('~/')) {
+    return join(homedir(), inputPath.slice(2));
+  }
+  return inputPath;
 }
--- a/packages/main/src/plugin/workspace-project/workspace-project-manager.ts
+++ b/packages/main/src/plugin/workspace-project/workspace-project-manager.ts
@@ -16,7 +16,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { exec } from 'node:child_process';
+import { execFile } from 'node:child_process';
 import { existsSync } from 'node:fs';
 import { mkdir, readdir, readFile, rm, writeFile } from 'node:fs/promises';
 import { homedir } from 'node:os';
@@ -40,7 +40,7 @@ import type {
   WorkspaceProjectUpdateOptions,
 } from '/@api/workspace-project-info.js';
 
-const execAsync = promisify(exec);
+const execFileAsync = promisify(execFile);
 
 @injectable()
 export class WorkspaceProjectManager {
@@ -194,7 +194,7 @@ export class WorkspaceProjectManager {
     }
 
     try {
-      await execAsync(`git clone ${gitUrl} ${resolvedPath}`);
+      await execFileAsync('git', ['clone', gitUrl, resolvedPath]);
     } catch (err: unknown) {
       const message = err instanceof Error ? err.message : String(err);
       if (message.includes('Authentication') || message.includes('could not read Username')) {
@@ -213,8 +213,8 @@ export class WorkspaceProjectManager {
 
     try {
       const [remoteResult, branchResult] = await Promise.all([
-        execAsync('git remote get-url origin', { cwd: folderPath }).catch(() => undefined),
-        execAsync('git rev-parse --abbrev-ref HEAD', { cwd: folderPath }).catch(() => undefined),
+        execFileAsync('git', ['remote', 'get-url', 'origin'], { cwd: folderPath }).catch(() => undefined),
+        execFileAsync('git', ['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: folderPath }).catch(() => undefined),
       ]);
 
       const remote = remoteResult?.stdout.trim();
--- a/packages/main/src/plugin/workspace-project/workspace-project-manager.ts
+++ b/packages/main/src/plugin/workspace-project/workspace-project-manager.ts
@@ -19,7 +19,6 @@
 import { execFile } from 'node:child_process';
 import { existsSync } from 'node:fs';
 import { mkdir, readdir, readFile, rm, writeFile } from 'node:fs/promises';
-import { homedir } from 'node:os';
 import { basename, join } from 'node:path';
 import { promisify } from 'node:util';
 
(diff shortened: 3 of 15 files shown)


packages/api/src/workspace-project-info.ts
```diff
@@ -38,3 +39,11 @@ export interface WorkspaceProjectInfo {
...
 
 export type WorkspaceProjectUpdateOptions = Partial<Omit<WorkspaceProjectInfo, 'id'>>;
+
+export interface WorkspaceProjectAnalysis {
+  name: string;
+  description: string;
+  folder: string;
+  gitRepository?: string;
+  gitBranch?: string;
+}
```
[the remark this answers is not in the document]
MarsKubeX: The analyze() method always returns a description string. The type accurately reflects its implementation. The fact that WorkspaceProjectInfo.description is optional is irrelevant — these are different types with different contracts.

packages/renderer/src/lib/projects/ProjectCreate.svelte
```diff
@@ -0,0 +1,238 @@
...
+    const selected = result?.[0];
+    if (selected) {
+      const repoSlug = extractRepoSlug(gitUrl);
+      cloneTo = `${selected}/${repoSlug}`;
+    }
+  } catch (err: unknown) {
+    const message = err instanceof Error ? err.message : String(err);
+    toast.error(`Failed to browse for directory: ${message}`);
+  }
+}
```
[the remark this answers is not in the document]
MarsKubeX: Electron's openDialog returns OS-appropriate paths, Node.js and git handle / on Windows fine, and the backend normalizes via path.join.

packages/main/src/plugin/workspace-project/workspace-project-manager.ts
```diff
@@ -144,6 +164,68 @@ export class WorkspaceProjectManager {
...
+    const gitInfo = await this.detectGitInfo(folderPath);
+
+    return {
+      name: name.charAt(0).toUpperCase() + name.slice(1),
+      description: 'Auto-detected project from local working directory.',
+      folder: folderPath,
+      gitRepository: gitInfo?.remote,
+      gitBranch: gitInfo?.branch,
+    };
+  }
```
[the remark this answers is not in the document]
MarsKubeX: The analyze() method is only reachable via IPC from two controlled sources: (1) the renderer after the user selects a path from the native Browse dialog or types one that already passes existsSync, and (2) internally from cloneAndAnalyze() after a successful git clone. In both cases, the path is already validated or known-valid by construction. Additionally, existsSync at line 168 will reject any path containing null bytes or control characters (it returns false), so the method already throws a clear "does not exist" error for malformed input. Adding explicit sanitization here would be defensive hardening with no practical benefit for a desktop Electron app where the user is the sole input source.

packages/main/src/plugin/workspace-project/workspace-project-manager.ts
```diff
@@ -144,6 +164,68 @@ export class WorkspaceProjectManager {
...
+
+  private async detectGitInfo(folderPath: string): Promise<{ remote?: string; branch?: string } | undefined> {
+    if (!existsSync(join(folderPath, '.git'))) {
+      return undefined;
+    }
+
+    try {
+      const [remoteResult, branchResult] = await Promise.all([
+        execFileAsync('git', ['remote', 'get-url', 'origin'], { cwd: folderPath }).catch(() => undefined),
```
jeffmaury: **issue:** you can configure git to clone using remote other than origin so I would either get the remote from the git config or use origin then upstream or the first remote

packages/main/src/plugin/workspace-project/workspace-project-manager.ts
```diff
@@ -257,6 +339,13 @@ export class WorkspaceProjectManager {
     return normalized;
   }
 
+  private resolveHomePath(inputPath: string): string {
```
jeffmaury: **suggestion:** maybe move it to a utility

packages/renderer/src/lib/projects/ProjectEmptyScreen.svelte
```diff
@@ -1,10 +1,19 @@
...
+function navigateToCreate(): void {
+  handleNavigation({ page: NavigationPage.PROJECT_CREATE });
+}
 </script>
 
 <EmptyScreen title="No projects" icon={faFolderOpen}>
   <span class="text-(--pd-content-text)">
     Projects will appear here once created.
   </span>
+  <Button icon={faPlus} onclick={navigateToCreate}>New Project</Button>
```
gastoner: Can we move this to another line ?

packages/renderer/src/lib/projects/ProjectCreateStepReview.svelte
```diff
@@ -0,0 +1,141 @@
...
+    <div>
+      <span class="block text-xs font-semibold uppercase tracking-wider text-(--pd-content-card-text) opacity-70 mb-2">
+        Git Repository
+      </span>
+      <div class="flex items-center gap-2 rounded-lg bg-(--pd-content-card-bg) px-3 py-2.5">
+        <Icon icon={faCodeBranch} class="text-(--pd-content-card-text) opacity-50" size="sm" />
+        <span class="text-sm font-mono text-(--pd-content-card-text)">{gitRepoDisplay}</span>
+        <span class="ml-auto text-xs font-medium bg-(--pd-label-quaternary-bg) text-(--pd-label-quaternary-text) px-2 py-0.5 rounded">
+          default
+        </span>
```
gastoner: The 'default' should be like 'main' branch I think?

gastoner (commented):
Other suggestions:
1. add some progress-bar/ task etc to the kaiden the cloning e.g. of podman-desktop takes like 5min even with a good network connection
2. changing the 'default' to the actual default branch name

packages/renderer/src/lib/projects/ProjectCreateStepReview.svelte
MarsKubeX: I've thought to let "main", but it's true that not all repositories has main as default. That's why a let just "default"

packages/main/src/plugin/workspace-project/workspace-project-manager.ts
MarsKubeX: Fixed. Check first if origin exists, if not, it list the remotes and takes the first one.

packages/main/src/plugin/workspace-project/workspace-project-manager.ts
MarsKubeX: Moved to a utility.

packages/renderer/src/lib/projects/ProjectEmptyScreen.svelte
MarsKubeX: Moved.

packages/renderer/src/lib/projects/ProjectCreateStepReview.svelte
gastoner: Maybe we can hide this when the registry is provided by an url and show this only when we select the folder? WDYT?

gastoner (approved):
LGTM Otherwise
Also WDYT about adding the progress bar when cloning the repo ?

packages/api/src/workspace-project-info.ts
```diff
@@ -38,3 +39,11 @@ export interface WorkspaceProjectInfo {
 export type WorkspaceProjectCreateOptions = Omit<WorkspaceProjectInfo, 'id'>;
 
 export type WorkspaceProjectUpdateOptions = Partial<Omit<WorkspaceProjectInfo, 'id'>>;
+
+export interface WorkspaceProjectAnalysis {
+  name: string;
+  description?: string;
```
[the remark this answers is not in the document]
MarsKubeX: We changed with one of your prior comments to be optional and consistent. Even if the analyze returns a description string the user can erase it.

MarsKubeX:
> LGTM Otherwise Also WDYT about adding the progress bar when cloning the repo ?

Thanks for the feedback.

Regarding the progress bar, it was included in the last commit.

Captura de pantalla 2026-06-04 a las 14 53 43

For hiding the git url and branch ("default"), I think we should keep the git url, since it helps as a user verification on which repo is going to be clone, an the branch ("default") it can be removed, since it does not adds info. WDYT?

gastoner:
Sorry havent noticed the progress bar :D 
Yes I ment the `an the branch ("default") it can be removed,` should have specify this more :D keep the git url, just remove the 'default' label 🚀

packages/main/src/plugin/util/resolve-home-path.ts
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { homedir } from 'node:os';
+import { join } from 'node:path';
+
+export function resolveHomePath(inputPath: string): string {
+  if (inputPath.startsWith('~/')) {
+    return join(homedir(), inputPath.slice(2));
+  }
```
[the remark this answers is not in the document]
MarsKubeX: The bare ~ case is not a valid user input for a clone target or project folder — nobody would clone directly into their home directory root. The ~\ convention doesn't exist on Windows (users use Browse dialogs or absolute paths). Only ~/subpath is a realistic scenario and it's handled.

packages/renderer/src/lib/projects/ProjectCreate.svelte
```diff
@@ -0,0 +1,225 @@
...
+});
+
+function goBack(): void {
+  if (currentStepIndex > 0) currentStepIndex--;
+}
+
+function handleStepClick(index: number): void {
+  if (index > currentStepIndex && !isGitSource && !analysis) {
+    return;
+  }
```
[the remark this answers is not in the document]
MarsKubeX: The wizard flow already prevents this scenario. Changing sourcePath puts the user back on Step 1 where handleStepClick blocks forward navigation unless analysis exists. The only way to reach the Create button is through handleAnalyze() which always updates analysis to match the current input.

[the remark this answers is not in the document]
MarsKubeX: Should be fixed now.

MarsKubeX:
@coderabbitai resolve

jeffmaury (approved):

PR #2075, opened 2026-06-04
fix: display workspace name instead of ID when deleting

### LINKED ISSUE #1778: Deleting a workpace reports progress with id not name
### Bug description
When deleting a workspace, we are displaying its ID, instead of its name. 

Image

Should align with create

### Operating system

All

### Installation Method

None

### Version

0.0.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

fbricon:
When deleting a workspace, the task notification now shows the
workspace name instead of the workspace ID, matching the behavior
shown when creating a workspace.

The implementation fetches the workspace list first to lookup the
name, falling back to the ID if the workspace is not found in the
list.

Updated tests to verify the new task title format and added a test
case for the fallback scenario.

Fixes #1778

Screenshot 2026-06-04 at 11 28 09
Screenshot 2026-06-04 at 11 28 38

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -816,6 +816,7 @@ describe('list', () => {
 
 describe('remove', () => {
   test('delegates to kdnCli.remove and returns the workspace id', async () => {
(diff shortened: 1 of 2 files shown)


slemeur:
Nice!!

MarsKubeX (approved):

PR #2076, opened 2026-06-04
feat(api): add command and ACP configuration to Agent interface

benoitf:
Extend the extension API with a required `command` field on Agent (the agent's CLI binary) and an optional `acp` field declaring Agent Client Protocol support with the args needed to enable ACP mode.

related to https://github.com/openkaiden/kaiden/issues/1935


jeffmaury (approved):
LGTM.

Just wondering if the command field is not required only for ACP

benoitf:
it's required to execute the command inside the sandbox (if the image has for example multiple agents)

benoitf:
before it was `kdn` that was in charge to find the right command to execute for a given agent's name

PR #2081, opened 2026-06-04
fix(milvus): sanitize connection name to prevent container creation errors

### LINKED ISSUE #2015: Can not create a Milvus Vector Database if name contains spaces
### Bug description

Can not create a Milvus Vector Database is name contains spaces, it errors with:

> Error: (HTTP code 500) server error - container create: running container create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument

Image

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

Go to Resources > Milvus > Create New...
Enter a name containing space(s)
Click Create

### Relevant log output

```shell

```

### Additional context

Either the UI should show an error message as soon as a space is entered in the field, or the name should be sanitized before creating the resource

bmahabirbu:
## Summary
- Sanitize user-provided Milvus connection names by replacing characters that don't match `[a-zA-Z0-9_.-]` with hyphens, preventing container runtime 500 errors when names contain spaces or special characters
- Apply sanitized name consistently to container name, storage path, and discovery labels
- Add test covering the spaces-in-name scenario

Fixes: https://github.com/openkaiden/kaiden/issues/2015

### DIFF
--- a/extensions/milvus/src/manager/connection-manager.ts
+++ b/extensions/milvus/src/manager/connection-manager.ts
@@ -29,7 +29,7 @@ import {
   RagProviderConnectionFactory,
 } from '@openkaiden/api';
 import { ContainerExtensionAPI } from '@openkaiden/container-extension-api';
-import { sanitizeContainerName } from '@openkaiden/container-extension-api/src/container-name';
+import { sanitizeContainerName } from '@openkaiden/container-extension-api/container-name';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import Dockerode from 'dockerode';
 import { inject, injectable } from 'inversify';
--- a/extensions/milvus/src/manager/connection-manager.ts
+++ b/extensions/milvus/src/manager/connection-manager.ts
@@ -29,7 +29,7 @@ import {
   RagProviderConnectionFactory,
 } from '@openkaiden/api';
 import { ContainerExtensionAPI } from '@openkaiden/container-extension-api';
-import { sanitizeContainerName } from '@openkaiden/container-extension-api/container-name';
+import { sanitizeContainerName } from '@openkaiden/container-extension-api/src/container-name';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import Dockerode from 'dockerode';
 import { inject, injectable } from 'inversify';
--- a/extensions/milvus/src/manager/connection-manager.ts
+++ b/extensions/milvus/src/manager/connection-manager.ts
@@ -227,14 +227,16 @@ export class ConnectionManager implements Disposable {
 
     logger?.log(`Connection name: ${name}`);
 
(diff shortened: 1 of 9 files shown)


bmahabirbu:
@coderabbit resolve

extensions/milvus/src/manager/connection-manager.ts
```diff
@@ -227,14 +227,16 @@ export class ConnectionManager implements Disposable {
 
     logger?.log(`Connection name: ${name}`);
 
+    const safeName = name.replace(/[^a-zA-Z0-9_.-]/g, '-').replace(/^[^a-zA-Z0-9]/, 'x');
```
fbricon: this should go into a utility function, as it'll be needed elsewhere, I think the same problem occurs for any extension using podman containers

fbricon (changes requested):

extensions/milvus/src/manager/connection-manager.ts
bmahabirbu: Hmm it seems there isnt currently a standard for having a shared utility for extensions in kaiden yet. we could add a util.ts to the container-extension-api package alongside the existing .d.ts file But i think thats out of scope for this pr? What are your thoughts?

bmahabirbu: the code is quite small too it could be inline in each extensions that needs it

jeffmaury (commented):
There is a side effect is that it changes the name of the Milvus connection that is given by the user.

MarsKubeX (approved):

bmahabirbu:
@fbricon updated!!

jeffmaury (approved):

extensions/milvus/src/manager/connection-manager.ts
```diff
@@ -227,14 +230,16 @@ export class ConnectionManager implements Disposable {
...
+
```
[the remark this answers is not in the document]
fbricon: I kind of agree with this, but I'm not sure how extensive such change would be

extensions/milvus/src/manager/connection-manager.ts
```diff
@@ -227,23 +230,34 @@ export class ConnectionManager implements Disposable {
...
+    }
+    const dockerode = containerConnection.dockerode;
+    const existingContainers = await dockerode.listContainers({ all: true });
+    const existingNames = new Set(existingContainers.flatMap(c => c.Names ?? []));
+    let suffix = 1;
+    const baseSafeName = safeName;
+    while (existingNames.has(`/milvus-${safeName}`)) {
+      suffix++;
+      safeName = `${baseSafeName}-${suffix}`;
+    }
```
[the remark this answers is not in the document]
fbricon: @bmahabirbu this

fbricon (approved):

bmahabirbu:
@coderabbit resolve

PR #2104, opened 2026-06-09
fix(renderer): load UI state before Keychain authentication

### LINKED ISSUE #2091: On initial startup, styles are not loaded until after Keychain authentication
### Bug description

On initial startup after installing from dmg, styles are not loaded until after Keychain authentication

Image

### Operating system

macOS

### Installation Method

Other

### Version

0.2.x

### Steps to reproduce

- install Kaiden from a dmg
- start it, Keychain auth prompt shows up, meanwhile, styles are not loaded (app uses black background)

### Relevant log output

```shell

```

### Additional context

_No response_

fbricon:
Fixes #2091

On macOS initial startup from DMG, UI was not properly initialized until
after Keychain authentication completed. Window showed with black background
and missing navigation labels while waiting for user to authenticate.

Root cause: SafeStorageRegistry.init() was called before ColorRegistry,
ConfigurationRegistry, and their IPC handlers were registered. When the
renderer loaded, critical IPC handlers were unavailable, preventing colors,
configuration properties, and navigation from loading until the entire
initialization sequence completed (including Keychain auth).

Changes:
- Reordered initialization: ConfigurationRegistry and ColorRegistry now
  initialize before SafeStorageRegistry
- Registered configuration and color IPC handlers immediately after their
  respective registries initialize
- Added immediate fetch in EventStores for colors, configuration properties,
  and navigation registry instead of waiting for events
- Added undefined guard in ColorsStyle for test environments

Result: Colors, configuration, and navigation labels all load immediately
when the renderer starts, before SafeStorageRegistry initialization. The UI
is fully styled and functional even when the Keychain authentication prompt
appears during extension activation.

Screenshot 2026-06-09 at 11 56 28

### DIFF
--- a/packages/renderer/src/AppNavigation.spec.ts
+++ b/packages/renderer/src/AppNavigation.spec.ts
@@ -75,6 +75,7 @@ beforeEach(() => {
   vi.resetAllMocks();
   callbacks.clear();
   vi.useFakeTimers({ shouldAdvanceTime: true });
+  vi.mocked(window.getConfigurationValue).mockResolvedValue(AppearanceSettings.IconAndTitle);
   onDidChangeConfiguration.addEventListener = vi.fn().mockImplementation((message: string, callback: () => void) => {
     callbacks.set(message, callback);
   });
@@ -99,6 +100,7 @@ test('Test rendering of the navigation bar with empty items', async (_arg: unkno
     meta,
     exitSettingsCallback: () => {},
   });
+  await tick();
 
   const navigationBar = screen.getByRole('navigation', { name: 'AppNavigation' });
   expect(navigationBar).toBeInTheDocument();
--- a/packages/renderer/src/AppNavigation.svelte
+++ b/packages/renderer/src/AppNavigation.svelte
@@ -24,7 +24,7 @@ interface Props {
 }
 let { exitSettingsCallback, meta = $bindable() }: Props = $props();
 
-let iconWithTitle = $state(false);
+let iconWithTitle = $state<boolean | undefined>(undefined);
 
 const iconSize = '22';
 const NAV_BAR_LAYOUT = `${AppearanceSettings.SectionName}.${AppearanceSettings.NavigationAppearance}`;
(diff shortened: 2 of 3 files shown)


packages/renderer/src/stores/colors.ts
```diff
@@ -45,4 +45,10 @@ export const colorsEventStore = new EventStore<ColorInfo[]>(
...
   listColors,
 );
-colorsEventStore.setup();
+const eventStoreInfo = colorsEventStore.setup();
+
+// Trigger initial color load immediately so colors are available before window shows
+// This prevents the black background flash on initial startup
+eventStoreInfo.fetch().catch((error: unknown) => {
+  console.error('Failed to load initial colors', error);
+});
```
gastoner: colorsEventStore.setup().fetch()?

packages/renderer/src/stores/configurationProperties.ts
```diff
@@ -39,7 +39,12 @@ const eventStore = new EventStore<IConfigurationPropertyRecordedSchema[]>(
   windowListeners,
   fetchConfigurationProperties,
 );
-eventStore.setup();
+const eventStoreInfo = eventStore.setup();
+
+// Trigger initial config load immediately so navigation can initialize
+eventStoreInfo.fetch().catch((error: unknown) => {
+  console.error('Failed to load initial configuration properties', error);
+});
```
gastoner: Same her

gastoner (commented):
Personally I would move the registration of IPC handles to init methods of colorRegistry and configurationRegistry

LGTM otherwise

fbricon:
> Personally I would move the registration of IPC handles to init methods of colorRegistry and configurationRegistry

This indirection idea is sometimes interesting, but here, I'd rather not introduce new couplings to these registries. At least we have a mostly consistent way of registering IPC handles here.

packages/renderer/src/stores/colors.ts
fbricon: done

packages/renderer/src/stores/configurationProperties.ts
fbricon: done

benoitf (changes requested):
this is something we inherit from rebasing from Podman Desktop, we should not change any logic only for kaiden

PR #2113, opened 2026-06-09
chore: fix lockfile

jeffmaury:


fbricon (approved):

bmahabirbu (approved):
Looks good thanks for the fix! seems all that was needed was just to remove the duplicates

PR #2114, opened 2026-06-10
fix(openshell): refresh sandbox list on page navigation

bmahabirbu:
## Summary
- Re-fetch the OpenShell sandbox list whenever the Agentic Workspaces page mounts, so externally deleted sandboxes are detected on page navigation
- OpenShell sandboxes are remote (gateway-managed) resources with no local state file to watch — unlike kdn workspaces which use `~/.kdn/instances.json` — so the existing file watcher cannot detect external deletions
- Exports the `EventStoreInfo` from the openshell sandboxes store so components can call `.fetch()` to force a re-fetch

### DIFF
--- a/packages/renderer/src/App.spec.ts
+++ b/packages/renderer/src/App.spec.ts
@@ -100,6 +100,7 @@ beforeEach(() => {
   };
   Object.defineProperty(window, 'dispatchEvent', { value: dispatchEventMock });
   (window.getConfigurationValue as unknown) = vi.fn();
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
Should be managed by the back end component

bmahabirbu:
Right now I beleive the backend has no way to know when OpenShell sandboxes change externally

There is an api to watch one sandbox in openshell but not a fleet. This might be a good alternative till we have the ability to fully watch all sandboxes

PR #2140, opened 2026-06-11
chore: downgrade windows runner

benoitf:
to downgrade Visual studio 2026 that is not working with older node-gyp


serbangeorge-m (approved):

jeffmaury (approved):
LGTM

We should probably update the list of required jobs

benoitf:
@jeffmaury yes but I can't update the required jobs until job is completed (else I don't find it in the list)

PR #2143, opened 2026-06-12
fix: remove double-prefix on openshell config key lookup

### LINKED ISSUE #2142: openshell extension does not seems to look to the openshell settings
### Bug description

If I configure a path for the openshell binary (same for the openshell-image-builder) then it seems to be ignored

### Operating system

All

### Installation Method

None

### Version

0.2.x

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
## Summary
- `getConfiguration('openshell').get('openshell.binary.path')` was resolving to `openshell.openshell.binary.path`, silently ignoring any user-configured custom binary path
- 
Fixes: https://github.com/openkaiden/kaiden/issues/2142

There is another issue at play here

since openshell needs multiple binaries and only the main openshell binary is loaded we should have the ability to select a custom path for the gateway binary and sandbox? Wondering if we should do a directory walkthrough on a folder having all the binaries? or seperate path configs for each one? Right now only openshell is registered in the clitools

its out of scope for this particular issue but definitely related

### DIFF
--- a/extensions/openshell/src/manager/openshell-cli-manager.ts
+++ b/extensions/openshell/src/manager/openshell-cli-manager.ts
@@ -48,7 +48,7 @@ export class OpenshellCliManager implements Disposable {
     const packageJsonPath = join(this.extensionContext.extensionUri.fsPath, 'package.json');
     const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'));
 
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):

jeffmaury:
There is no need to have settings for other binaries as it may introduce playing with different versions which is error prone

PR #2149, opened 2026-06-12
fix: compute agent supported types on the inference connections factories

### LINKED ISSUE #2148: agent registry computes supportedTypes for an agent only if at least one inference connection exists
### Bug description

The supported types for an agent is computed based on existing models so if no models exists it is not computed

### Operating system

All

### Installation Method

None

### Version

0.2.x

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

jeffmaury:
Fixes #2148


gastoner (approved):
Codewiese LGTM

PR #2187, opened 2026-06-16
feat: use new agent registration and implement workspace creation on OpenShell

### LINKED ISSUE #2186: Update OpenShell create workspace
### Task content

- remove call to the image builder
- integrate the new agent preWorkspaceStart
- upload the configuration files

See #2168 
See #2135

jeffmaury:
To test:
- define env variable KAIDEN_OPENSHELL
- create a workspace (excluding Gemini models before #2199 is merged)
- check the OpenCode configuration file inside the sandbox

Fixes #2186

Replace Docker image building logic in AgentWorkspaceManager with agent-driven preWorkspaceStart() hooks. Agents now customize their own configuration files before workspace startup, enabling more flexible and agent-specific setup.

- Add WritableConfigurationFile wrapper for mutable agent config
- Add getAgentRegistration() method to AgentRegistry
- Remove OpenshellImageBuilder dependency from workspace manager
- Upload agent config files to sandbox instead of baking into images

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -17,7 +17,7 @@
  ***********************************************************************/
 
 import { access, readFile, rm, writeFile } from 'node:fs/promises';
-import { homedir } from 'node:os';
+import { homedir, tmpdir } from 'node:os';
 import { basename, join } from 'node:path';
 
 import type { Disposable, FileSystemWatcher } from '@openkaiden/api';
@@ -26,11 +26,12 @@ import { inject, injectable, preDestroy } from 'inversify';
 import type { IPty } from 'node-pty';
 import { spawn } from 'node-pty';
 
+import { AgentRegistry } from '/@/plugin/agent-registry.js';
+import { WritableConfigurationFile } from '/@/plugin/agent-workspace/writable-configuration-file.js';
 import { IPCHandle, WebContentsType } from '/@/plugin/api.js';
 import { FilesystemMonitoring } from '/@/plugin/filesystem-monitoring.js';
 import { KdnCli } from '/@/plugin/kdn-cli/kdn-cli.js';
 import { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
-import { OpenshellImageBuilder } from '/@/plugin/openshell-cli/openshell-image-builder.js';
 import { ProviderRegistry } from '/@/plugin/provider-registry.js';
 import { SafeStorageRegistry } from '/@/plugin/safe-storage/safe-storage-registry.js';
 import { SecretManager } from '/@/plugin/secret-manager/secret-manager.js';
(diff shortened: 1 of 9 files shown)


packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -121,43 +122,51 @@ export class AgentWorkspaceManager implements Disposable {
...
-      endpoint,
-      cwd: options.sourcePath,
-    });
 
     await this.openshellCli.createSandbox({
       name: sandboxName,
-      from: imageTag,
       providers: options.secrets,
       labels: { 'ai.openkaiden.kaiden.workspace': Buffer.from(options.sourcePath).toString('base64url') },
+      uploads: uploads.length > 0 ? uploads : undefined,
```
benoitf, line 171: should use the mount rather than upload

(can be a separate task but if the gateway is local + driver is supporting the option) then we need to send the mount through 

so probably need a `mount` options


```
--driver-config-json '{"podman":{"mounts":[{"type":"bind","source":"/Users/benoitf/git/tmp/hello","target":"/sandbox/work"}]}}
```

benoitf:
looks like the title and the start of the body of this PR are corrupted

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
jeffmaury: Agree with will be a follow up issue as I'm not even sure we can get this detail about gateway ?

fbricon:
so it technically works but I ran into https://github.com/openkaiden/kaiden/issues/2204, which is an orthogonal issue

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -121,43 +122,51 @@ export class AgentWorkspaceManager implements Disposable {
     }
   }
```
bmahabirbu, line 125: Maybe we can have a guardrail here so that for openshell we dont suppot skills mcp etc yet

something like

```
const hasUnsupportedOptions =
  !!options.skills?.length ||
  !!options.mcp?.servers?.length ||
  !!options.mcp?.commands?.length ||
  !!options.network ||
  !!options.mounts?.length ||
  Object.keys(options.workspaceConfiguration ?? {}).length > 0;
if (hasUnsupportedOptions) {
  console.warn('[openshell] native OpenShell workspace creation received unsupported options', options);
  throw new Error(
    'Native OpenShell workspace creation currently supports only the basic setup flow. Remove MCP, skills, network, mounts, or custom workspace settings, or use the Podman runtime instead.',
  );
}
```

jeffmaury: Will be follow up issue keeping this one errorless so that people can start testing

benoitf (approved):

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -121,43 +122,51 @@ export class AgentWorkspaceManager implements Disposable {
...
     await this.kdnCli.writeWorkspaceConfig(options);
 
+    const agent = this.agentRegistry.getAgentRegistration(options.agent);
+    const uploads: Array<{ local: string; remote: string }> = [];
+
+    if (agent) {
+      const writable = await Promise.all(
+        agent.configurationFiles.map(
+          async (base, i) =>
+            new WritableConfigurationFile(base, await base.read(), join(tmpdir(), `kaiden-config-${Date.now()}-${i}`)),
```
MarsKubeX, line 143: Question: Should this tmpdir() be deleted after creation, or in case of error?

jeffmaury:
@coderabbitai resolve

PR #2198, opened 2026-06-17
ci: use frozen lockfile for dependency installation in all CI jobs

fbricon:
Replace all `pnpm install` calls in GitHub Actions workflows with
`pnpm install --frozen-lockfile`, ensuring CI jobs never modify
pnpm-lock.yaml. This mirrors `npm ci` behavior.

Rename install steps to 'Install dependencies' for consistency.

Migration to pnpm 11 will allow us to use `pnpm ci` eventually.

### DIFF
--- a/.github/workflows/e2e-test-job.yaml
+++ b/.github/workflows/e2e-test-job.yaml
@@ -92,8 +92,8 @@ jobs:
           node-version: 24.15.0
           cache: 'pnpm'
 
-      - name: Execute pnpm
-        run: pnpm install
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
 
       - name: Get Playwright version and cache path
         id: playwright-info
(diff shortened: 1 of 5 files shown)


jeffmaury:
Is that aligned with podman desktop?

fbricon:
> Is that aligned with podman desktop?

no. PD should do it too, but it uses a lot more workflows.

benoitf (changes requested):
AFAIK this is useless in a CI environment

benoitf:
we don't set frozen option as it's the default for pnpm on CI environments

see https://pnpm.io/cli/install#--frozen-lockfile

fbricon:
mmm then why does CI fail when it detects changes to the lockfile ?

benoitf:
> mmm then why does CI fail when it detects changes to the lockfile ?

not sure to get your question ?

fbricon:
@benoitf that kind of failure https://github.com/openkaiden/kaiden/actions/runs/27680005260/job/81864789796?pr=2167

PR #2209, opened 2026-06-18
feat(api): add skillsFolder field to Agent interface

### LINKED ISSUE #2207: Agents implementation should specify folder where skills are to be copied

This field is required for the core to copy the skills

- Add a new skillsFolder field in the extension API Agent type
- Update accordingly the claude, codex, copilot, cursor, gemini, goose, openclaw, opencode extensions

_No response_

### Additional context

_No response_

jeffmaury:
## Summary
- Add a required `skillsFolder` property to the `Agent` and `AgentInfo` types so the core can copy skills to the correct agent-specific directory during workspace creation
- Update `AgentRegistry.toAgentInfo()` to pass `skillsFolder` through
- Update all 8 agent extensions (claude, codex, copilot, cursor, gemini, goose, openclaw, opencode) to provide their `skillsFolder` path

## Test plan
- [x] `pnpm run typecheck` passes
- [x] `pnpm run test:main` passes (agent-registry + extension-loader tests)
- [x] `pnpm run test:extensions` passes (all 8 agent extension tests)
- [x] Renderer workspace-create tests pass

Fixes #2207

### DIFF
--- a/extensions/claude/src/claude-extension.ts
+++ b/extensions/claude/src/claude-extension.ts
@@ -63,7 +63,7 @@ export class ClaudeExtension {
       command: 'claude',
       tags: ['Cloud'],
       configurationFiles: [],
-      skillsFolder: '.claude/skills',
+      destinationSkillsFolder: '.claude/skills',
       isSupportedModelType: (type): boolean => type.name === 'anthropic' || type.name === 'vertexai',
       async preWorkspaceStart(): Promise<void> {
         throw new Error('not implemented');
--- a/extensions/claude/src/claude-extension.ts
+++ b/extensions/claude/src/claude-extension.ts
@@ -16,9 +16,6 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { homedir } from 'node:os';
-import { join } from 'node:path';
-
 import type { Disposable, ExtensionContext } from '@openkaiden/api';
 import { agents, provider } from '@openkaiden/api';
 import type { Container } from 'inversify';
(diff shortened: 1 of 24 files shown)


extensions/claude/src/claude-extension.ts
```diff
@@ -63,6 +66,7 @@ export class ClaudeExtension {
       command: 'claude',
       tags: ['Cloud'],
       configurationFiles: [],
+      skillsFolder: join(homedir(), '.claude', 'skills'),
```
benoitf: it's not clear to me when reading the property if it's a source or destination folder

I was assuming it's the destination because I don't want to have the agent copying all my skills by default from my environment (as it may contain some stuff I don't want in the sandbox) but I would assume this is where we need to copy the skills


but then if it says 'homeDir()' it resolves to my local computer then I'm confused as I'm thinking it's a source directory ?

jeffmaury: Good catch, fixed now

benoitf (commented):
I think it should contains a variable to be replaced later

like `${HOME}/.claude/skills` or `%HOME%/.claude/skills` or `{HOME}/.claude/skills` so we know it would land to the home directory

or a complex struct

`skillsFolder` with attribute `usersHomeDirectory` and then we can set the value `.agents/skills`

benoitf (commented):
and I would prefix skillsFolder by `dest` to really know it's a destination folder, not a source folder

jeffmaury:
> and I would prefix skillsFolder by `dest` to really know it's a destination folder, not a source folder

Fixed by 36d3ba2

jeffmaury:
> I think it should contains a variable to be replaced later
> 
> like `${HOME}/.claude/skills` or `%HOME%/.claude/skills` or `{HOME}/.claude/skills` so we know it would land to the home directory
> 
> or a complex struct
> 
> `skillsFolder` with attribute `usersHomeDirectory` and then we can set the value `.agents/skills`

Fixed by 9d5fa40

benoitf (approved):
thanks for the changes

bmahabirbu (approved):
LGTM

jeffmaury:
@coderabbitai resolve

PR #2216, opened 2026-06-18
refactor(gemini): use zod to validate parsed settings.json

MarsKubeX:
Replaces manual typeof/Array.isArray guards with a Zod schema for parsing the configuration file in preWorkspaceStart, aligning with the project's established validation pattern.

### DIFF
--- a/extensions/gemini/src/extension.ts
+++ b/extensions/gemini/src/extension.ts
@@ -18,10 +18,13 @@
 
 import type { AgentWorkspaceContext, ExtensionContext } from '@openkaiden/api';
 import { agents, provider } from '@openkaiden/api';
+import { z } from 'zod';
 
 import { Gemini } from './gemini';
 
 export const GEMINI_SETTINGS_PATH = '.gemini/settings.json';
+
+const GeminiSettingsSchema = z.record(z.string(), z.unknown()).catch({});
 
 export async function activate(extensionContext: ExtensionContext): Promise<void> {
   console.log('starting gemini extension');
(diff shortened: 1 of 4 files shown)


extensions/gemini/src/extension.ts
```diff
@@ -60,17 +63,14 @@ export async function activate(extensionContext: ExtensionContext): Promise<void
...
-          typeof parsed === 'object' && parsed !== null && !Array.isArray(parsed)
-            ? (parsed as Record<string, unknown>)
-            : {};
+        parsed = JSON.parse(content);
       } catch {
-        config = {};
+        parsed = {};
       }
 
+      const config = GeminiSettingsSchema.parse(parsed);
```
benoitf: hello, I'm not sure why we have two parse ? one with JSON parse and then the other one with zod

MarsKubeX: What I have in mind is to first parse if the string is JSON valid (first parse) and second if follows the zod schema (second parse).

benoitf: but why not let do everything by zod ?

MarsKubeX: Good point, it should be clearer. I've changed so the zod pipeline also check if the string is valid.

extensions/gemini/src/extension.ts
```diff
@@ -18,11 +18,24 @@
 
 import type { AgentWorkspaceContext, ExtensionContext } from '@openkaiden/api';
 import { agents, provider } from '@openkaiden/api';
+import { z } from 'zod';
 
 import { Gemini } from './gemini';
 
 export const GEMINI_SETTINGS_PATH = '.gemini/settings.json';
 
+const GeminiSettingsSchema = z
```
jeffmaury: **suggestion:** why not using:

`z.object({ model: z.object({ name: z.string() }) })`

which seems way more readable

jeffmaury (commented):
otherwise LGTM

extensions/gemini/src/extension.ts
MarsKubeX: z.object() would discard unknown keys from the existing config by default and fail validation on first run when model doesn't exist yet. We need z.record() here because we're preserving arbitrary existing fields and writing model into them, not reading/validating a known shape.

jeffmaury: looseObject seems your friend

MarsKubeX: Should use looseObject now.

jeffmaury (approved):
LGTM. I'm just wondering if we should use looseObject at all levels

MarsKubeX:
> LGTM. I'm just wondering if we should use looseObject at all levels

Good idea if in the future the logic changes to merge instead of overwriting.

benoitf (approved):

PR #2223, opened 2026-06-18
fix: split workspace label to respect 63-char limit

### LINKED ISSUE #2218: Sandbox creation fails if label > 63 char
### Bug description

I tried creating an openshell sandbx and kaiden failed with:

> openshell failed: /opt/homebrew/bin/openshell sandbox create --name backend --label ai.openkaiden.kaiden.workspace=L1VzZXJzL2Zicmljb24vRGV2L3NvdWsvaWRlYXMvc3RvY2stdHJhZGluZy1haS9iYWNrZW5k --upload /var/folders/zn/5w5phtl94v13jm4sh7cxdh4m0000gn/T/kaiden-config-1781785554124-0:.config/opencode/opencode.json --no-tty -- true — Command execution failed with exit code 1

When calling the same command in a terminal, we get a bit more detail:

> openshell sandbox create --name backend --label ai.openkaiden.kaiden.workspace=L1VzZXJzL2Zicmljb24vRGV2L3NvdWsvaWRlYXMvc3RvY2stdHJhZGluZy1haS9iYWNrZW5k --upload /var/folders/zn/5w5phtl94v13jm4sh7cxdh4m0000gn/T/kaiden-config-1781785554124-0:.config/opencode/opencode.json --no-tty -- true
Error:   × code: 'Client specified an invalid argument', message: "label value exceeds 63 characters: 'L1VzZXJzL2Zicmljb24vRGV2L3NvdWsvaWRlYXMvc3RvY2stdHJhZGluZy1haS9iYWNrZW5k'"

The label or (or workspace id) generator should limit the result length to 63 characters

### Operating system

macOS

### Installation Method

None

### Version

0.2.x

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

fbricon:
The workspace label stores a base64url-encoded source path so it can
be decoded and displayed in the UI. Long paths can exceed the 63-char
label value limit, causing sandbox creation to fail.

Add encodeWorkspaceLabels/decodeWorkspaceLabels functions that use a
single label when the value fits, and split into indexed labels
(workspace.0, workspace.1, ...) when it exceeds the limit.

The decoded workspace path is shown as a tooltip on the sandbox name
in the workspace list.
Closes #2218

### DIFF
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -38,7 +38,43 @@ export const SandboxInfoSchema = z.object({
   resource_version: z.number().optional(),
 });
 
-export type SandboxInfo = z.output<typeof SandboxInfoSchema>;
+export type SandboxInfo = z.output<typeof SandboxInfoSchema> & {
+  sourcePath?: string;
+};
+
+export const WORKSPACE_LABEL = 'ai.openkaiden.kaiden.workspace';
+
+export function decodeWorkspaceLabels(labels: Record<string, string>): string | undefined {
+  let encoded: string;
+  if (WORKSPACE_LABEL in labels) {
+    encoded = labels[WORKSPACE_LABEL]!;
+  } else {
+    const chunks = Object.entries(labels)
+      .flatMap(([key, value]) => {
+        if (!key.startsWith(`${WORKSPACE_LABEL}.`)) {
+          return [];
+        }
+        const suffix = key.slice(WORKSPACE_LABEL.length + 1);
+        if (!/^\d+$/.test(suffix)) {
+          return [];
+        }
+        return [{ index: Number(suffix), value }];
+      })
+      .sort((a, b) => a.index - b.index);
+    if (chunks.length === 0 || chunks.some((chunk, i) => chunk.index !== i)) {
+      return undefined;
+    }
+    encoded = chunks.map(chunk => chunk.value).join('');
+  }
+  try {
+    const base64 = encoded.replace(/-/g, '+').replace(/_/g, '/');
+    const binary = atob(base64);
+    const bytes = Uint8Array.from(binary, c => c.charCodeAt(0));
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
I am not sure to understand: the idea of this label is to retrieve the path used by Kaiden when the workspace was created so that we can locate the workspace.json when we need to display the details. If this is a hash how do we retrieve the full path ?

fbricon:
so I guess we have 2 options: 
- keep a sandbox id / workspace path mapping at the Kaiden level (global file?)
- @feloy's idea of splitting the base64'ed path into multiple labels:
     ai.openkaiden.kaiden.workspace.0=\<first 63 chars>
     ai.openkaiden.kaiden.workspace.1=\<next 63 chars>
     
  I think @feloy's idea can work

fbricon:
Decoded path now added as a tooltip on the sandbox

Screenshot 2026-06-18 at 16 25 53

jeffmaury:
> so I guess we have 2 options:
> 
>     * keep a sandbox id / workspace path mapping at the Kaiden level (global file?)
> 
>     * @feloy's idea of splitting the base64'ed path into multiple labels:
>       ai.openkaiden.kaiden.workspace.0=<first 63 chars>
>       ai.openkaiden.kaiden.workspace.1=<next 63 chars>
>       I think @feloy's idea can work

Whart about having a Kaiden specific folder inside the VM that where we would store:
- either the mapping
- the workspace.json file ?

fbricon:
> Whart about having a Kaiden specific folder inside the VM that where we would store:
> 
> * either the mapping
> * the workspace.json file ?

but then you'd have to peek inside the sandbox, I don't know how costly that will be

jeffmaury (changes requested):
LGTM but WORKSPACE_LABEL is defined twice should be defined in a central place

jeffmaury (approved):

fbricon:
@coderabbitai resolve

PR #2227, opened 2026-06-19
feat(openshell): enable bind mounts for local gateways

### LINKED ISSUE #2254: Kaiden needs to wire bind mounts for Openshell

the gateway determins what config we need to set so we can have bind mounts so that we can have filesystem and home work 

maybe register cli gateway and go from there?

_No response_

### Additional context

_No response_

bmahabirbu:
## Summary
- detect the local OpenShell compute driver before auto-starting a gateway and pin it in a generated gateway config
- enable bind mounts in the generated Podman or Docker driver config so local gateways are ready for follow-up mount-based sandbox work
- add gateway tests covering driver selection, fallback behavior, generated config contents, and config cleanup

Fixes:  https://github.com/openkaiden/kaiden/issues/2254

Test plan

first stop any existing gateway using openshell
`openshell gateway remove (name default is openshell)`
cleanup the port
`kill $(lsof -t -i :17670)`

start kaiden via pnpm watch

kaiden should have written to local/share/kaiden/openshell-gateway/gateway.toml

which should show how enable_bind_mounts = true and compute_drivers = ["podman"] (or docker). This proves the template renders correctly with driver detection

Now theres a few things I ran into (maybe linux specific) that Kaiden couldnt get the gateway fully operational so heres a way to manually test the config works

Prerequisites: openshell 0.0.76+, podman installed, mTLS certs already provisioned (~/.local/state/openshell/tls/)

  1. Clean up any existing gateway
```
  openshell gateway remove kaiden-local
  kill $(lsof -t -i :17670)
  rm -f ~/.local/share/kaiden/openshell-gateway/gateway.toml
```

  2. Write the config (or let Kaiden generate it)
```
  mkdir -p ~/.local/share/kaiden/openshell-gateway
  cat > ~/.local/share/kaiden/openshell-gateway/gateway.toml << 'EOF'
  [openshell]
  version = 1
  
  [openshell.gateway]
  compute_drivers = ["podman"]

  [openshell.drivers.podman]
  enable_bind_mounts = true
  supervisor_image = "ghcr.io/nvidia/openshell/supervisor:0.0.76"
  EOF

```

  3. Start the gateway with the config
```
  openshell-gateway \
    --config ~/.local/share/kaiden/openshell-gateway/gateway.toml \
    --port 17670 \
    --bind-address 0.0.0.0 &

```
  ▎ Note (Linux only): --bind-address 0.0.0.0 is required because sandbox containers reach the host via host.containers.internal (169.254.1.2), not 
  ▎ 127.0.0.1. On macOS this isn't needed since podman runs in a VM with port forwarding.

  4. Register the gateway
`  openshell gateway add https://127.0.0.1:17670 --name kaiden-local --local`

  5. Create a sandbox with a bind mount
```
  mkdir -p /tmp/bind-test
  echo "hello from host" > /tmp/bind-test/test.txt
  openshell sandbox create --name e2e-bind \
    --driver-config-json '{"podman":{"mounts":[{"type":"bind","source":"/tmp/bind-test","target":"/tmp/bind-test","read_only":false}]}}'
```

  6. Verify the file is mounted inside the sandbox
`  openshell sandbox exec -n e2e-bind -- cat /tmp/bind-test/test.txt`
  # Expected output: hello from host

  Update the file on the host and verify that the change is immediately visible in the sandbox:
```
  echo "updated from host" > /tmp/bind-test/test.txt
  openshell sandbox exec -n e2e-bind -- cat /tmp/bind-test/test.txt
```
  # Expected output: updated from host

  7. Cleanup
`  openshell sandbox delete e2e-bind`

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
@@ -1,10 +1,22 @@
 [openshell]
 version = 1
 
+{{#driver}}
+[openshell.gateway]
+compute_drivers = ["{{{driver}}}"]
+
+[openshell.drivers.{{{driver}}}]
+enable_bind_mounts = true
+{{#supervisorImage}}
+supervisor_image = "{{{supervisorImage}}}"
+{{/supervisorImage}}
+{{/driver}}
+{{^driver}}
 [openshell.drivers.podman]
 {{#supervisorImage}}
 supervisor_image = "{{{supervisorImage}}}"
 {{/supervisorImage}}
+{{/driver}}
 
 [openshell.gateway.auth]
 allow_unauthenticated_users = true
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.ts
@@ -46,6 +46,8 @@ const STOP_TIMEOUT_MS = 5000;
 const SUPERVISOR_IMAGE_BASE = 'ghcr.io/nvidia/openshell/supervisor';
 const GATEWAY_LOG_FILENAME = 'gateway.log';
 
+type LocalComputeDriver = 'docker' | 'podman';
+
 /**
  * Manages the `openshell-gateway` server binary lifecycle.
  *
@@ -320,6 +322,8 @@ export class OpenshellGateway implements Disposable {
         }
       }
 
+      const driver = await this.detectLocalComputeDriver();
+
       const storageDirectory = join(this.directories.getDataDirectory(), 'openshell-gateway');
       const configPath = join(storageDirectory, 'gateway.toml');
       await this.generateCerts(binaryPath, storageDirectory);
@@ -327,6 +331,7 @@ export class OpenshellGateway implements Disposable {
         supervisorImage: image,
         gatewayDir: storageDirectory,
         q: '"',
+        driver,
       });
 
       await mkdir(storageDirectory, { recursive: true });
@@ -340,6 +345,25 @@ export class OpenshellGateway implements Disposable {
       const message = err instanceof Error ? err.message : String(err);
       console.warn(`[openshell-gateway] failed to generate gateway config: ${message}`);
       return undefined;
+    }
+  }
+
+  private async detectLocalComputeDriver(): Promise<LocalComputeDriver | undefined> {
+    if (await this.isCommandAvailable('podman')) {
+      return 'podman';
+    }
+    if (await this.isCommandAvailable('docker')) {
+      return 'docker';
+    }
+    return undefined;
+  }
+
+  private async isCommandAvailable(command: string): Promise<boolean> {
+    try {
+      await this.exec.exec(command, ['--version']);
+      return true;
+    } catch {
+      return false;
     }
   }
 
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.ts
@@ -349,21 +349,12 @@ export class OpenshellGateway implements Disposable {
   }
 
   private async detectLocalComputeDriver(): Promise<LocalComputeDriver | undefined> {
-    if (await this.isCommandAvailable('podman')) {
-      return 'podman';
-    }
-    if (await this.isCommandAvailable('docker')) {
-      return 'docker';
-    }
-    return undefined;
-  }
-
-  private async isCommandAvailable(command: string): Promise<boolean> {
     try {
-      await this.exec.exec(command, ['--version']);
-      return true;
+      const info = await this.openshellCli.getGatewayInfo();
+      const driver = info.compute_drivers[0]?.capabilities.driver_name;
+      return driver === 'podman' || driver === 'docker' ? driver : undefined;
     } catch {
-      return false;
+      return undefined;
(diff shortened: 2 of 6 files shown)


packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -235,6 +249,66 @@ export class OpenshellGateway implements Disposable {
...
+      return undefined;
+    }
+
+    const configPath = join(tmpdir(), `kaiden-openshell-gateway-${randomUUID()}.toml`);
+    await writeFile(configPath, this.buildGatewayConfig(driver), 'utf-8');
+    console.log(`[openshell-gateway] generated local gateway config for ${driver} at ${configPath}`);
+    return configPath;
+  }
+
+  private async detectLocalComputeDriver(): Promise<LocalComputeDriver | undefined> {
```
benoitf, line 351: I don't think kaiden should do detection of tools. I think we delegate that to the gateway.

benoitf (commented):
I think there is a problem to solve to start the gateway with right config file but I'm not sure that the approach taken is the more robust as we're duplicating some OpenShell gateway logic there

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Gotcha so in that case should we register the gateway as a clitool first? since we dont have that now and then we can check if the gateway driver is podman and start with the right config for mount use?

bmahabirbu:
@benoitf updated!

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -235,6 +247,41 @@ export class OpenshellGateway implements Disposable {
     return args;
   }
 
+  private async createGatewayConfig(): Promise<string | undefined> {
+    const configPath = join(tmpdir(), `kaiden-openshell-gateway-${randomUUID()}.toml`);
```
benoitf: I'm not sure why we need to use a tmpDir.

we've a storage folder where we hold kaiden stuff, should it be part of this ?

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -235,6 +247,41 @@ export class OpenshellGateway implements Disposable {
...
+    await writeFile(configPath, this.buildGatewayConfig(), 'utf-8');
+    console.log(`[openshell-gateway] generated local gateway config at ${configPath}`);
+    return configPath;
+  }
+
+  private buildGatewayConfig(): string {
+    return [
```
benoitf: it needs to use mustache for templates

(or here it seems it's a default file) so might be a separate .toml file that we include using raw vite import

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Understood! Ill use the storage folder

benoitf:
note: on my end I also need to setup TLS, etc so I don't see how this simple config file can work (at least on macOS)

were you able to start a sandbox using that gateway ?

bmahabirbu:
I was on linux but good catch I didnt test on mac

bmahabirbu:
The install script / brew shouldve handled the tsl for mac i dont think kaiden should need to do it?

for linux i did notice for podman you need 0.0.0.0 for the gateway ip for it to work which seems odd to me. Should be a upstream fix in openshell so will ignore it for docker the default gateway works fine

bmahabirbu:
@jeffmaury please take a look as well and your opinion on this! thank you

types/mustache.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'mustache' {
+  export function render(template: string, view: object): string;
+
+  const Mustache: {
+    render: typeof render;
+  };
+
+  export default Mustache;
+}
```
benoitf: need to use `@types/mustache`

benoitf (changes requested):
mustache types should not be part of types folder, there are `@types/` npmjs packages

packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
```diff
@@ -0,0 +1,9 @@
+[openshell]
```
benoitf, line 1: should have a .template suffix to know it's a mustache template

bmahabirbu:
@coderabbit resolve

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: update gateway now registered which is good

bmahabirbu:
@coderabbit resolve

MarsKubeX (approved):
LGTM. Tested removing the gateway and after pnpm watch, checking the values in gateway.toml.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Hmm seems 0.083 and future 0.0084 wont have the compute driver info as part of the gateway info argument sadly

benoitf (approved):

PR #2228, opened 2026-06-19
ci: single definition of the node version to use

serbangeorge-m:
Replace all hardcoded node-version: 24.15.0 with node-version-file: '.nvmrc' across workflow files


benoitf (approved):
thx

serbangeorge-m:
@coderabbitai resolve

PR #2240, opened 2026-06-22
fix(ci): testing build windows arch

serbangeorge-m:
Fixes the daily testing build workflow to produce correct Windows x64 and arm64 installers separately


MarsKubeX (approved):
LGTM

PR #2242, opened 2026-06-22
feat(models): wire model selection to semantic router create workflow

### LINKED ISSUE #2103: Wire model selection to the create semantic router workflow

We should be able to select one of more models from the catalog

Allow selection of models from the catalog. Only models which do not have llmMetadata.semanticRouter defined should be considered

From the mockups: https://github.com/openkaiden/mockups/blob/synchronise-sprint290-main/models/semantic-router-create.html

Image

_No response_

### Additional context

_No response_

MarsKubeX:
Add a multi-select model step to the semantic router creation wizard so users can pick backend models from the catalog. Models with llmMetadata.semanticRouter defined are filtered out. Extends ModelSelectionTable with multiSelect support to avoid duplicating the table rendering logic.

https://github.com/user-attachments/assets/03606671-3e51-4a8c-a6d6-d8cf06aa46f6

Closes #2103

### DIFF
--- a/packages/extension-api/src/extension-api.d.ts
+++ b/packages/extension-api/src/extension-api.d.ts
@@ -653,6 +653,7 @@ declare module '@openkaiden/api' {
 
   export interface LLMMetadata {
     name?: string;
+    semanticRouter?: string;
   }
 
   export type InferenceProviderConnection = {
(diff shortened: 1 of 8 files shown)


gastoner (approved):
LGTM

PR #2249, opened 2026-06-22
fix(openshell): pass workspace env vars to sandbox create

### LINKED ISSUE #2246: Wire environment variables to the OpenShell create workspace

When creating a workspace the environment variables defined in the workspace.json should be set to the sandbox

See the environment variables to the sandbox.

See https://github.com/openkaiden/kdn/blob/7fb21875d102a004a2b528487d01d558846bdf6e/pkg/runtime/openshell/create.go#L284 but use --env instead

_No response_

### Additional context

_No response_

bmahabirbu:
Fixes #2246.

## Summary
- pass workspace environment entries through the OpenShell workspace creation path
- add repeatable `--env KEY=VALUE` support to the OpenShell CLI wrapper used by Kaiden
- cover the new wiring with focused unit tests and a local OpenShell smoke test

## Test plan
- [x] `pnpm exec vitest run packages/main/src/plugin/openshell-cli/openshell-cli.spec.ts packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts`
- [x] `openshell sandbox create --name <temp> --from base --env KAIDEN_ENV_VERIFY=works --no-keep --no-tty -- sh -lc 'printf %s "$KAIDEN_ENV_VERIFY"'`

### DIFF
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -85,7 +85,7 @@ export interface CreateSandboxOptions {
   cpu?: string;
   memory?: string;
   providers?: string[];
-  env?: string[];
+  env?: Record<string, string>;
   labels?: Record<string, string>;
   uploads?: Array<{ local: string; remote: string }>;
   command?: string[];
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -85,6 +85,7 @@ export interface CreateSandboxOptions {
   cpu?: string;
   memory?: string;
   providers?: string[];
+  env?: string[];
   labels?: Record<string, string>;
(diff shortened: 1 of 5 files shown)


packages/api/src/openshell-gateway-info.ts
```diff
@@ -85,6 +85,7 @@ export interface CreateSandboxOptions {
   cpu?: string;
   memory?: string;
   providers?: string[];
+  env?: string[];
```
jeffmaury: **suggestion:** should be typed as Record<string, string>

bmahabirbu: updated! thanks for the suggesiton

MarsKubeX (approved):
LGTM. Only take a look to the codeRabbit comment regarding leaking secrets.

Issue #2129, opened 2026-06-10
When configuring an agent check connections


jeffmaury:
### Is your feature request related to a problem? Please describe

If there are no compatible models when configuring an agent (onboarding or coding agents) let user create a compatible connection

### Describe the solution you'd like

Offer the user to create a compatible connection based on the LLMMetadata field

MarsKubeX:
@jeffmaury should the connection tiles be removed? or maybe just filtered by the selected provider (claude code, codex, git hub copilot...)?

<img width="1047" height="424" alt="Image" src="https://github.com/user-attachments/assets/b3e13525-2bcb-453f-a80c-8192853c2257" />

jeffmaury:
Removed but if the model list is empty offer the same kind of list but filtered on only of the list is greated than 1. On the creation of the connection should be done in the flow and not through a navigation. HTH

PR #2250, opened 2026-06-23
fix(onboarding): recover compatible models in guided setup

### LINKED ISSUE #2241: Update onboarding
### Task content

As per the new design, the new onboarding should be:

- select a default agent
- once it is selected, select a model compatible with that agent
- if no models are avalailble for compatibily, create a compatible connection and then select the model

See #2129 as it is almost the same workflow

bmahabirbu:
Fixes #2241.

## Summary
- move guided setup recovery to the model step so users can stay in onboarding when an agent has no compatible models
- keep Claude and Claude Vertex setup inline in guided setup while leaving local runtime recovery for OpenCode and OpenClaw
- bind onboarding state through the guided-setup step chain and add focused tests for the new recovery behavior

https://github.com/user-attachments/assets/bef9a5b4-fd85-4274-b25b-5f88804ef785

## Test plan
- [x] `pnpm vitest run packages/renderer/src/lib/guided-setup/CodingAgentStep.spec.ts packages/renderer/src/lib/guided-setup/ModelStep.spec.ts packages/renderer/src/lib/guided-setup/GuidedSetup.spec.ts packages/renderer/src/lib/guided-setup/panels/ClaudePanel.spec.ts packages/renderer/src/lib/guided-setup/panels/ClaudeVertexPanel.spec.ts`
- [x] `ReadLints` on the edited guided-setup files
- [ ] manual onboarding smoke test in the app


MarsKubeX (commented):
- I find a bit confusing when you are in step 2 and does not have any compatible model and you are able to create a connection to get models: you click on "go to dashboard" when the action performed is "create a connection". IMO the create connection should have a dedicated button in step 2 to create connection and once created, listing models, all in step 2. So when the user clicks on go to dashboard he really goes to dashboard.
- Also I have a question, should we keep the "API KEY" section in step 1? could it be better to only let the agents types and check them in the next step?

WDYT @jeffmaury?

bmahabirbu:
Thanks for your feedback @MarsKubeX I agree with your thoughts

so step 1 will jsut pick the agent

then step 2 will pick the models if a connection was found if not prompt users to create a new connection load the models upon successful connection then continue to dashboard

in that case is this issue https://github.com/openkaiden/kaiden/issues/2129 blocking mine? since that specifically ties to having the user be able to create a compatible connection?

MarsKubeX:
> Thanks for your feedback @MarsKubeX I agree with your thoughts
> 
> so step 1 will jsut pick the agent
> 
> then step 2 will pick the models if a connection was found if not prompt users to create a new connection load the models upon successful connection then continue to dashboard
> 
> in that case is this issue #2129 blocking mine? since that specifically ties to having the user be able to create a compatible connection?

Hey @bmahabirbu. Yes I think issue #2129 is blocking for yours. I'm creating a shared utility to add it to each place needing a compatible connection. I'll keep you update when I do the PR for it.

MarsKubeX:
> > Thanks for your feedback @MarsKubeX I agree with your thoughts
> > so step 1 will jsut pick the agent
> > then step 2 will pick the models if a connection was found if not prompt users to create a new connection load the models upon successful connection then continue to dashboard
> > in that case is this issue #2129 blocking mine? since that specifically ties to having the user be able to create a compatible connection?
> 
> Hey @bmahabirbu. Yes I think issue #2129 is blocking for yours. I'm creating a shared utility to add it to each place needing a compatible connection. I'll keep you update when I do the PR for it.

added PR #2259 for the common utility.

jeffmaury (changes requested):
For me there is no need to have agent specific panels anymore

You should also rebase it on #2259 once it's merged

bmahabirbu:
@jeffmaury @MarsKubeX updated! Check the new video attached above

MarsKubeX:
> @jeffmaury @MarsKubeX updated! Check the new video attached above

LGTM. Thanks for the changes. I have just one question, since there is only one step, the stepper is still needed? We can maybe just change the stepper for a message. WDYT?

bmahabirbu:
@MarsKubeX I was thinking the same think but I believe the onboarding would have the auto discovery of mcp or skills? or other steps that's useful?

MarsKubeX:
> @MarsKubeX I was thinking the same think but I believe the onboarding would have the auto discovery of mcp or skills? or other steps that's useful?

you're right I forgot about those. LGTM then. Great work

MarsKubeX (approved):

packages/renderer/src/lib/mcp/MCPServerRemoteActions.svelte
```diff
@@ -1,16 +1,10 @@
 <script lang="ts">
```
jeffmaury, line 1: **issue:** not related to this PR

packages/renderer/src/lib/guided-setup/agent-registry.ts
```diff
@@ -28,30 +28,15 @@ import OpenCodeIcon from '/@/lib/images/agents/OpenCodeIcon.svelte';
 import VertexAIIcon from '/@/lib/images/agents/VertexAIIcon.svelte';
 
 import type { CliAgent } from './guided-setup-steps';
```
jeffmaury: **question:** can't this file be removed ? But maybe another PR

packages/renderer/src/lib/guided-setup/guided-setup-steps.ts
```diff
@@ -17,16 +17,28 @@
...
 
 import type { DefaultWorkspaceModelSettings, DefaultWorkspaceSettings } from '/@api/onboarding-settings-info';
 
-import CodingAgentStep from './CodingAgentStep.svelte';
-import ModelStep from './ModelStep.svelte';
+import GuidedSetupPage from './GuidedSetupPage.svelte';
 
 export type CliAgent = 'opencode' | 'openclaw' | 'claude' | 'claude-vertex' | 'cursor' | 'goose';
 
+const CLI_AGENTS: ReadonlySet<string> = new Set<CliAgent>([
```
jeffmaury: **issue:** should use the existing store

jeffmaury (changes requested):
LGTM feature wise but still some issues remaining related to agent discovery but maybe another PR.

The only required change is the one about unrelated change

bmahabirbu:
@coderabbit resolve

jeffmaury (approved):
LGTM

PR #2259, opened 2026-06-24
feat(renderer): offer inline connection creation when no compatible models exist

### LINKED ISSUE #2129: When configuring an agent check connections

MarsKubeX:
When an agent has no compatible models available, the UI now shows compatible unconfigured providers as selectable buttons and renders the connection creation form inline — removing the need to navigate away to Preferences. ProviderConnectionTiles removed from agent detail.

https://github.com/user-attachments/assets/49e5936b-c1c3-47c6-b69d-63f058fd285c

https://github.com/user-attachments/assets/083e1475-f0d1-4fcd-8603-08de33045b94

Closes #2129


jeffmaury (approved):
LGTM. But I would expect that if there is single provider we don't have to select it but display the connection creation page

Approving because it's not a blocker for me

Nice work creating a new shared componet

PR #2271, opened 2026-06-26
fix(openshell): upload all filesystem mounts including broad paths

### LINKED ISSUE #2253: Wire filesystem to the OpenShell create workspace

We should see the files in the sandbox

Apply infoirmation for the filesystem part of the workspace.json and uploads them to the sandbox

_No response_

### Additional context

_No response_

bmahabirbu:
## Summary

- Removes the `isBroadOpenshellHostPath` filter that was silently skipping `$HOME` and `/` mounts during OpenShell sandbox creation
- All workspace file access options (Home Directory, Full System Access, Custom Paths) now work correctly with OpenShell via `--upload`
- OpenShell bind mounts are not reliably supported yet

Fixes #2253

  Setup

  1. Have a project directory with some files (e.g. ~/projects/test-app)
  2. Launch Kaiden, create a workspace pointing at that directory
  
  ## Manual Test Plan

  ### Setup
  - Have a project directory with some files (e.g. `~/projects/test-app`)
  - Launch Kaiden, select Claude as the agent

  ### Test Cases

  - [ ] **Source directory uploads** — Create workspace with no custom mounts. Verify: `openshell sandbox exec -- ls -la` shows your project files in the
  working directory.
  - [ ] **Custom mount with $HOME path** — Create workspace with custom mount `$HOME/Downloads` → `$HOME/Downloads`. Verify: `openshell sandbox exec -- ls
  -la '~/Downloads'` shows your Downloads contents.
  - [ ] **Custom mount with $SOURCES subdir** — Create workspace with custom mount `$SOURCES/src` → `$SOURCES/src`. Verify: `openshell sandbox exec -- ls
  src/` shows your project's src files.
  - [ ] **Skills upload** — Create workspace with skills selected. Verify: `openshell sandbox exec -- ls .claude/skills/` shows the selected skill
  directories.
  - [ ] **Claude config uploaded** — Verify: `openshell sandbox exec -- cat .claude.json` shows onboarding completed and trust accepted.
  
  

https://github.com/user-attachments/assets/50a3e96c-1aa2-4186-b991-776b36a3d45e

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -16,9 +16,9 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { access, readFile, rm, writeFile } from 'node:fs/promises';
+import { access, lstat, readFile, rm, writeFile } from 'node:fs/promises';
 import { homedir, tmpdir } from 'node:os';
-import { basename, join, posix } from 'node:path';
+import { basename, isAbsolute, join, posix, resolve } from 'node:path';
 
 import type { Disposable, FileSystemWatcher } from '@openkaiden/api';
 import type { WebContents } from 'electron';
@@ -52,6 +52,10 @@ import { AGENT_LABEL, decodeWorkspaceLabels, WORKSPACE_LABEL } from '/@api/opens
 
 const HOME_VARIABLE = '${HOME}';
 const LABEL_MAX_LENGTH = 63;
+const SOURCES_VARIABLE = '$SOURCES';
+const MOUNT_HOME_PREFIX = '$HOME';
+
+type OpenshellUpload = { local: string; remote: string };
 
 export function encodeWorkspaceLabels(sourcePath: string): Record<string, string> {
   const encoded = Buffer.from(sourcePath).toString('base64url');
@@ -138,7 +142,7 @@ export class AgentWorkspaceManager implements Disposable {
 
     const workspace = await writeWorkspaceConfig(options);
     const agent = this.agentRegistry.getAgentRegistration(options.agent);
-    const uploads: Array<{ local: string; remote: string }> = [];
+    const uploads: OpenshellUpload[] = [];
 
     if (agent) {
       const writable = await Promise.all(
@@ -184,6 +188,8 @@ export class AgentWorkspaceManager implements Disposable {
       }
     }
 
+    uploads.push(...(await this.buildOpenshellFilesystemUploads(options.sourcePath, workspace)));
+
     const sandboxName = options.name ?? basename(options.sourcePath);
     const env = workspace.environment
       ?.filter(entry => typeof entry.value === 'string' && entry.value !== '')
@@ -191,6 +197,7 @@ export class AgentWorkspaceManager implements Disposable {
         acc[entry.name] = entry.value as string;
         return acc;
       }, {});
+    const dedupedUploads = this.dedupeOpenshellUploads(uploads);
 
     const t0 = performance.now();
 
@@ -207,7 +214,7 @@ export class AgentWorkspaceManager implements Disposable {
         providers: options.secrets,
         env: env && Object.keys(env).length > 0 ? env : undefined,
         labels: { ...encodeWorkspaceLabels(options.sourcePath), [AGENT_LABEL]: options.agent },
-        uploads: uploads.length > 0 ? uploads : undefined,
+        uploads: dedupedUploads.length > 0 ? dedupedUploads : undefined,
         noTty: true,
         command: ['true'],
         policy: policyFile,
@@ -242,10 +249,7 @@ export class AgentWorkspaceManager implements Disposable {
     }
   }
 
-  private buildOpenshellSkillUploads(
-    skills: string[] | undefined,
-    destinationSkillsFolder: string,
-  ): Array<{ local: string; remote: string }> {
+  private buildOpenshellSkillUploads(skills: string[] | undefined, destinationSkillsFolder: string): OpenshellUpload[] {
     if (!skills?.length) {
       return [];
     }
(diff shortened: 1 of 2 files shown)


jeffmaury (changes requested):
Failed. You should upload the sources files to the current (use . as the target) as hard coding paths is error prone

Failed while testing

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -253,6 +259,137 @@ export class AgentWorkspaceManager implements Disposable {
...
   }
 
+  private buildOpenshellFilesystemUploads(
+    sourcePath: string,
+    workspace: AgentWorkspaceConfiguration,
+  ): OpenshellUpload[] {
+    const sandboxSourcesPath = this.resolveOpenshellSourcesPath(sourcePath);
+    const uploads: OpenshellUpload[] = [
+      { local: sourcePath, remote: sandboxSourcesPath },
+      { local: join(sourcePath, '.kaiden', 'workspace.json'), remote: '.kaiden/workspace.json' },
```
jeffmaury: **praise:** I don't understand this

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -253,6 +259,137 @@ export class AgentWorkspaceManager implements Disposable {
...
+      const remote = this.resolveOpenshellSandboxPath(mount.target, sourcePath);
+      if (!local || !remote) {
+        continue;
+      }
+      uploads.push({ local, remote });
+    }
+    return uploads;
+  }
+
+  private resolveOpenshellHostPath(path: string, sourcePath: string): string | undefined {
```
jeffmaury: **suggestion:** should be rename as resolveHostPath as it is confusing

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -53,6 +53,11 @@ import type { SecretValue } from '/@api/secret-info.js';
 
 const HOME_VARIABLE = '${HOME}';
 const LABEL_MAX_LENGTH = 63;
+const SOURCES_VARIABLE = '$SOURCES';
+const MOUNT_HOME_VARIABLE = '$HOME';
+const OPENSHELL_SANDBOX_HOME = '/home/agent';
```
jeffmaury: **issue:** does not exists for me.

Don't hardcode this value and use $HOME or  ~ as they are supported by OpenShell

jeffmaury (commented):
I also think it should share some code with the skills upload (target destination)

bmahabirbu:
Tested on linux need to give it a shot on macos to confirm

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: yeah that should be removed we dont need the workspace.json will update now

bmahabirbu:
I found some issues while testing, using $Home (ie mounting the entire home directory) would unfortunately crash because openshell couldn't upload files it deemed not compatible. 

so if openshell has a req for files then this feature might be really flaky and we need bind mounts @jeffmaury  WDYT?

The og test I did was on a computer with barely any stuff in it so it makes sense it passed

benoitf:
I'm not sure I would anyway use the HOME folder as upload as it'll corrupt my sandbox trying to upload so many GB

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -258,21 +265,135 @@ export class AgentWorkspaceManager implements Disposable {
...
+    }
+    return undefined;
+  }
+
+  private resolveOpenshellSandboxPath(path: string, _sourcePath: string): string | undefined {
+    if (path === SOURCES_VARIABLE) {
+      return '.';
+    }
+    if (path.startsWith(`${SOURCES_VARIABLE}/`)) {
+      const resolved = posix.normalize(posix.join('.', `.${path.slice(SOURCES_VARIABLE.length)}`));
```
jeffmaury: **question:** why is there a dot in the second expression ?

so if I have $SOURCES/toto it will get resolved to ././toto looks weird to me

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
jeffmaury: **issue:** acked but not processed

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: this was satisfying the coderabbit suggestions https://github.com/openkaiden/kaiden/pull/2271#discussion_r3493752909 which normalized $SOURCES/foo/../bar into bar so the containment check works

It does look weird it can be simplified to make it more readable

bmahabirbu: You can test via `openshell sandbox create --name test-upload-escape --upload /tmp/test-upload:../escape --no-keep -- ls -la /escape`
 ```
      Error: Exit code 1

     Created sandbox: test-upload-escape

       [0.0s] Requesting compute...
       • Uploading files to ../escape...
     mkdir: cannot create directory '../escape': Permission denied
     Error:   × ssh tar extract exited with status exit status: 1
```
      
If the sandbox then has root permissions it could access it. That said I wonder if we should just delegate to openshell for safety concerns like this. It seems to catch the error with user permission in the sandbox so kaiden here doesnt need to check for path escapes for security then
      
will remove

File: packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts [hunk 5]
```diff
@@ -251,6 +258,106 @@ export class AgentWorkspaceManager implements Disposable {
...
+      return sourcePath;
+    }
+    if (path.startsWith(`${SOURCES_VARIABLE}/`)) {
+      return resolve(sourcePath, `.${path.slice(SOURCES_VARIABLE.length)}`);
+    }
+    if (path === MOUNT_HOME_VARIABLE) {
+      return homedir();
+    }
+    if (path.startsWith(`${MOUNT_HOME_VARIABLE}/`)) {
+      return resolve(homedir(), `.${path.slice(MOUNT_HOME_VARIABLE.length)}`);
```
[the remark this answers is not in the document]
bmahabirbu: wouldnt kaiden not be the orchestator for this type of security? lets say a user tries this on the cli side using openshell then if the sandbox user has root then it would work.

so kaiden doesnt really control the path security then. If kaiden manages this openshell sandbox creation we should delegate to upstream openshell for the fix?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: cause we will move away from --upload eventually so the logic wont persist

bmahabirbu:
@coderabbit resolve

bmahabirbu:
@jeffmaury ready-to-test

jeffmaury (approved):

MarsKubeX (approved):
LGTM

PR #2288, opened 2026-06-29
feat(agent-workspace): encode agent label on sandbox creation

### LINKED ISSUE #2286: When creating label add label for agentId
### Task content

When creating an agent, add a new label ai.openkaiden.kaiden.agent to store the agentId used

gastoner:
Generalize encodeWorkspaceLabels/decodeWorkspaceLabels into encodeLabel/decodeLabel that accept an arbitrary label key. Add AGENT_LABEL constant and attach it alongside the workspace label when creating OpenShell sandboxes. Update getAgentId to decode the base64url-encoded agent label instead of reading it raw.

Closes https://github.com/openkaiden/kaiden/issues/2286

### DIFF
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -43,6 +43,7 @@ export type SandboxInfo = z.output<typeof SandboxInfoSchema> & {
 };
 
 export const WORKSPACE_LABEL = 'ai.openkaiden.kaiden.workspace';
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
In most of the time the agent id would fit on a single label and it would be more user friendly to be stored directly in the label rather than encoded, can we have that option for this label ?

Codewise, LGTM

jeffmaury (approved):
LGTM

PR #2300, opened 2026-06-29
fix: check stdout/stderr and augment error message in extractCliError

### LINKED ISSUE #2285: Improve extractCliError
### Task content

Sometimes error information is missing as stdout and stderr are not returned in certain use case

See https://github.com/openkaiden/kaiden/blob/f59e648f57ea7026cd6164246e42417e657bd60f/packages/main/src/plugin/openshell-cli/openshell-cli.ts#L82

jeffmaury:
## Summary
- Extract JSON error parsing into a reusable `tryExtractJsonError` helper to avoid duplication
- Extend `extractCliError` to inspect both `stdout` and `stderr` for JSON error payloads
- When JSON parsing fails, augment `err.message` with raw `stderr` or `stdout` content so CLI error details are no longer silently discarded

Fixes #2285

## Test plan
- [x] Existing tests pass (76 total)
- [x] New test: `stderr` contains JSON `{"error": "..."}` → returns that error string
- [x] New test: both `stdout` and `stderr` have JSON errors → `stdout` wins
- [x] New test: raw `stderr` augments `err.message` when not JSON
- [x] New test: raw `stdout` augments `err.message` when `stderr` is empty

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-cli.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-cli.spec.ts
@@ -318,6 +318,54 @@ describe('createSandbox', () => {
 
     await expect(openshellCli.createSandbox()).rejects.toThrow('gateway connection refused');
   });
+
+  test('extracts JSON error from stderr on failure', async () => {
+    vi.spyOn(console, 'log').mockImplementation(() => undefined);
+    vi.spyOn(console, 'error').mockImplementation(() => undefined);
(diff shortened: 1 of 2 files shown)


benoitf (approved):
looks like zod could be used to check if there an error field ? rather than all the if

note that there is an issue on OpenShell side where errors are not always wrapped into the JSON but in raw output (breaking JSON formatting)

PR #2308, opened 2026-06-30
feat(secrets): tie openshell provider lifecycle to inference connections

### LINKED ISSUE #2303: Providers should be related to inference connections

When a workspace is created, a provider is created from the inference connection deduced from the model. But as it is never deleted, when the same workspace is created, then it leads to an error

We should now create and delete the providers when a new inference connection is created or deleted. The secret is created if the information is present in the configuration. The secret should be called $type-$id where the type is the provider type given by the configuration property _type and id is the connection id. The secret would also have a label ai.openkaiden.kaiden.inferenceConnectionId with value $providerId-$i where $providerId is the provider is that created the connection and id is the connection id. 
In create workspace we would only look for a secret linked to the connection and add it to the workspace if it exists

_No response_

### Additional context

_No response_

jeffmaury:
## Summary

- Move openshell provider create/delete from workspace creation into `SecretManager`, driven by inference connection register/unregister events
- Workspace creation now looks up the existing provider by name (`providerId-connectionId`) instead of creating one on the fly, eliminating the duplicate-provider error when recreating a workspace
- Fix bug: wrong emitter in `onDidUnregisterInferenceConnectionCallback` (was firing Kubernetes instead of inference)
- Reorder extension lifecycle: set config before registering the connection so the centralized listener has configuration available when the event fires
- Enrich register/unregister events with the `connection` object
- `ProviderRegistry.getInferenceConnection` now returns `providerId` instead of `extensionId`

## Test plan

- [x] TypeScript compiles with no errors across all modified packages
- [x] All 140 tests pass in `secret-manager/` and `agent-workspace/`
- [x] All 228 tests pass including `provider-registry.spec.ts`
- [ ] Manual: create an inference connection, verify openshell provider is created. Delete connection, verify provider is deleted. Create workspace, verify it attaches the existing provider. Recreate workspace, verify no duplicate error.

Fixes #2303

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -114,8 +114,8 @@ export class AgentWorkspaceManager implements Disposable {
         await rm(configPath, { force: true });
       }
 
-      const credentialsEnvironment = await this.ensureModelSecret(options);
-      const workspaceId = await this.createOpenshell(options, credentialsEnvironment);
+      const secretName = await this.ensureModelSecret(options);
+      const workspaceId = await this.createOpenshell(options, secretName);
       this.apiSender.send('agent-workspace-update');
       task.status = 'success';
       return workspaceId;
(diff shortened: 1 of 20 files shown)


packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -135,25 +131,18 @@ export class AgentWorkspaceManager implements Disposable {
 
   private async createOpenshell(
     options: AgentWorkspaceCreateOptions,
-    credentialsEnvironment: Record<string, string>,
+    secretName: string | undefined,
```
benoitf:
```suggestion
    secretName?: string,
```

benoitf (approved):
LGTM

MarsKubeX (commented):
LGTM overall. Just check the codeRabbit comments regarding ``startAsIs()`` and the missing ``removeConnection(id)``.

Nit: update the stale doc comment of  ``ensureModelSecret()``in agent-workspace-manager.ts since it describes de old behavior ("create a matching kdn vault secret")

jeffmaury:
@coderabbitai resolve

PR #2310, opened 2026-06-30
feat: configurable openshell binary resolution priority

### LINKED ISSUE #2309: fix: bundled openshell binary should take priority over system PATH
When running from an Electron Builder packaged binary, the openshell CLI is resolved from the system PATH (e.g. `/usr/local/bin/openshell`) instead of the bundled copy at `process.resourcesPath/openshell/openshell`.

The system copy may be outdated and not support flags like `-o json`, causing errors such as:
```
error: unexpected argument '-o' found
```

**Root cause:** In `discoverBinary()` (`extensions/openshell/src/manager/openshell-cli-manager.ts`), the system PATH is checked **before** the bundled resources directory. Any system-installed openshell — regardless of version — shadows the embedded binary.

**Fix:** Swap the resolution order so bundled resources are checked before system PATH:
1. Custom config path (unchanged)
2. Extension storage directory (unchanged)
3. Bundled resources (`process.resourcesPath`) — moved up
4. System PATH — moved down as fallback

benoitf:
Fixes #2309

## Summary
- Fix bundled openshell binary being shadowed by an outdated system-installed version in packaged Electron builds
- Add `openshell.binary.resolution` setting to control the full discovery order between bundled resources, extension storage, and system PATH (6 permutations available as a dropdown in Settings > OpenShell)
- Default order: `bundled,storage,system` — custom config path is always checked first regardless
- Remove debug console traces from `openshell-cli.ts`

## Test plan
- [x] `cd extensions/openshell && pnpm test` — all 26 tests pass
- [x] Verified default order prefers bundled over storage and system PATH
- [x] Verified `system,bundled,storage` order prefers system PATH over bundled
- [x] Verified `storage,bundled,system` order prefers extension storage over bundled
- [x] Verified custom config path always wins regardless of resolution setting

### DIFF
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -24,6 +24,16 @@
           "default": "",
           "description": "Custom path to OpenShell binary (Default is blank)"
         },
+        "openshell.binary.resolution": {
+          "type": "string",
+          "enum": [
+            "bundled",
+            "system"
+          ],
+          "scope": "DEFAULT",
+          "default": "bundled",
+          "description": "Binary resolution priority: 'bundled' prefers the embedded binary over system PATH (default); 'system' prefers binaries found in system PATH."
+        },
         "openshell.imageBuilder.binary.path": {
           "type": "string",
           "format": "file",
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -27,12 +27,16 @@
         "openshell.binary.resolution": {
           "type": "string",
           "enum": [
-            "bundled",
-            "system"
+            "bundled,storage,system",
+            "bundled,system,storage",
+            "storage,bundled,system",
+            "storage,system,bundled",
+            "system,bundled,storage",
+            "system,storage,bundled"
           ],
           "scope": "DEFAULT",
-          "default": "bundled",
-          "description": "Binary resolution priority: 'bundled' prefers the embedded binary over system PATH (default); 'system' prefers binaries found in system PATH."
+          "default": "bundled,storage,system",
+          "description": "Binary resolution priority order. Custom config path is always checked first regardless of this setting."
         },
         "openshell.imageBuilder.binary.path": {
           "type": "string",
--- a/packages/main/src/plugin/openshell-cli/openshell-cli.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-cli.ts
@@ -75,12 +75,8 @@ export class OpenshellCli {
   ) {}
 
   getCliPath(): string {
-    console.log('[openshell] getting CLI path');
-
     const tool = this.cliToolRegistry.getCliToolInfos().find(t => t.name === 'openshell');
-    console.log('[openshell] found CLI tool info:', tool);
     if (tool?.path) {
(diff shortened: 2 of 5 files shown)


jeffmaury (approved):

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -389,7 +407,9 @@ export class OpenshellCli {
   }
 
   private async execCLI<T>(args: string[], options?: RunOptions): Promise<T> {
+    console.log(`[openshell] executing CLI command: ${args.join(' ')}`);
```
fbricon: wouldn't that log sensitive data that should be obfuscated?

fbricon (changes requested):

jeffmaury (approved):

packages/main/src/plugin/openshell-cli/openshell-cli.ts
benoitf: sure, it was some traces when it was in draft mode

now it's gone

benoitf:
updated to use preferences as suggested there https://github.com/openkaiden/kaiden/issues/2309#issuecomment-4845753809 by @fbricon 

now Kaiden prefer the binaries embedded into Kaiden but it's possible to say to use the one from your system

extensions/openshell/package.json
```diff
@@ -24,6 +24,20 @@
...
+            "bundled,system,storage",
+            "storage,bundled,system",
+            "storage,system,bundled",
+            "system,bundled,storage",
+            "system,storage,bundled"
+          ],
+          "scope": "DEFAULT",
+          "default": "bundled,storage,system",
+          "description": "Binary resolution priority order. Custom config path is always checked first regardless of this setting."
+        },
```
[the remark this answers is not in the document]
benoitf: yes it's to avoid any issue with user's installed binaries by default

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -76,6 +79,15 @@ export class OpenshellCli {
     if (tool?.path) {
       return tool.path;
     }
+
+    const resourcesPath = (process as NodeJS.Process & { resourcesPath?: string }).resourcesPath;
+    if (resourcesPath) {
+      const bundledPath = join(resourcesPath, 'openshell', 'openshell');
```
MarsKubeX, line 85: Question: the hardcoded 'openshell' is the binary name? if this is the case, should it be 'openshell.exe' for windows as it is in the openshell-cli-manager.ts in ``private async discoverBinary()``

benoitf: there is no openshell.exe CLI from OpenShell yet

https://github.com/NVIDIA/OpenShell/releases/tag/v0.0.73

benoitf: also it was returning before `return 'openshell';` so the semantic is kept

fbricon (approved):

MarsKubeX (approved):

PR #2317, opened 2026-07-01
feat(openshell): use bundled binaries with VM driver and setup TLS

bmahabirbu:
## Summary
- Switch OpenShell gateway to use the VM compute driver (libkrun micro-VMs) with proper mTLS instead of running with TLS disabled
- Bootstrap mTLS certificates via `generate-certs` before first gateway start, spawn gateway with `--config gateway.toml` (VM driver, TLS enabled), and register as "kaiden" over `https://127.0.0.1:17670`
- Remove `openshell-sandbox` from Linux download maps since it is now auto-pulled as an OCI image

## Test plan
- [ ] Verify gateway starts with VM driver on Linux
- [ ] Verify gateway starts with VM driver on macOS
- [ ] Verify mTLS certificates are generated on first run
- [ ] Verify gateway registers as "kaiden" over https
- [ ] Verify sandbox binary is no longer downloaded (pulled as OCI image instead)
- [ ] Run `pnpm test:unit packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts`
- [ ] Run `pnpm test:unit extensions/openshell/src/openshell-download.spec.ts`

### DIFF
--- a/extensions/openshell/src/openshell-download.spec.ts
+++ b/extensions/openshell/src/openshell-download.spec.ts
@@ -121,19 +121,18 @@ describe('downloadOpenshellBinaries', () => {
       const cwd = opts.cwd ?? '';
       fileMap.set(normPath(path.join(cwd, 'openshell')), true);
       fileMap.set(normPath(path.join(cwd, 'openshell-gateway')), true);
-      fileMap.set(normPath(path.join(cwd, 'openshell-sandbox')), true);
       fileMap.set(normPath(path.join(cwd, 'openshell-driver-vm')), true);
     });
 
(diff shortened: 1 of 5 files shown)


bmahabirbu:
Havent tested yet just getting this one out before I log off but the idea is to keep the bundling simple. use native cli gateway + vm driver where possible dont rely on podman and setup tls and other stuff for eventual bind mounting

bmahabirbu:
from my understanding the openshell-sandbox binary isnt needed its automatically bundled inside each sandbox and kaiden doesnt really need to interact with it

jeffmaury:
Switching to draft as it comes from nowhere

PR #2338, opened 2026-07-02
fix(openshell): download binaries only for host platform during build

### LINKED ISSUE #2305: openshell download scripts downloads for all arches/OS
### Task content

when using `pnpm build` in `extensions/openshell` folder

it's downloading binaries for all platforms/arches

I expect it should download it only for the current OS

bmahabirbu:
## Summary

- Removes `--all` flag from the openshell extension `build` script so that `pnpm build` only downloads binaries for the current host OS/arch instead of all supported platforms
- The dedicated `download:all` and `download:image-builder:all` scripts remain available for explicit cross-platform downloads

Closes #2305

## Test plan

- [ ] Run `pnpm build` and check in `extensions/openshell` that your platform either linux or mac is there
- [ ] Run `pnpm run download:all` — verify all platform directories are created

### DIFF
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -57,7 +57,7 @@
     }
   },
   "scripts": {
-    "build": "vite build && node ./scripts/build.js && tsx ./scripts/download.ts --all && tsx ./scripts/download-image-builder.ts --all",
+    "build": "vite build && node ./scripts/build.js && tsx ./scripts/download.ts && tsx ./scripts/download-image-builder.ts",
     "download": "tsx ./scripts/download.ts",
     "download:all": "tsx ./scripts/download.ts --all",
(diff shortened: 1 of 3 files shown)


bmahabirbu:
the ci is per platform so it shouldnt effect how the workflow release is done

jeffmaury (commented):
Not sure the issue is valid as the electron build is failing @benoitf WDYT ?

bmahabirbu:
Will update so for the script = download all for a given OS (all arches) but then don't download windows or macOS on Linux

bmahabirbu:
@coderabbit resolve

bmahabirbu:
@coderabbit resolve

benoitf (approved):

PR #2344, opened 2026-07-02
chore: ensure unbinding inversify container on closing the Application

benoitf:
with e2e tests it was failing, maybe because the gateway is not stopped.

there is a `@preDestroy` hook but it was never called

note: I'm backporting the fix on Podman Desktop as well


jeffmaury (approved):

benoitf:
draft until it's reviewed upstream

PR #2346, opened 2026-07-02
feat(openshell-gateway): pin supervisor container image to gateway binary version

### LINKED ISSUE #2315: Gateway should pin supervisor image to its own version instead of :latest
## Summary

When running `openshell-gateway` v0.0.72, the default `supervisor_image` is `ghcr.io/nvidia/openshell/supervisor:latest`. If the supervisor image is updated to a newer version (e.g. v0.0.73), the gateway still creates containers with its v0.0.72 configuration but the v0.0.73 supervisor binary runs inside them. This version mismatch causes the sandbox to crash on startup.

## Steps to reproduce

1. Install openshell v0.0.72 (gateway + CLI)
2. Wait for `ghcr.io/nvidia/openshell/supervisor:latest` to be updated to v0.0.73
3. Run `openshell sandbox create --from base`

## Observed behavior

The sandbox container immediately exits with code 1:

```
2026-06-30T19:14:24.516Z WARN openshell_supervisor_network::proxy: host.openshell.internal maps to a non-link-local IP; trusted-gateway SSRF exemption disabled
2026-06-30T19:14:24.533Z WARN openshell_supervisor_process::netns: Failed to delete network namespace
Error:   × Invalid argument (os error 22)
```

The gateway logs show repeated restart failures:

```
[gateway] [WARN ] Container exited with code 1
```

## Workaround

Pin the supervisor image in `gateway.toml` to match the gateway version:

```toml
[openshell.drivers.podman]
supervisor_image = "ghcr.io/nvidia/openshell/supervisor:0.0.72"
```

## Suggested fix

`DEFAULT_SUPERVISOR_IMAGE` in `crates/openshell-core/src/config.rs` currently uses `:latest`:

```rust
pub const DEFAULT_SUPERVISOR_IMAGE: &str = "ghcr.io/nvidia/openshell/supervisor:latest";
```

The gateway should default to a tag that matches its own version (e.g. `ghcr.io/nvidia/openshell/supervisor:0.0.72` when the gateway is v0.0.72) so that the supervisor and gateway are always compatible without manual configuration.

## Environment

- macOS (darwin/arm64)
- Podman 6.1.0-dev with libkrun VM
- openshell CLI / gateway: v0.0.72
- supervisor image (`latest`): v0.0.73

gastoner:
Write a TOML config file that pins the supervisor_image to the gateway binary's own semver, preventing :latest tag drift. When version detection fails, the gateway starts without --config so behavior is unchanged.

Also adds a supervisorImage option to allow explicit image overrides.

image
You should see this line +- and by using `cat` on the config path yous should see output: 

```
[openshell.drivers.podman]
supervisor_image = "ghcr.io/nvidia/openshell/supervisor:0.0.72"
```

Closes https://github.com/openkaiden/kaiden/issues/2315

### DIFF
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "@types/mime-types": "^3.0.1",
     "@types/minimist": "^1.2.5",
     "@types/ms": "^2.1.0",
+    "@types/mustache": "^4.2.6",
     "@types/node": "^24",
     "@types/tar": "^6.1.13",
     "@types/tar-fs": "^2.0.4",
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -109,6 +109,8 @@ export interface OpenshellGatewayStartOptions {
   disableTls?: boolean;
   /** Skip CLI registration when restarting an already-registered gateway. */
   skipRegistration?: boolean;
+  /** Override the supervisor container image. When unset, the image tag is pinned to the gateway binary version. */
+  supervisorImage?: string;
 }
 
 export interface GatewaySandboxes {
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
@@ -18,6 +18,8 @@
 
 import type { ChildProcess } from 'node:child_process';
 import { EventEmitter } from 'node:events';
+import { writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
 
 import type { RunResult } from '@openkaiden/api';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
@@ -37,8 +39,6 @@ vi.mock(import('node:os'));
 vi.mock(import('/@/plugin/util/exec.js'));
 
 const { spawn } = await import('node:child_process');
-const { writeFile } = await import('node:fs/promises');
-const { tmpdir } = await import('node:os');
 
 const GATEWAY_BINARY = '/usr/local/bin/openshell-gateway';
 const CLI_BINARY = '/usr/local/bin/openshell';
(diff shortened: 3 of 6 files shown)


jeffmaury (approved):

benoitf:
might overlap with https://github.com/openkaiden/kaiden/pull/2227

in this related PR, it's using a mustache template file so it looks more extensible approach to add new entries to the toml file

I also filed an upstream PR in OpenShell so I don't know

bmahabirbu:
@benoit if this pr goes in `OpenShell PR 2070` then would this fix that Evzen has still be needed?

I was taking a look at the config creation and for 2227 which has the mustache template architecture inline it could be worth getting that in since bind moutning will be a no op as we are using --upload for 0.3?  then have this fix ontop

@jeffmaury  WDYT?

benoitf:
> @benoit if this pr goes in OpenShell PR 2070 then would this fix that Evzen has still be needed?

no but it's still unapproved :-/

jeffmaury:
@benoitf and @bmahabirbu this one is targeted for 0.3 and bind mounts are for 0.4

benoitf:
yes I'm just saying it should use mustache approach there (as the one for bind mount)
I'm still hoping also fix could land upstream (so we don't need this one)

bmahabirbu:
This pr could just copy the mustache template implementation from 2227 

It wouldn't make sense to merge this as it is if we would have to rewrite it later on 

(Also 2227 doesn't wire the bind mounts up to the workspace creation it just enables it via a config for the gateway so its a no op since my other 0.3 pr will use --upload for mounting)

gastoner:
@jeffmaury @benoitf @bmahabirbu moved to mustache template

benoitf (approved):
LGTM but it contains stuff for bind mount in the mustache template which seems unsalted to this PR

gastoner:
Removed bind mounts

jeffmaury (approved):
LGTM

gastoner:
@coderabbitai resolve

PR #2354, opened 2026-07-02
perf(workspaces): batch policies into sandbox create --policy YAML

fbricon:
- **chore: add timing logs to openshell workspace creation**
- **perf(workspaces): batch policies into sandbox create --policy YAML**
    Build a single policy YAML file combining network and model rules,
    then pass it via --policy to openshell sandbox create instead of
    running sequential openshell policy update CLI calls.

    This replaces 5-7 sequential CLI invocations (~49s) with a single
    call, reducing workspace creation time from ~54s to ~4s.


jeffmaury (approved):
LGTM but:

- no related issue
- we should remove logging before merging

benoitf (approved):

PR #2355, opened 2026-07-02
fix(agent-workspace): make model mandatory in create workspace

### LINKED ISSUE #2326: Make model mandatory in create workspace
### Task content

See https://github.com/openkaiden/kaiden/blob/bd998a5aa5e55c3bfd610dcdae041d9748666de8/packages/api/src/agent-workspace-info.ts#L67

The model is declared as optional but rejected by the backend

Change the requirement to mandatory and report impacts on the UI (mainly when create is execute from the first screen). It should be rejected if no model is selected

MarsKubeX:
The backend rejects workspace creation without a model, but the type declared it optional. Make `model` required in `AgentWorkspaceCreateOptions`, remove the redundant runtime guard, and disable all creation buttons in the UI when no model is selected.

Closes #2326

### DIFF
--- a/packages/api/src/agent-workspace-info.ts
+++ b/packages/api/src/agent-workspace-info.ts
@@ -64,7 +64,7 @@ export type AgentWorkspaceMount = configComponents['schemas']['Mount'];
 export interface AgentWorkspaceCreateOptions {
   sourcePath: string;
   agent: string;
-  model?: string;
+  model: string;
   runtime?: string;
   name?: string;
   project?: string;
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -258,6 +258,11 @@ describe('watchInstancesFile', () => {
   });
 });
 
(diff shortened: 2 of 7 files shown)


benoitf (approved):
I faced the issue

approving but seems e2e are failing, maybe rebase ?

bmahabirbu (approved):
codewise LGTM

I gave it a test but for providers like claude it autoselected a model before me with no way to deselect a model so i couldnt verify if the buttons greyed out if no model was preselected

on another note I think a nice follow up would be for a provider with no connection i.e "No compatible models available" we should also grey out the continue button till a user created that connection and selects a model

MarsKubeX:
> codewise LGTM
> 
> I gave it a test but for providers like claude it autoselected a model before me with no way to deselect a model so i couldnt verify if the buttons greyed out if no model was preselected
> 
> on another note I think a nice follow up would be for a provider with no connection i.e "No compatible models available" we should also grey out the continue button till a user created that connection and selects a model

I have added the grey out when there is no compatible models @bmahabirbu @serbangeorge-m 


https://github.com/user-attachments/assets/dd38467f-de9d-4031-9ec3-0c562229c2c1

jeffmaury (approved):

PR #2364, opened 2026-07-03
feat(agent-workspace): auto-execute agent command on terminal connect

### LINKED ISSUE #2357: When switching to the workspace terminal, execute the agent command

Agents now registers their command.

Have the agent automatically executed

_No response_

### Additional context

_No response_

MarsKubeX:
When switching to the workspace terminal for the first time, look up the agent's registered command from the sandbox labels and write it to the PTY once the shell is ready, removing the need for users to type it manually.

Closes #2357

https://github.com/user-attachments/assets/51214cbf-380a-4458-9305-901d7e3c6dd1

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -89,6 +89,7 @@ const openshellCli = new OpenshellCli({} as Exec, {} as CliToolRegistry);
 
 const agentRegistry = {
   getAgentRegistration: vi.fn(),
+  getAgent: vi.fn(),
 } as unknown as AgentRegistry;
 
 const mockTask = {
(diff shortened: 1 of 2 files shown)


jeffmaury (changes requested):
The command should be set at creation time see https://github.com/openkaiden/kaiden/blob/0e5e19bf5ad3e2ce844137b65da1e5acb7137634/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts#L212

MarsKubeX:
> The command should be set at creation time see
> 
> https://github.com/openkaiden/kaiden/blob/0e5e19bf5ad3e2ce844137b65da1e5acb7137634/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts#L212

Thanks for the suggestion @jeffmaury . Tried with ``command: [agent.command] `` and it doesn't work. The command in openshell sandbox create runs immediately during provisioning, before providers are ready (fails with "Missing provider: claude-code"). It's not stored as a connect-time command. So we need to write it to the terminal after the sandbox is fully up.

bmahabirbu (approved):
Image 

Gave it a whirl tested on my openclaw pr and it works!!

LGTM test wise

jeffmaury:
> > The command should be set at creation time see
> > https://github.com/openkaiden/kaiden/blob/0e5e19bf5ad3e2ce844137b65da1e5acb7137634/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts#L212
> 
> Thanks for the suggestion @jeffmaury . Tried with `command: [agent.command] ` and it doesn't work. The command in openshell sandbox create runs immediately during provisioning, before providers are ready (fails with "Missing provider: claude-code"). It's not stored as a connect-time command. So we need to write it to the terminal after the sandbox is fully up.

Can you try adding --no-auto-providers when creating the sandbox (as we are supposed to give all the required credential) as it fixes the Missing provider on my side

MarsKubeX:
> > > The command should be set at creation time see
> > > https://github.com/openkaiden/kaiden/blob/0e5e19bf5ad3e2ce844137b65da1e5acb7137634/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts#L212
> > 
> > 
> > Thanks for the suggestion @jeffmaury . Tried with `command: [agent.command] ` and it doesn't work. The command in openshell sandbox create runs immediately during provisioning, before providers are ready (fails with "Missing provider: claude-code"). It's not stored as a connect-time command. So we need to write it to the terminal after the sandbox is fully up.
> 
> Can you try adding --no-auto-providers when creating the sandbox (as we are supposed to give all the required credential) as it fixes the Missing provider on my side

Tested with ``--no-auto-providers`` + ``command: [agent.command]`` (both with and without --no-tty). The provider skip works, but claude still fails because createSandbox spawns via child_process from Node.js, no real TTY exists to forward, so Claude CLI falls into --print mode and errors out.

bmahabirbu:
I just thought of a nice follow up it would be great to have an override feature for agents that need a more specific commands (like maybe for an agent that defaults to something else and we want to override so it runs the cli)

jeffmaury (approved):
LGTM

PR #2370, opened 2026-07-03
fix(openshell): generate JWT signing certs for auto-started gateway

### LINKED ISSUE #2369: Auto started gateway is missing tls information
### Bug description

The auto started gateway is missing the signing information.

See https://github.com/openkaiden/kaiden/blob/2719b5c37e380d337020b8af50eab7e7d154f9ec/packages/main/src/plugin/openshell-cli/openshell-gateway.ts#L130

A configuration file needs to be generated (see https://github.com/openkaiden/kaiden/blob/35094c9b12506386af35eb8d8b395595b3a3ae73/extensions/openshell-podman-gateway/src/manager/gateway-container-manager.ts#L40)

In order to generate the 3 files, call openshell-gateway generate-certs --server-san 127.0.0.1 --server-san localhost --server-san host.openshell.internal --output dir tmpdir and then reference the files in the configuration from that file.

### Operating system

MacOS

### Installation Method

None

### Version

0.2.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

jeffmaury:
## Summary
- Before spawning the `openshell-gateway` binary, generate JWT signing certificates via `openshell-gateway generate-certs` using the internal `Exec` utility
- Write a `gateway.toml` config file referencing the generated JWT key paths
- Pass `--config <path>` to the gateway process so it starts with proper signing configuration
- Falls back gracefully if cert generation fails (starts without JWT config)

Fixes #2369

## Test plan
- [x] Existing 26 tests updated to account for new `--config` argument
- [x] New test: generates certs by calling the gateway binary with `generate-certs`
- [x] New test: writes `gateway.toml` config with JWT paths
- [x] New test: starts gateway without `--config` when cert generation fails (graceful degradation)
- [x] `pnpm run typecheck:main` passes
- [x] `pnpm run lint:check` passes

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
@@ -27,8 +27,7 @@ import { beforeEach, describe, expect, test, vi } from 'vitest';
 import type { CliToolRegistry } from '/@/plugin/cli-tool-registry.js';
 import type { Directories } from '/@/plugin/directories.js';
 import type { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
-import type { Proxy } from '/@/plugin/proxy.js';
-import { Exec } from '/@/plugin/util/exec.js';
+import type { Exec } from '/@/plugin/util/exec.js';
 import type { CliToolInfo } from '/@api/cli-tool-info.js';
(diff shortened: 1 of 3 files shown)


benoitf (approved):
LGTM but untested

PR #2373, opened 2026-07-06
fix(agent-workspace): warn when workspace name is already taken

### LINKED ISSUE #2312: Workspace wizard doesn't warn about name uniqueness
### Bug description

openshell sandbox names are unique. If you create a sandbox for a folder `foo`, it will name the sandbox `foo` by default. After creation, if you want to create another sandbox for the same folder, the same name will be used. only after you went all the way through the wizard steps, will you get an error like 

>Failed to create agent workspace Error: Command execution failed with exit code 1 (stderr: Error:   × sandbox 'foo' already exists
  │ 
  │ hint: delete it first with: openshell sandbox delete <name>
  │       or use a different name)

Image

We could validate the name early by checking the existing sandbox names and prevent the user from using an existing name.

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
## Summary
- Validates workspace name against existing sandboxes (case-insensitive) during creation
- Shows inline error using the `Input` component's built-in error styling when a duplicate name is detected
- Disables Continue and quick-create buttons until the user picks a unique name
- Error clears reactively as the user types a new name

Fixes https://github.com/openkaiden/kaiden/issues/2312

Create a workspace, then try creating another with the same name — error should appear and buttons should be disabled

Screenshot From 2026-07-06 04-25-27

### DIFF
--- a/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
+++ b/packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
@@ -249,14 +249,17 @@ let error = $state('');
 let currentStepId = $derived(wizardSteps[wizard.draft.currentStepIndex]?.id ?? '');
 let isLastStep = $derived(wizard.draft.currentStepIndex === wizardSteps.length - 1);
 let hasModel = $derived(wizard.draft.selectedModel !== undefined);
-let isNameDuplicate = $derived.by(() => {
+let validationErrors = $derived.by(() => {
+  const errors: { name?: string } = {};
   const name = wizard.draft.sessionName.trim().toLowerCase();
-  if (!name) return false;
-  return $allOpenshellSandboxes.some(s => s.name.toLowerCase() === name);
+  if (name && $allOpenshellSandboxes.some(s => s.name.toLowerCase() === name)) {
+    errors.name = 'A workspace with this name already exists. Please choose a different name.';
+  }
+  return errors;
 });
 let isCurrentStepComplete = $derived.by(() => {
   if (currentStepId === 'workspace') {
-    return wizard.draft.sessionName.trim() !== '' && wizard.draft.sourcePath.trim() !== '' && !isNameDuplicate;
+    return wizard.draft.sessionName.trim() !== '' && wizard.draft.sourcePath.trim() !== '' && !validationErrors.name;
   }
   if (currentStepId === 'agent-model') {
     return hasModel;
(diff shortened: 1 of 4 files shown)


bmahabirbu:
@coderabbit resolve

gastoner (approved):
Codewise LGTM

jeffmaury (approved):

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
```diff
@@ -248,9 +249,14 @@ let error = $state('');
 let currentStepId = $derived(wizardSteps[wizard.draft.currentStepIndex]?.id ?? '');
 let isLastStep = $derived(wizard.draft.currentStepIndex === wizardSteps.length - 1);
 let hasModel = $derived(wizard.draft.selectedModel !== undefined);
+let isNameDuplicate = $derived.by(() => {
```
fbricon: I would use isNameValid, as we could later enhance validation requirements. isNameDuplicate is too focused

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
```diff
@@ -529,7 +535,8 @@ async function startWorkspace(): Promise<void> {
                 startAsIsDisabled={!hasModel}
                 projects={[...$workspaceProjectInfos]}
                 selectedProjectId={wizard.draft.selectedProjectId}
-                onProjectSelect={handleProjectSelect} />
+                onProjectSelect={handleProjectSelect}
+                nameDuplicate={isNameDuplicate} />
```
fbricon: how about passing an errors object instead, with errors.name containing the error message for the name field (or an error code'), that way we can easily cover other field validation if needed

fbricon (changes requested):

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
bmahabirbu: sounds good will name change

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
bmahabirbu: thats fair ill add this!

bmahabirbu: I also moved the error message to the parent which keeps validation logic and messaging in one place, so the child just displays whatever error it receives! Keeps things more universal

bmahabirbu:
@coderabbit resolve

bmahabirbu:
@fbricon updated!

fbricon (approved):

PR #2403, opened 2026-07-08
fix(openshell): register gateway CLI tool when binary is not found

### LINKED ISSUE #2379: openshell-gateway is not registered if not found
https://github.com/openkaiden/kaiden/blob/351c92b0c027fb679e24540fe1ae0dfd32ab3480/extensions/openshell/src/manager/openshell-cli-manager.ts#L106

bmahabirbu:
## Summary
- Removed early `return` in `OpenshellCliManager.init()` that prevented `openshell-gateway` from being registered when the binary was absent
- The gateway CLI tool is now always registered with a fallback entry, matching the existing pattern for `openshell` and `openshell-image-builder`
- Added test verifying all three CLI tools are registered even when no binaries are found

Fixes https://github.com/openkaiden/kaiden/issues/2379

## Test plan
if gateway binary isn't there we should still register it!

### DIFF
--- a/extensions/openshell/src/manager/openshell-cli-manager.spec.ts
+++ b/extensions/openshell/src/manager/openshell-cli-manager.spec.ts
@@ -68,6 +68,26 @@ beforeEach(() => {
 });
 
 describe('OpenshellCliManager', () => {
+  test('registers gateway CLI tool even when binary is not found', async () => {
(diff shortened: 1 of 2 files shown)


bmahabirbu:
@coderabbit resolve

jeffmaury (approved):

MarsKubeX (approved):

PR #2437, opened 2026-07-10
fix: ensure provider network rules are applied to sandbox

### LINKED ISSUE #2417: Apply provider related rules in the generated policy

Since we switched to a single policy applied to a sandbox instead of performing an update, this one is not having the default policy applied.

So we must add the rules defined in the corresponding provider profile for each provider/secret when generating the policy file

_No response_

### Additional context

_No response_

fbricon:
Enable v2 providers globally (--global) before creating the sandbox so
provider network rules are active at creation time. After creation, add
workspace-specific network endpoints incrementally via
`openshell policy update --add-endpoint` instead of `policy set`,
which did a full replace and broke filesystem policy on live sandboxes.

- Move v2 check+enable before createSandbox, use --global flag
- Add `updatePolicy` method to OpenshellCli using --add-endpoint and
  --binary flags
- Add `formatEndpointFlag` and `collectEndpointFlags` helpers to
  convert OpenshellEndpoint objects to CLI flag strings
- Remove `js-yaml` dump import (no longer writing temp policy files)

https://github.com/user-attachments/assets/4388b8e9-be28-4ee3-8813-10b332d893e5

Fixes #2417

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -353,7 +353,7 @@ describe('create – OpenShell mode', () => {
     vi.mocked(openshellCli.isV2ProviderEnabled).mockResolvedValue(false);
     await manager.create(defaultOptions);
 
-    expect(openshellCli.enableV2Provider).toHaveBeenCalledWith('my-sandbox');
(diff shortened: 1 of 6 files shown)


jeffmaury (approved):

MarsKubeX (approved):
LGTM

PR #2110, opened 2026-06-09
feat: add agents session

benoitf:
allow to interact with agents using ACP protocol.

fixes https://github.com/openkaiden/kaiden/issues/1935

PR #2455, opened 2026-07-14
refactor(agent-workspace): consolidate terminal sessions

### LINKED ISSUE #2430: `AgentWorkspaceManager` tracks terminal session state across four parallel data structures

`AgentWorkspaceManager` tracks terminal session state across four parallel data structures:

- `terminalCallbacks` (Map<callbackId, {write, resize}>)
- `terminalProcesses` (Map<callbackId, IPty>)
- `commandExecutedWorkspaces` (Set<workspaceId>)
- `activeWorkspaceTerminals` (Map<workspaceId, {callbackRef}>)

These all represent facets of the same concept — "the terminal session for a workspace" — but are maintained independently. This makes cleanup coordination fragile (every lifecycle event must remember to touch all four collections) and increases the risk of them drifting out of sync, as seen in #2425.

Consolidate the four collections into a single `Map<string, WorkspaceTerminalSession>` keyed by workspace ID:

```typescript
interface WorkspaceTerminalSession {
  callbackId: number;
  pty: IPty;
  write: (param: string) => void;
  resize: (w: number, h: number) => void;
  commandExecuted: boolean;
}

private readonly workspaceTerminals = new Map<string, WorkspaceTerminalSession>();

bmahabirbu:
## Summary

- Consolidates native OpenShell terminal state into a single workspace-keyed `WorkspaceTerminalSession` map
- Routes send, resize, close, remove, and dispose cleanup through that consolidated session object
- Preserves current refresh behavior by closing any stale workspace PTY before opening a fresh terminal instead of rebinding a new renderer callback to an existing agent TUI
- Adds unit coverage for fresh terminal creation, stale output isolation, and session cleanup

## Context

Closes #2430.

This follows the cleanup concern raised by Marcel on #2426: terminal session state had become split across multiple collections, which made lifecycle cleanup fragile.

It intentionally does not carry forward the session-preservation behavior from #2426. As discussed in #2454, native TUI agents such as OpenCode can react differently to SIGWINCH, resize, and terminal state restoration than Claude, so reattaching arbitrary TUIs after a renderer refresh is fragile. The expected native OpenShell behavior remains: after a confirmed refresh, the old agent terminal is lost and the user gets a fresh sandbox shell.

Longer-term agent session resume should live in ACP/native agent sessions, as explored in #2110, where the app has protocol-level session state instead of trying to reconstruct a raw PTY/TUI screen.

## Test plan

- `pnpm exec vitest --run --project=main packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts`

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -1245,6 +1245,88 @@ describe('dispose', () => {
   });
 });
 
+describe('terminal IPC session lifecycle', () => {
+  function createTerminalMockPty(): { pty: IPty; triggerData: (data: string) => void } {
+    let onDataCb: ((data: string) => void) | undefined;
+    const pty = {
+      onData: vi.fn((cb: (data: string) => void) => {
+        onDataCb = cb;
+        return { dispose: vi.fn() };
+      }),
+      onExit: vi.fn(() => ({ dispose: vi.fn() })),
+      write: vi.fn(),
+      resize: vi.fn(),
+      kill: vi.fn(),
+      pid: 789,
+    } as unknown as IPty;
+    return {
+      pty,
+      triggerData: (data: string): void => {
+        onDataCb?.(data);
+      },
+    };
+  }
+
+  function getIpcHandler<T>(channel: string): T {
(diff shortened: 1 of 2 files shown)


MarsKubeX (approved):
LGTM. Thanks for the refactor.

PR #2464, opened 2026-07-15
fix(agent-workspace): validate workspace name against hostname limit

### LINKED ISSUE #2462: Workspace creation fails when name produces a hostname longer than 64 characters
### Bug description

Creating a workspace with a long name that exceeds 64 characters fails during sandbox creation.

Root cause (suspected) OpenShell/Podman uses the sandbox container name as the hostname. Linux/crun enforces a 64-byte hostname limit. Kaiden does not validate workspace name length before create.

### Operating system

macOS

### Installation Method

### Version

0.3.0

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

serbangeorge-m:
## Summary
- Validate workspace names against the 64-character hostname limit before sandbox creation
- Show inline validation errors in the create wizard and block submission for names longer than 64 characters
- Add a backend guard in `AgentWorkspaceManager` for explicit and basename-derived names

Fixes openkaiden/kaiden#2462

## Test plan
- [x] Unit tests for shared validation helper
- [x] Unit tests for backend rejection of names longer than 64 characters
- [x] Unit tests for create wizard validation and disabled Continue button
- [x] Manually verify a 56 -character workspace name is accepted
- [x] Manually verify a 57+ character workspace name shows an error and cannot be created

Screenshot 2026-07-16 at 9 59 35 AM


fbricon:
I've read that DNS hostname limit is 63 chars, but in practice, I can't create an openshell workspace with a name bigger than 56 chars. 57 fails:

> sandbox create --name abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklzx --no-tty -- true 
> Error:   × code: 'Internal error', message: "create sandbox failed: podman API error (500): crun: sethostname: Invalid argument: OCI runtime error"

56 pass:
> openshell sandbox create --name abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklz --no-tty -- true
> Created sandbox: abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklz

serbangeorge-m:
> I've read that DNS hostname limit is 63 chars, but in practice, I can't create an openshell workspace with a name bigger than 56 chars. 57 fails:
> 
> > sandbox create --name abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklzx --no-tty -- true
> > Error:   × code: 'Internal error', message: "create sandbox failed: podman API error (500): crun: sethostname: Invalid argument: OCI runtime error"
> 
> 56 pass:
> 
> > openshell sandbox create --name abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklz --no-tty -- true
> > Created sandbox: abcdefghijklmnopqrstuvwxyz1234567890qwertyuiopasdfghjklz

ok, I can set the limit to 56

packages/api/src/agent-workspace-info.ts
```diff
@@ -58,6 +58,16 @@ export type AgentWorkspaceMcpConfig = configComponents['schemas']['McpConfigurat
 
 export type AgentWorkspaceMount = configComponents['schemas']['Mount'];
 
+/** Maximum sandbox name length for OpenShell. Podman/crun hostname composition leaves a conservative 56-char limit. */
+export const SANDBOX_NAME_MAX_LENGTH = 56;
```
benoitf, line 62: question:should it be a configuration setting rather than an hardcoded value ?

another question: should it depend on the driver being used ?

serbangeorge-m: I think it can be dynamic, yes. Should I update it?

serbangeorge-m: or we can leave it like this for now and create a follow up ticket to make it dynamic as it will require more testing? I'm ok with both options. What do you think?

MarsKubeX (approved):
LGTM. Just one thing, when creating a workspace with empty name, the workspace name default is the project folder name. If creating 2 workspace with empty name for the same project folder, the "A workspace with this name already exists. Please choose a different name." error won't be displayed. This could be solved by forcing to be a non-empty field. Could be checked in a follow-up issue since it's not fully related to the issue.

fbricon:
@MarsKubeX can be dealt with as part of https://github.com/openkaiden/kaiden/issues/1250

PR #2471, opened 2026-07-16
fix(workspaces): resolve tilde and symlinks in workspace paths

### LINKED ISSUE #1483: Workspace path starting with ~ are not interpolated to the real path leading to errors
### Bug description

When creating a new workspace, if the workspace path starts with ~, it's not interpolated to the home directory, so the generated path is made relative to the kaiden working directory

Image

kdn supports ~/path/to/workspace natively

### Operating system

macOS

fbricon:
**fix(workspaces): resolve tilde and symlinks in workspace paths (#1483)**
Workspace paths starting with ~ were not expanded to the real home
directory, causing filesystem operations to fail or create files
relative to the working directory.

- Normalize sourcePath with resolveHomePath() in create() so tilde
  is expanded but symlinks are preserved for mount resolution
- Resolve symlinks with realpath() at the upload level in
  buildOpenshellFilesystemUploads so $SOURCES-relative mounts
  resolve against the user's original path, not the symlink target
- Apply resolveHomePath() in checkWorkspaceConfigExists()
- Handle bare ~ (not just ~/) in the resolveHomePath utility

related small fix:
**fix(workspaces): display host path instead of target for mounts**
The workspace details overview was showing mount.target (the sandbox-side
path) instead of mount.host (the source path on the host machine). This
caused mounts like ~/Dev/projects/gh-dashboard to display as just
'gh-dashboard' in the Filesystem section.

fixes #1483

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -16,8 +16,9 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
+import type { Stats } from 'node:fs';
 import { access, lstat, readFile, realpath, rm, writeFile } from 'node:fs/promises';
-import { homedir } from 'node:os';
+import { homedir, tmpdir } from 'node:os';
 import { join, resolve } from 'node:path';
 
 import type {
@@ -55,6 +56,7 @@ import type { TaskState, TaskStatus } from '/@api/taskInfo.js';
 import { AgentWorkspaceManager, encodeWorkspaceLabels } from './agent-workspace-manager.js';
 
 vi.mock(import('node:fs/promises'));
+vi.mock(import('node:os'));
 vi.mock(import('js-yaml'));
 vi.mock(import('yaml'));
 vi.mock(import('node-pty'));
@@ -169,6 +171,7 @@ beforeEach(() => {
   vi.mocked(writeFile).mockResolvedValue(undefined);
   vi.mocked(rm).mockResolvedValue(undefined);
   vi.mocked(readFile).mockResolvedValue('{}');
+  vi.mocked(realpath).mockImplementation(async (p: unknown) => p as string);
   vi.mocked(configurationRegistry.getConfiguration).mockReturnValue({
     get: vi.fn().mockReturnValue(undefined),
   } as unknown as ReturnType<IConfigurationRegistry['getConfiguration']>);
@@ -187,8 +190,10 @@ beforeEach(() => {
       isDirectory: () => isDirectory,
       isFile: () => !isDirectory,
       isSymbolicLink: () => false,
-    } as Awaited<ReturnType<typeof lstat>>;
+    } as Stats;
   });
+  vi.mocked(homedir).mockReturnValue('/home/testuser');
+  vi.mocked(tmpdir).mockReturnValue('/tmp');
   gatewayStartCallback = undefined;
   gatewayInitFailedCallback = undefined;
   sandboxListChangeCallback = undefined;
@@ -311,7 +316,6 @@ describe('create – OpenShell mode', () => {
     vi.mocked(openshellCli.createSandbox).mockResolvedValue(undefined);
     vi.mocked(agentRegistry.getAgentRegistration).mockReturnValue(mockAgent);
     vi.mocked(readFile).mockRejectedValue(mockEnoent());
-    vi.mocked(realpath).mockImplementation(async (p: unknown) => p as string);
   });
 
   test('calls openshellCli.createSandbox with gateway, name, providers, workspace label, and agent label', async () => {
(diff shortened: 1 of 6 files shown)


packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
```diff
@@ -56,6 +56,10 @@ import type { TaskState, TaskStatus } from '/@api/taskInfo.js';
 import { AgentWorkspaceManager, encodeWorkspaceLabels } from './agent-workspace-manager.js';
 
 vi.mock(import('node:fs/promises'));
+vi.mock(import('node:os'), async importOriginal => {
+  const actual = await importOriginal();
+  return { ...actual, homedir: vi.fn().mockReturnValue('/home/testuser') };
```
jeffmaury: **suggestion:** I would prefer mocking the whole module then vi.mocked(homedir)...

packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
```diff
@@ -374,6 +379,71 @@ describe('create – OpenShell mode', () => {
...
+  test('resolves symlinks in sourcePath via realpath', async () => {
+    vi.mocked(realpath).mockImplementation(async (p: unknown) => {
+      if (String(p) === '/tmp/symlinked-project') return '/real/path/project' as never;
+      return p as string;
+    });
+    vi.mocked(lstat).mockResolvedValue({
+      isDirectory: () => true,
+      isFile: () => false,
+      isSymbolicLink: () => false,
+    } as Awaited<ReturnType<typeof lstat>>);
```
jeffmaury: **question:** why not using Stats ?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
```diff
@@ -374,6 +379,71 @@ describe('create – OpenShell mode', () => {
...
+  test('resolves symlinks in mount host paths via realpath', async () => {
+    vi.mocked(realpath).mockImplementation(async (p: unknown) => {
+      if (String(p) === join(homedir(), 'linked-dir')) return '/real/linked-dir' as never;
+      return p as string;
+    });
+    vi.mocked(lstat).mockResolvedValue({
+      isDirectory: () => true,
+      isFile: () => false,
+      isSymbolicLink: () => false,
+    } as Awaited<ReturnType<typeof lstat>>);
```
jeffmaury: **question:** idem

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -113,6 +114,7 @@ export class AgentWorkspaceManager implements Disposable {
   ) {}
 
   async create(options: AgentWorkspaceCreateOptions): Promise<AgentWorkspaceId> {
+    options.sourcePath = await realpath(resolveHomePath(options.sourcePath));
```
MarsKubeX: Should this be inside the try/catch block?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -281,11 +283,12 @@ export class AgentWorkspaceManager implements Disposable {
...
 
     for (const mount of workspace.mounts ?? []) {
-      const local = this.resolveHostPath(mount.host, sourcePath);
+      const raw = this.resolveHostPath(mount.host, sourcePath);
       const remote = this.resolveOpenshellSandboxPath(mount.target, sourcePath);
-      if (!local || !remote) {
+      if (!raw || !remote) {
         continue;
       }
+      const local = await realpath(raw);
```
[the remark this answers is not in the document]
fbricon: should be fixed

bmahabirbu (commented):
So the pr is correctly translating the symlink to a real path but there is a problem with how the $SOURCES are handled imo

For example:

User selects: /home/brian/work/app-link
Symlink target: /mnt/repos/app

For the main upload, Kaiden should use:

/mnt/repos/app

because OpenShell must receive the real directory contents.

But consider:

$SOURCES/../shared-lib

Currently

It canonicalizes first:

sourcePath = /mnt/repos/app

Then resolves the mount:

/mnt/repos/app/../shared-lib
→ /mnt/repos/shared-lib

That may be the wrong repository.

Instead

It first resolves the mount according to what the user selected:

/home/brian/work/app-link/../shared-lib
→ /home/brian/work/shared-lib

Then it canonicalizes that final mount path for upload:

realpath(/home/brian/work/shared-lib)
→ whatever physical directory that selected sibling represents

This gives OpenShell a physical directory while preserving the intended relationship and fixes the $SOURCES/../… regression

bmahabirbu (approved):
Ok cool looks like the symlink for sources has been fixed. Appreciate the new fix!

PR #2474, opened 2026-07-16
feat(semantic-router): add default model support

### LINKED ISSUE #2436: No default model available

When adding models to a semantic router, use should be able to define a default model

The first model added is the default one. Then if there are several models added, user should have a way to change the default model

_No response_

### Additional context

_No response_

jeffmaury:
## Summary
- Add `defaultModelRef` optional field to `RoutingConfigSchema` so semantic routers can designate a fallback model for unmatched requests
- First selected model is automatically set as default; users can change it via an inline "Default" checkbox column in the model selection table
- Default model is persisted in the router JSON config and displayed with a badge in the router cards view

## Test plan
- [x] `pnpm run typecheck` passes (0 errors)
- [x] All 72 unit tests pass across 4 test files (manager, create wizard, step models, cards)
- [ ] Manual: create a semantic router with 2+ models, verify first is auto-defaulted
- [ ] Manual: change default via checkbox, confirm config includes correct `defaultModelRef`
- [ ] Manual: verify card shows "default" badge next to the default model
- [ ] Manual: verify old routers without `defaultModelRef` load and display correctly

Fixes #2436

### DIFF
--- a/packages/api/src/semantic-router-info.ts
+++ b/packages/api/src/semantic-router-info.ts
@@ -59,6 +59,7 @@ export const DecisionConfigSchema = z.object({
 export const RoutingConfigSchema = z.object({
   keywords: z.array(KeywordGroupSchema),
   decisions: z.array(DecisionConfigSchema),
+  defaultModelRef: ModelRefSchema.optional(),
 });
 
 export const SemanticRouterConfigSchema = z.object({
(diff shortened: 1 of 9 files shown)


bmahabirbu (approved):
LGTM codwise and gave it a test. The checkbox correctly gets the default and the badge should up

small nit 

we could have a test that when Kaiden creates the semantic router, the config passed into factory.create({ config: ... }) includes routing.defaultModelRef just to check the manager to provider factory route

PR #2475, opened 2026-07-16
fix(openshell-cli): auto-refresh sandbox list while deleting sandboxes

### LINKED ISSUE #2330: Openshell workspaces does not disappear after deleting.
### Bug description

When deleting an openshell workspace, it remains in the workspace list showing its last known phase (e.g. "Deleting"). The only way to remove it is to restart the app.

Root cause: openshell sandbox delete returns before the sandbox is fully purged from the gateway. Subsequent list refreshes still see the sandbox in a transitional phase, so it lingers indefinitely.

Fix: Track confirmed deletions optimistically on both the main process (pendingDeletions set in AgentWorkspaceManager) and renderer side (deletingSandboxNames store) to immediately show "Deleting" status and filter out the entry once the CLI confirms acceptance.

https://github.com/user-attachments/assets/3bd20e58-5cd9-4644-ad7f-39929d24d4e4

### Operating system

all

### Installation Method

_No response_

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

jeffmaury:
## Summary

- When listing sandboxes via `openshell sandbox list`, if any sandbox is in "Deleting" phase, automatically re-poll after 5 seconds to detect when the deletion completes
- The emitter only fires when the total sandbox count or deleting count actually changes, avoiding unnecessary UI refreshes
- `AgentWorkspaceManager` subscribes to the new `onDidSandboxListChange` event and forwards `agent-workspace-update` to the renderer

## Test plan

- [x] Unit tests for auto-refresh polling logic (schedule, dedup, change detection, error handling, dispose)
- [x] Unit test for `AgentWorkspaceManager` subscription wiring
- [x] Typecheck passes (`pnpm run typecheck:main`)
- [x] Lint passes (`pnpm run lint:check`)

Fixes #2330

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -127,6 +127,7 @@ const providerRegistry = {
 } as unknown as ProviderRegistry;
 
 let gatewayStartCallback: (() => void) | undefined;
+let sandboxListChangeCallback: (() => void) | undefined;
 
 const openshellGateway = {
   onDidGatewayStart: vi.fn((cb: () => void) => {
@@ -178,6 +179,15 @@ beforeEach(() => {
     } as Awaited<ReturnType<typeof lstat>>;
   });
   gatewayStartCallback = undefined;
+  sandboxListChangeCallback = undefined;
+  Object.defineProperty(openshellCli, 'onDidSandboxListChange', {
+    value: vi.fn((cb: () => void) => {
+      sandboxListChangeCallback = cb;
+      return { dispose: vi.fn() };
+    }),
+    writable: true,
+    configurable: true,
+  });
   manager = new AgentWorkspaceManager(
     apiSender,
     ipcHandle,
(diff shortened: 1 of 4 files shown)


bmahabirbu (approved):
LGTM codewise and tested this I think 5 seconds is a good time for polling!

Orthogonal, while testing I noticed that Kaiden currently treats these two situations the same:

  - The gateway is reachable and has no sandboxes.
  - The gateway cannot be reached, so Kaiden cannot determine whether it has sandboxes.

So would be a good follow up to have kaiden give an error for a gateway cant be reached

benoitf (approved):

fbricon:
I think we could still improve the design. Right now the deletion is blocking, so this.apiSender.send('agent-workspace-update'); will only be triggered after the actual deletion happened. In the meantime, the UI still shows the sandbox in Ready mode. Then the sandbox just disappears.
We could probably start the deletion promise, tell the UI to refresh within 1 sec (would show Deleting), then await deletion, while the UI keeps polling the sandbox list status. I'd use a Fibonacci-style backoff delays [1s, 2s, 3s, 5s]
  (capped at 5s) instead of a fixed 5s interval

PR #2481, opened 2026-07-17
fix(opencode): use anthropic provider prefix for vertexai models

### LINKED ISSUE #2473: Configure Opencode to work with Vertex AI-backed models
### Task content

Currently opencode is configured to use a vertexai provider (as in opencode provider), however, no google ADC file is injected in the sandbox, instead, we should reproduce the configuration presented in https://docs.nvidia.com/openshell/latest/providers/google-vertex-ai and inject 

> ANTHROPIC_BASE_URL="https://inference.local/v1" 
ANTHROPIC_API_KEY=unused 

It's unclear yet what model name should be injected in the opencode config file

fbricon:
Opencode needs `anthropic/<modelname>` in its config, not `vertexai/<modelname>`,
because Vertex AI inference is handled transparently at the Openshell level.
Opencode uses the Anthropic SDK with ANTHROPIC_BASE_URL and ANTHROPIC_API_KEY
env vars (already injected) to reach the inference endpoint.

Add a PROVIDER_ALIASES map to remap vertexai → anthropic for the opencode
config model string and provider block.

fixes #2473

### DIFF
--- a/extensions/opencode/src/extension.spec.ts
+++ b/extensions/opencode/src/extension.spec.ts
@@ -506,6 +506,36 @@ describe('activate', () => {
       expect(written.mcp['minimal']).not.toHaveProperty('environment');
     });
 
+    test('maps vertexai to anthropic provider with correct config and env vars', async () => {
+      await activate(extensionContextMock);
+      const agent = vi.mocked(agents.registerAgent).mock.calls[0]![0];
+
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

fbricon:
@coderabbitai resolve

PR #2486, opened 2026-07-17
fix(openshell): use extension assets folder for bundled binaries in dev mode

### LINKED ISSUE #2485: OpenShell binaries assets are not registered in development mode
### Bug description

if we run the `pnpm build` in the OpenShell extension in order to fetch OpenShell binaries into the `assets` folder
and then run `kaiden`, it'll try to execute commands like `openshell` from the global PATH and not from the bundled assets.

I would expect that if these assets exist, it should reuse them so no need to export manually PATH even in development mode

### Operating system

macOS 26.5.2

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

- `pnpm build` in the OpenShell extension
- `pnpm watch` on the root folder

### Relevant log output

```shell

```

### Additional context

_No response_

benoitf:
fixes #2485 

In development mode, process.resourcesPath is not set, so bundled OpenShell binaries were never found and the system PATH was used instead. Now checks the extension's assets/<platform>-<arch>/ folder when not in production, matching the layout created by download scripts.

note: [tsconfig.json](https://github.com/openkaiden/kaiden/compare/main...benoitf:kortex:KAIDEN-2485?expand=1#diff-eb770dba3ebcbcbaa8d276a032a9746a0f02734156e0dc414003341b24cbce7e) is updated to allow the usage of `if (import.meta.env.PROD)`

### DIFF
--- a/extensions/openshell/src/manager/openshell-cli-manager.spec.ts
+++ b/extensions/openshell/src/manager/openshell-cli-manager.spec.ts
@@ -21,7 +21,7 @@ import { existsSync, readFileSync } from 'node:fs';
 import { join } from 'node:path';
 
 import { cli, configuration, process as extensionProcess } from '@openkaiden/api';
-import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { assert, beforeEach, describe, expect, test, vi } from 'vitest';
 
 import { OpenshellCliManager } from './openshell-cli-manager';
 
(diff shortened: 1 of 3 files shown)


extensions/openshell/src/manager/openshell-cli-manager.spec.ts
```diff
@@ -221,8 +210,54 @@ describe('OpenshellCliManager', () => {
       expect(manager.getRegisteredPath()).toBe('/usr/local/bin/openshell');
       // bundled binary should NOT have been version-checked because system PATH was found first
       expect(extensionProcess.exec).not.toHaveBeenCalledWith(bundledPath, expect.anything());
+    });
+
+    test('uses process.resourcesPath with original subdir in production mode', async () => {
+      vi.stubEnv('PROD', 'true');
```
bmahabirbu: vi.stubEnv('PROD', true); instead of the string 'true'

extensions/openshell/src/manager/openshell-cli-manager.ts
```diff
@@ -199,9 +199,19 @@ export class OpenshellCliManager implements Disposable {
     binaryName: string,
     resourceSubdir: string,
   ): Promise<BinaryDiscoveryResult | undefined> {
-    const resourcesPath = (process as NodeJS.Process & { resourcesPath?: string }).resourcesPath;
+    let resourcesPath: string | undefined;
+    let bundledResourceSubdir: string;
+    if (import.meta.env.PROD) {
+      resourcesPath = (process as NodeJS.Process & { resourcesPath?: string }).resourcesPath;
+      bundledResourceSubdir = resourceSubdir;
+    } else {
```
bmahabirbu, line 210: Need something like

`} else {
    resourcesPath = join(
      this.extensionContext.extensionUri.fsPath,
      'assets',
      resourceSubdir === 'openshell-image-builder' ? 'image-builder' : '',
    );
    bundledResourceSubdir = `${process.platform}-${process.arch}`;
  }
`

To include the image-builder binary since it has a /image-builder/ before the /platform-arch/ or else dev wouldnt be able to use the binary. 

Production handles it differently because it copies the source directory and flattens it into a production location

jeffmaury (changes requested):
Typecheck errors to be fixed and image-builder is not correctly processed

benoitf:
@jeffmaury / @bmahabirbu this is fixed

jeffmaury (approved):

PR #2489, opened 2026-07-20
fix: remove duplicate entries in pnpm-lock.yaml and add vitePreprocess for TypeScript with experimental.async

fbricon:
The @ai-sdk/svelte 5.0.28->5.0.30 bump (59af9590f) introduced
duplicate YAML keys in the lockfile snapshots section. This caused
pnpm to ignore the lockfile entirely and re-resolve all dependencies
from scratch, pulling breaking versions (e.g. msw 2.15.0 which
changes the setupServer() return type).

add vitePreprocess for TypeScript with experimental.async

The Svelte 5 native TypeScript handling fails to strip optional
parameter syntax (?) when experimental.async compiler option is
enabled, causing rollup parse errors during the renderer build.

Adding vitePreprocess with script:true ensures TypeScript is
explicitly preprocessed before the Svelte compiler processes it.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -100,10 +100,10 @@ importers:
         version: 0.3.3
       dockerode:
         specifier: ^4.0.9
-        version: 4.0.9
+        version: 4.0.12
       drizzle-orm:
         specifier: ^0.45.2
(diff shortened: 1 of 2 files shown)


fbricon:
superseded by #2490

PR #2490, opened 2026-07-20
chore: fix lock file

benoitf:
remove duplicated entries


fbricon (approved):

fbricon:
There's still something weird, as if we actually update all dependencies to latest (like generating the lockfile from scratch) then we run into these compilation issues (#2489).

PR #2495, opened 2026-07-20
chore: apply correct auth in update OpenShell workflow

benoitf:


.github/workflows/nightly-openshell-update.yaml
```diff
@@ -100,6 +100,9 @@ jobs:
 
       - name: Update existing PR
         if: steps.compare.outputs.needs_update == 'true' && steps.existing_pr.outputs.exists == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.KAIDEN_BOT_TOKEN }}
+          GH_TOKEN: ${{ secrets.KAIDEN_BOT_TOKEN }}
```
fbricon: why do we need 2 envvars for the same token?

benoitf: gh cli is using GH_TOKEN env var while github actions read GITHUB_TOKEN

benoitf: reading again it seems that now `gh` cli is also defaulting to GITHUB_TOKEN

will remove one env var

fbricon (approved):

PR #2501, opened 2026-07-20
feat: support gateway selection for workspaces

### LINKED ISSUE #2487: Bind workspace operations to the proper gateway
### Task content

When performing action on workspace (create/delete), ensure the gateway name is part of the request so that the correct gateway is targeting. This mean that if there is several gateways, the create wizard should offer a selector to select the proper gateway

bmahabirbu:
## Summary

- bind workspace creation and deletion to the selected OpenShell gateway
- add a gateway selector when multiple gateways are available and default it to OpenShell's active gateway
- scope duplicate-name validation and cleanup operations to the target gateway

Fixes #2487

## Tests

Ok to test this on podman is pretty involved because if we have multiple gateways connected to the same podman socket (i.e rootless socket) openshell sandbox list will show dups of each sandbox.

The reason is since the gateways have access to both. Its a openshell bug because the sandbox list should filter by namespace which it currently does not as of 0.0086

Docker driver works fine however because it correctly filters the namespice even if both gateways have access to containers 

Now if you want to continue with podman the workaround is to create a gateway that connects to the rootful socket. On mac since only one podman machine can be run at the same time it wont work. Maybe its possible to have docker container runtime running at the same time to create a gateway targetted to that socket but i havent tested this?

On linux just start the rootful socket

`sudo systemctl start podman.socket`

  Create a minimal development-auth configuration:

```
  cat > /tmp/openshell-rootful-demo.toml <<'EOF'
  [openshell]
  version = 1

  [openshell.gateway.auth]
  allow_unauthenticated_users = true
EOF
```

Start the gateway in a new terminal

```
  sudo -H env \
    OPENSHELL_LOCAL_TLS_DIR=/home/brian/.local/state/openshell/tls \
    OPENSHELL_PODMAN_SOCKET=/run/podman/podman.sock \
    OPENSHELL_DRIVERS=podman \
    /usr/bin/openshell-gateway \
      --config /tmp/openshell-rootful-demo.toml \
      --disable-tls \
      --bind-address 0.0.0.0 \
      --port 17675 \
      --db-url 'sqlite::memory:?cache=shared'

```

  Now create the sandbox again:

```
  openshell --gateway podman-rootful-demo sandbox create \
    --name rootful-demo-box \
    --no-tty \
    -- sh -lc 'echo "Created by the rootful gateway"'
```

`openshell --gateway openshell sandbox create     --name rootless-demo-box     --no-tty     -- sh -lc 'echo "Created by the rootless gateway"'`

So now that created a simple sandbox using both gateways

Start kaiden with this pr you should see both sandboxes with both gateways able to delete them and in the workspace creation now see a gateway category 

https://github.com/user-attachments/assets/4c5d2b96-6979-41b7-b6e5-4271fabd4fab

gateway selector will only show if multi gateways is detected

### DIFF
--- a/packages/api/src/agent-workspace-info.ts
+++ b/packages/api/src/agent-workspace-info.ts
@@ -75,6 +75,7 @@ export interface AgentWorkspaceCreateOptions {
   sourcePath: string;
   agent: string;
   model: string;
+  gateway: string;
   name?: string;
   project?: string;
   skills?: string[];
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.spec.ts
@@ -287,6 +287,7 @@ describe('create – OpenShell mode', () => {
     agent: 'claude',
     name: 'my-sandbox',
     model: 'ramalama::granite-4.6::',
+    gateway: 'kaiden',
   };
 
   const mockAgent: Agent = {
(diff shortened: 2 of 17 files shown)


bmahabirbu:
@coderabbit resolve

bmahabirbu:
@coderabbit resolve

jeffmaury (approved):

jeffmaury (commented):
The list of secrets available to set into the workspace needs to be filtered by gateway as well

bmahabirbu:
> The list of secrets available to set into the workspace needs to be filtered by gateway as well

I will address this too but I think this adds to the scope of this pr. I have a potential solution but increases the size by a lot since it has to touch the secrets manager code parts too

bmahabirbu:
@coderabbit resolve

PR #2502, opened 2026-07-21
refactor(openshell): share GitHub artifact downloader

### LINKED ISSUE #2488: Factorise OpenShell and Openshell image builder download
### Task content

I noticed we have two classes
https://github.com/openkaiden/kaiden/blob/59af9590f478c4d5fae8a3eb0ca99bd2f8bebb6d/extensions/openshell/src/openshell-download.ts
and
https://github.com/openkaiden/kaiden/blob/59af9590f478c4d5fae8a3eb0ca99bd2f8bebb6d/extensions/openshell/src/openshell-image-builder-download.ts

and also two scripts
https://github.com/openkaiden/kaiden/blob/59af9590f478c4d5fae8a3eb0ca99bd2f8bebb6d/extensions/openshell/scripts/download-image-builder.ts

and
https://github.com/openkaiden/kaiden/blob/59af9590f478c4d5fae8a3eb0ca99bd2f8bebb6d/extensions/openshell/scripts/download.ts

issue: I think it's sharing almost the same code but for some reasons, there are differences/duplication while their goal is to fetch GitHub artifacts

To be consistent, it should use a single download script that would fetch GitHub artifacts and share the same logic.
Only the name of the artifact/github repository should be the parameters/difference

bmahabirbu:
## Summary

- replace the duplicate OpenShell and openshell-image-builder download modules with one configuration-driven GitHub artifact pipeline
- consolidate build-time downloads into one script while preserving the existing package commands and output layouts
- use the shared downloader from both runtime installers and cover both artifact formats

## Local verification

The refactor was verified with real pinned release artifacts, not only mocked unit tests:

```bash
pnpm --dir extensions/openshell exec tsx scripts/download.ts --platform linux --arch x64
```

- downloaded OpenShell v0.0.69 and openshell-image-builder v0.9.0 through the unified script
- verified every GitHub release checksum
- executed both downloaded binaries with `--version` and confirmed the expected versions
- reran the unified command and confirmed both components reused their cached artifacts

Additional checks:

- `pnpm exec vitest --run extensions/openshell/src` — 35 tests passed
- `pnpm run typecheck:extensions:openshell`
- scoped ESLint, Biome formatting, and `git diff --check`

Fixes #2488


bmahabirbu:
@coderabbit resolve

MarsKubeX (approved):
LGTM

bmahabirbu:
@coderabbit resolve

PR #2520, opened 2026-07-21
fix(opencode): avoid provider ID collision with OpenCode built-in loaders

### LINKED ISSUE #2519: TypeError: sdk.responses is not a function when using OpenAI Compatible provider with OpenCode
## Problem

When creating an Agent Workspace with OpenCode using a model served through the "OpenAI Compatible" provider (e.g. LiteMaaS/LiteLLM), the workspace terminal immediately crashes with:

```
TypeError: sdk.responses is not a function. (In 'sdk.responses(modelID)', 'sdk.responses' is undefined)
```

This affects any model connected via Kaiden's "OpenAI Compatible" extension (LiteMaaS, vLLM, Ollama via OpenAI compat, etc.) when used with OpenCode.

## Root Cause

The OpenCode extension generates a config file with the provider ID `"openai"` (from `llmMetadata.name`) and the npm package `@ai-sdk/openai-compatible`. However, OpenCode has a built-in custom loader for the `"openai"` provider ID that unconditionally calls `sdk.responses(modelID)` — the Responses API. The `@ai-sdk/openai-compatible` SDK package does not expose a `.responses()` method (it only supports `/v1/chat/completions`), causing the crash.

| SDK Package | `.responses()` | `.languageModel()` |
|---|---|---|
| `@ai-sdk/openai` | Yes | Yes |
| `@ai-sdk/openai-compatible` | **No** | Yes |

## Expected behavior

Workspaces using OpenCode with any OpenAI-compatible endpoint should start successfully and use the Chat Completions API (`/v1/chat/completions`) without errors.

## Acceptance criteria

- [ ] Workspaces using an OpenAI Compatible provider with OpenCode start without `sdk.responses` errors
- [ ] The generated `opencode.json` uses a provider ID that does not collide with OpenCode's built-in loaders (`openai`, `azure`, `azure-cognitive-services`)
- [ ] Existing providers (Anthropic, Ollama, Vertex AI, native OpenAI via built-in, Mistral, Google) continue to work as before
- [ ] Unit tests cover the provider ID renaming for affected providers

### Operating system

all

### Installation Method

None

### Version

0.3.0

## Reproduction steps

1. Add an inference connection via the "OpenAI Compatible" provider (e.g. a LiteMaas endpoint with Qwen3.6-35B-A3B)
2. Create a new Agent Workspace with OpenCode as the agent
3. Select a model from the OpenAI Compatible connection
4. Observe the `TypeError: sdk.responses is not a function` error in the workspace terminal

### Relevant log output

```shell

```

### Additional context

_No response_

MarsKubeX:
## Summary

- Fix `TypeError: sdk.responses is not a function` when using the OpenAI Compatible provider with OpenCode agent workspaces
- Suffix the provider ID to `"openai-compat"` when the configured npm package is `@ai-sdk/openai-compatible` and the ID would collide with OpenCode's built-in loaders (`openai`, `azure`, `azure-cognitive-services`)
- Add unit test covering the provider ID renaming

## Problem

OpenCode has built-in custom loaders for specific provider IDs (e.g. `"openai"`) that unconditionally call `sdk.responses(modelID)`. When a user connects via Kaiden's "OpenAI Compatible" extension (LiteMaaS, vLLM, etc.), `llmMetadata.name` is `"openai"`, which was used directly as the provider key in the generated `opencode.json`. Since the SDK package `@ai-sdk/openai-compatible` does not expose `.responses()`, the workspace crashes immediately on start.

## Fix

The extension now detects when the provider ID would collide with a built-in loader and suffixes it to `"openai-compat"`. This bypasses the built-in loader, causing OpenCode to use `sdk.languageModel()` instead — which works correctly with `@ai-sdk/openai-compatible`.

## Test plan

- [x] Unit tests pass (31/31)
- [ ] Create a workspace with OpenCode + OpenAI Compatible provider (e.g. LiteMaaS with Qwen3.6-35B-A3B) → no `sdk.responses` error
- [ ] Create a workspace with OpenCode + Anthropic (native) → still works as before
- [ ] Create a workspace with OpenCode + Ollama → still works as before
- [ ] Create a workspace with OpenCode + Vertex AI → still works as before

Closes #2519

### DIFF
--- a/extensions/opencode/src/extension.spec.ts
+++ b/extensions/opencode/src/extension.spec.ts
@@ -234,6 +234,32 @@ describe('activate', () => {
       expect(written.provider.anthropic.options.baseURL).toBe('https://custom.anthropic.example.com');
     });
 
+    test('renames openai provider ID to avoid OpenCode built-in loader collision', async () => {
+      await activate(extensionContextMock);
+      const agent = vi.mocked(agents.registerAgent).mock.calls[0]![0];
+
+      const configFile = createConfigFile();
+      await agent.preWorkspaceStart(
+        createContext([configFile], {
+          provider: 'openai',
+          modelLabel: 'Qwen3.6-35B-A3B',
+          endpoint: 'https://litellm.example.com/v1',
+        }),
+      );
+
+      const written = JSON.parse(configFile.updateMock.mock.calls[0]![0] as string);
+      expect(written.model).toBe('openai-compat/Qwen3.6-35B-A3B');
+      expect(written.provider['openai-compat']).toEqual({
+        name: 'openai',
+        npm: '@ai-sdk/openai-compatible',
+        options: { apiKey: '{env:OPENAI_API_KEY}', baseURL: 'https://litellm.example.com/v1' },
+        models: {
+          'Qwen3.6-35B-A3B': { _launch: true, name: 'Qwen3.6-35B-A3B' },
+        },
+      });
+      expect(written.provider).not.toHaveProperty('openai');
+    });
+
     test('does not add provider block for native provider with endpoint', async () => {
       await activate(extensionContextMock);
       const agent = vi.mocked(agents.registerAgent).mock.calls[0]![0];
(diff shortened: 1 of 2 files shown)


jeffmaury (approved):
LGTM

PR #2524, opened 2026-07-22
fix(secret-vault): remove unsupported other secret type

### LINKED ISSUE #2506: Remove Other type of secret
### Task content

The Other type of secret should be removed as it is not supported by OpenShell, it should have been removed when kdn was removed

bmahabirbu:
## Summary

- remove the unsupported Other secret option and its custom injection fields
- require users to select an available secret card before creating a secret
- remove obsolete secret grouping and Other-specific utilities
- update secret-vault tests for service-profile-only creation

## Tests

- `vitest run --project renderer packages/renderer/src/lib/secret-vault` (27 tests passed)
- ESLint on all changed files
- Svelte autofixer on changed Svelte components

Closes #2506


MarsKubeX (approved):
LGTM

PR #2572, opened 2026-07-23
feat: add acp session list

### LINKED ISSUE #1935: Evaluate ACP integration into Kaiden
### Task content

Evaluate various options for integrating ACP in Kaiden

- libraries
- impact
- OpenShell relationship

benoitf:
I had to move my fork under a new name so it closed all my existing PRs, reopening it

it's successor of https://github.com/openkaiden/kaiden/pull/2110

fixes https://github.com/openkaiden/kaiden/issues/1935


bmahabirbu (commented):
While testing I noticed the agent doesnt have claude as a selection? it seems for sandboxes that wasnt created by kaiden the agent isnt auto detected which is ok by design but in doing so i didnt see claude

testing with opencode and switching models doesnt work because the sandbox only
permits the attached model inference router at runtime. So u cant switch models. 

tested with opencode with ramalama and that worked fine

benoitf:
Claude is not supported. It requires a separate agent that is not Claude code but something based on the Claude sdk

bmahabirbu (approved):
Gave it a test again and end to end works for me! I think the small things can be handled as follow ups!!

PR #2575, opened 2026-07-23
feat: allow creating empty workspaces without a project folder

### LINKED ISSUE #2518: Allow creating an empty workspace
### Task content

It should be possible to create an empty workspace/sandbox, just like openshell, so do not require a project folder in the workspace wizard. You would expect the agent will be able to clone a repository once the workspace is ready (on its own or via its human).

fbricon:
Make sourcePath optional in AgentWorkspaceCreateOptions so users can
create sandbox-only workspaces. For no-folder workspaces, store the
workspace.json config under the global state directory scoped by
gateway and workspace name (<dataDir>/agent-workspaces/<gateway>/<name>/).

- Add getAgentWorkspacesConfigDirectory() to Directories interface
- Support optional overrideConfigDir in writeWorkspaceConfig()
- Update AgentWorkspaceManager to handle no-folder creation, config
  read/write, and cleanup on deletion
- Process mounts (absolute//Users/fbricon paths) even without sourcePath;
  warn when $SOURCES-relative mounts are skipped
- Update UI wizard to make project folder optional
- Add tests for all new paths

Screenshot 2026-07-23 at 14 29 47

fixes #2518

### DIFF
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -30,6 +30,7 @@ import { AgentRegistry } from '/@/plugin/agent-registry.js';
 import { updateWorkspaceConfig, writeWorkspaceConfig } from '/@/plugin/agent-workspace/workspace-config-writer.js';
 import { WritableConfigurationFile } from '/@/plugin/agent-workspace/writable-configuration-file.js';
 import { IPCHandle, WebContentsType } from '/@/plugin/api.js';
+import { Directories } from '/@/plugin/directories.js';
 import { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
 import { OpenshellGateway } from '/@/plugin/openshell-cli/openshell-gateway.js';
 import {
@@ -111,7 +112,16 @@ export class AgentWorkspaceManager implements Disposable {
     private readonly agentRegistry: AgentRegistry,
     @inject(OpenshellGateway)
     private readonly openshellGateway: OpenshellGateway,
+    @inject(Directories)
+    private readonly directories: Directories,
   ) {}
+
+  private getGlobalConfigDir(gateway: string, sandboxName: string): string {
+    if (/[\\/]|\.\./.test(gateway) || /[\\/]|\.\./.test(sandboxName)) {
+      throw new Error(`Invalid workspace identifier: "${gateway}/${sandboxName}"`);
+    }
+    return join(this.directories.getAgentWorkspacesConfigDirectory(), gateway, sandboxName);
+  }
 
   async create(options: AgentWorkspaceCreateOptions): Promise<AgentWorkspaceId> {
     const suffix = options.name ? ` "${options.name}"` : '';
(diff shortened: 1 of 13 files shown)


packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -111,8 +112,14 @@ export class AgentWorkspaceManager implements Disposable {
     private readonly agentRegistry: AgentRegistry,
     @inject(OpenshellGateway)
     private readonly openshellGateway: OpenshellGateway,
+    @inject(Directories)
+    private readonly directories: Directories,
   ) {}
 
+  private getGlobalConfigDir(gateway: string, sandboxName: string): string {
+    return join(this.directories.getAgentWorkspacesConfigDirectory(), gateway, sandboxName);
+  }
```
fbricon, line 124: this will be handled by https://github.com/openkaiden/kaiden/issues/2576

fbricon: ... but the proposed fix for `getGlobalConfigDir` is fine, we can add it anyways

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -111,8 +112,17 @@ export class AgentWorkspaceManager implements Disposable {
...
+    @inject(Directories)
+    private readonly directories: Directories,
   ) {}
 
+  private getGlobalConfigDir(gateway: string, sandboxName: string): string {
+    if (/[\\/]|\.\./.test(gateway) || /[\\/]|\.\./.test(sandboxName)) {
+      throw new Error(`Invalid workspace identifier: "${gateway}/${sandboxName}"`);
+    }
+    return join(this.directories.getAgentWorkspacesConfigDirectory(), gateway, sandboxName);
+  }
```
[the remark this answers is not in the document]
fbricon: `my..app` would actually be an invalid name, according to the new rules in https://github.com/openkaiden/kaiden/pull/2577

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
```diff
@@ -442,7 +441,6 @@ function buildMountsFrom(fileAccess: string, mounts: CustomMount[]): AgentWorksp
 async function startAsIs(): Promise<void> {
   if (
     !wizard.draft.selectedGateway ||
-    !wizard.draft.sourcePath.trim() ||
```
MarsKubeX, line 453: if this line is removed, you can have space only paths, resulting in something like ' /.kaiden'. Consider to trim the path ``const sourcePath = draftSnapshot.sourcePath.trim() || undefined;``

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
```diff
@@ -481,7 +479,6 @@ async function startWorkspace(): Promise<void> {
   if (
     !wizard.draft.selectedGateway ||
     !getEffectiveWorkspaceName() ||
-    !wizard.draft.sourcePath.trim() ||
```
MarsKubeX, line 492: Same here

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte
fbricon: nice catch!

MarsKubeX (approved):
LGTM

benoitf (approved):
was able to start a workspace without specifying a folder

PR #2584, opened 2026-07-24
fix: align CLI Tools card separators with Resources page

serbangeorge-m:
## Summary
- Apply the Podman Desktop Resources separator pattern to the CLI Tools page: replace `divide-x` with explicit `border-r` and even `pr-5` / `px-5 py-2` padding.
- Use a fixed `w-[170px]` left column so separators stay vertically aligned across cards with different tool name lengths.
- Drop float-based header layout in favor of a simple flex row for the display name / “Registered by” line.

Screenshot 2026-07-24 at 11 24 42 AM

## Test plan
- [x] Open Settings → CLI Tools with multiple registered tools (different name lengths)
- [x] Confirm vertical separators are evenly padded and aligned across cards
- [x] Compare visually with Settings → Resources separator treatment
- [x] `pnpm exec vitest run packages/renderer/src/lib/preferences/PreferencesCliTool.spec.ts`

### DIFF
--- a/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
+++ b/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte
@@ -164,7 +164,7 @@ function getLoggerHandler(_cliToolId: string): ConnectionCallback {
           {/if}
           <span
             id={cliTool.id}
-            class="my-auto ml-3 break-words font-semibold text-[var(--pd-invert-content-header-text)]"
+            class="my-auto ml-3 break-all font-semibold text-[var(--pd-invert-content-header-text)]"
             aria-label="cli-name">{cliTool.name}</span>
         </div>
         <div class="flex flex-row space-x-1 w-full">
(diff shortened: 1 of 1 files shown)


benoitf (dismissed):
hello, is it a cherry-pick of the file from Podman Desktop ?

because the upstream file does not look like your changes ?
https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte

serbangeorge-m:
> hello, is it a cherry-pick of the file from Podman Desktop ?
> 
> because the upstream file does not look like your changes ? https://github.com/podman-desktop/podman-desktop/blob/main/packages/renderer/src/lib/preferences/PreferencesCliTool.svelte

Kaiden’s CLI Tools page already matched Podman Desktop. The problem here is longer names like openshell-image-builder make the left column wider, so the lines between cards don’t line up.

This PR first used the Resources page fix from PD [#15570](https://github.com/podman-desktop/podman-desktop/pull/15570). That change only updated Resources though — CLI Tools in Podman Desktop still uses the old layout I think.

Options:

1. Keep the Podman Desktop layout and just set a fixed width on the left column for long names.
2. Do that same fixed-width change in Podman Desktop first, then cherry-pick it into Kaiden.

MarsKubeX (commented):
LGTM but IMHO it might be better to upstream this to PD first and then cherry picking it. WDYT @benoitf ?

benoitf:
files need to be the aligned so as we're rebasing we should yes do fixes upstream and cherry-pick files after

serbangeorge-m:
@benoitf @MarsKubeX now it is a cherry pick

MarsKubeX (approved):
LGTM. Thanks for the changes.

serbangeorge-m:
@coderabbitai resolve

fbricon (approved):

PR #2587, opened 2026-07-24
ci: disable scheduled runs for workspaces e2e tests on azure instances

serbangeorge-m:


fbricon:
can you just comment as to why e2e tests on azure are disabled?

serbangeorge-m:
> can you just comment as to why e2e tests on azure are disabled?

this workflow is running workspace creation e2e tests and for windows we do not have the openshell tools bundled yet.

fbricon (approved):
ok thx

PR #2590, opened 2026-07-27
fix(openshell): write gateway logs to file

### LINKED ISSUE #2549: Auto started gateway should not log message in Kaiden console
### Task content

As it generates a lot of messages, the Kaiden console is then barely readable.

I think it should log to a file located under the Kaiden directories

bmahabirbu:
## Summary

This pr pipes the gateway.process logs for the "kaiden-local" gateway to a logfile preserving the console clarity since those logs were verbose and made the console unreadable. 

https://github.com/user-attachments/assets/f8e78623-7741-48ed-aa9b-3f4a8e1a20ed

https://github.com/user-attachments/assets/8a257cd3-0274-4f53-99f8-32e7de41bc7a

## Local testing

1. Remove the registered default gateway:

   ```sh
   openshell gateway remove openshell
   ```

2. Stop any existing `openshell-gateway` process.

   Linux:

   ```sh
   ps aux | grep -i openshell-gateway | grep -v grep
   kill <PID>
   ```

   macOS:

   ```sh
   ps aux | grep -i "[o]penshell-gateway"
   kill <PID>
   ```

3. Start Kaiden and allow it to create a new gateway:

   ```sh
   pnpm watch
   ```

4. Inspect the gateway log.

   Linux:

   ```sh
   tail -f "${XDG_DATA_HOME:-$HOME/.local/share}/kaiden/openshell-gateway/gateway.log"
   ```

   macOS:

   ```sh
   tail -f "$HOME/.local/share/kaiden/openshell-gateway/gateway.log"
   ```

   If `KAIDEN_HOME_DIR` is set, inspect `$KAIDEN_HOME_DIR/openshell-gateway/gateway.log` instead.

5. Verify that:

the logfile was created and u see the log process

Fixes #2549

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
@@ -18,6 +18,7 @@
 
 import type { ChildProcess } from 'node:child_process';
 import { EventEmitter } from 'node:events';
+import type { WriteStream } from 'node:fs';
 import { mkdir, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
 
@@ -35,16 +36,29 @@ import type { GatewayInfo } from '/@api/openshell-gateway-info.js';
 import { OpenshellGateway } from './openshell-gateway.js';
 
 vi.mock(import('node:child_process'));
+vi.mock(import('node:fs'));
 vi.mock(import('node:fs/promises'));
 vi.mock(import('/@/plugin/util/exec.js'));
 
 const { spawn } = await import('node:child_process');
+const { createWriteStream } = await import('node:fs');
 
 const GATEWAY_BINARY = '/usr/local/bin/openshell-gateway';
 const KAIDEN_DATA_DIRECTORY = '/home/user/.local/share/kaiden';
 const GATEWAY_STORAGE_DIRECTORY = join(KAIDEN_DATA_DIRECTORY, 'openshell-gateway');
 const GATEWAY_CONFIG_PATH = join(GATEWAY_STORAGE_DIRECTORY, 'gateway.toml');
 const GATEWAY_DB_URL = `sqlite:${join(GATEWAY_STORAGE_DIRECTORY, 'gateway.db')}?mode=rwc`;
+const GATEWAY_LOG_PATH = join(GATEWAY_STORAGE_DIRECTORY, 'gateway.log');
+
+type MockWriteStream = WriteStream & {
+  write: ReturnType<typeof vi.fn>;
+  end: ReturnType<typeof vi.fn>;
+};
+
+const gatewayLogStream = Object.assign(new EventEmitter(), {
+  write: vi.fn(),
+  end: vi.fn(),
+}) as unknown as MockWriteStream;
 
 function createMockChildProcess(): ChildProcess & { _stdout: EventEmitter; _stderr: EventEmitter } {
   const proc = new EventEmitter() as ChildProcess & { _stdout: EventEmitter; _stderr: EventEmitter };
@@ -91,13 +105,14 @@ beforeEach(() => {
   vi.mocked(cliToolRegistry.getCliToolInfos).mockReturnValue([
     { name: 'openshell-gateway', path: GATEWAY_BINARY },
   ] as unknown as CliToolInfo[]);
+  vi.mocked(createWriteStream).mockReturnValue(gatewayLogStream);
   vi.mocked(exec.exec).mockResolvedValue({ command: '', stdout: '', stderr: '' });
   gateway = new OpenshellGateway(cliToolRegistry, openshellCli, directories, exec, notificationRegistry);
 });
 
 describe('init', () => {
-  test('skips auto-start when existing gateway is healthy and already active', async () => {
-    vi.spyOn(console, 'log').mockImplementation(() => undefined);
+  test('creates the gateway log and reuses a healthy active gateway', async () => {
+    const consoleLog = vi.spyOn(console, 'log').mockImplementation(() => undefined);
     const existingGateways: GatewayInfo[] = [
       { name: 'local-gw', endpoint: 'https://127.0.0.1:8443', active: true, type: 'local' },
     ];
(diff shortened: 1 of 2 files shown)


bmahabirbu:
Looking for some feedback on what logs should stay in console and what ones should be piped to the logfile and if the logfile directory is in the right spot? Also if we should name the logfile gateway.log but have a directory called logfiles and have scaffolding for future logfile specific things?

benoitf (changes requested):
hello, I'm not sure I follow why the we remove console.log for the gateway and move it to logToFile

My understanding is that the log file should contain the output of the gateway start command, nothing more, nothing less

MarsKubeX (commented):
Agree with @benoitf, the file should capture everything (for debugging), but the console should retain its existing lifecycle messages

bmahabirbu:
The issue mention that the "Auto started gateway should not log message in Kaiden console as it generates a lot of messages, the Kaiden console is then barely readable" so I gathered that the logs from the gateway clog the console but after testing it doesnt seem that way maybe it was for a gateway kaiden spawns

Intead it could be that the issue means log all gateway specific console logs to its own file so its easier to see just the gateway logs? I'll update the pr to do this

benoitf:
I read the same description of the issue and to me it's to redirect the output of the gateway process

bmahabirbu:
OK confirmed thanks Florent after some testing the noisy logs that Jeff was talking about comes from when Kaiden starts a gateway not when it detects an already running gateway

So these processes 

  gatewayProcess.stdout.on('data', ...)
  gatewayProcess.stderr.on('data', ...)

Those listeners remain active for the gateway’s entire lifetime. Everything was forwarded through console.log/console.error

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -355,10 +376,47 @@ export class OpenshellGateway implements Disposable {
...
+  }
+
+  private logGatewayMessage(level: GatewayLogLevel, ...args: unknown[]): void {
+    console[level](...args);
+    this.#gatewayLogStream?.write(`${format(...args)}\n`);
+  }
+
+  private logGatewayProcessOutput(data: Buffer): void {
+    this.#gatewayLogStream?.write(data);
+  }
```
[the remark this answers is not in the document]
bmahabirbu: This is the technically correct viewpoint but i feel like is definitely overkill for a simple logfile like the one i propose here

MarsKubeX (approved):
LGTM

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -355,10 +381,48 @@ export class OpenshellGateway implements Disposable {
...
+  private async initializeGatewayLog(): Promise<void> {
+    if (this.#gatewayLogStream) {
+      return;
     }
+
+    const storageDirectory = join(this.directories.getDataDirectory(), 'openshell-gateway');
+    const logPath = join(storageDirectory, GATEWAY_LOG_FILENAME);
+    try {
+      await mkdir(storageDirectory, { recursive: true });
+      const stream = createWriteStream(logPath, { flags: 'w' });
```
fbricon, line 386: On my machine I see growth rate: ~2.4 MB/hour → ~19 MB per 8-hour session
I was worried the log file would grow indefinitely, but the w flag overwrites it on each new session. That's OK I guess

fbricon (approved):

benoitf (changes requested):
please do not change existing console.log entries

only wrap the output of the process to a file

benoitf (commented):
`logGatewayMessage` method is useless

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Thanks for bringing the issue of size up and I agree the rewrite should eliminate that. I think for a production environment it makes sense to rewrite on new process just have to be clear that if a user wants to debug an issue they cannot start kaiden or else the gateway logs get rewritten. 

as a follow up It could be worthwhile to have logfile rotations to save previous logs

bmahabirbu:
Ok appreciate the clarity so just the gateway.process logs to the logfile. Gateway lifecycle will stay only in console (no copy to logfile)

packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
```diff
@@ -35,16 +36,29 @@ import type { GatewayInfo } from '/@api/openshell-gateway-info.js';
 import { OpenshellGateway } from './openshell-gateway.js';
 
 vi.mock(import('node:child_process'));
+vi.mock(import('node:fs'));
 vi.mock(import('node:fs/promises'));
 vi.mock(import('/@/plugin/util/exec.js'));
 
 const { spawn } = await import('node:child_process');
+const { createWriteStream } = await import('node:fs');
```
benoitf: I don't know who introduced lazy imports, but it should NOT import using await import, all imports should be on top of the file (edited)

benoitf (approved):

bmahabirbu:
@benoitf all set for a new review

bmahabirbu:
ah thanks

packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
bmahabirbu: will address maybe an update to agent.md will be useful to catch anymore lazy imports

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -173,29 +177,31 @@ export class OpenshellGateway implements Disposable {
...
-      console.error(`[openshell-gateway] ${text}`);
+      this.#gatewayLogStream?.write(data);
       stderrChunks.push(text);
     });
 
-    this.#gatewayProcess.on('exit', (code, signal) => {
+    gatewayProcess.on('exit', (code, signal) => {
       console.log(`[openshell-gateway] exited with code=${code ?? 'none'} signal=${signal ?? 'none'}`);
       this.#gatewayProcess = undefined;
     });
```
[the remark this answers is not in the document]
bmahabirbu: New scope only gateway.process logs pipe to the logfile console.logs stay in console. This pr aims to solve the issue of console being overwhelmed by logs that dont adhere to its design

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -173,29 +177,31 @@ export class OpenshellGateway implements Disposable {
...
 
-    this.#gatewayProcess.on('error', (err: Error) => {
+    gatewayProcess.on('error', (err: Error) => {
       console.error(`[openshell-gateway] failed to start: ${err.message}`);
       this.#gatewayProcess = undefined;
     });
```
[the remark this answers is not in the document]
bmahabirbu: This race condition was originally already in upstream. So it doesnt have to do with this fix per say. Given that the fix is straightforward i can add this

PR #2596, opened 2026-07-28
fix(ci): utilize GITHUB_TOKEN for binary downloads and CI workflows

### LINKED ISSUE #2592: download script should leverage GITHUB_TOKEN env var if defined
### Task content

on CI jobs, we see more and more 403 rate limit exceeded when downloading OpenShell/gateway binaries at the post-install step

the script should check if GITHUB_TOKEN env var is available and use it

it might also be that GITHUB_TOKEN env var is not propagated to the PR checks workflows

fbricon:
- Use GITHUB_TOKEN in openshell-download.ts when downloading binaries from GitHub
- Pass Authorization header with Bearer token to bypass rate limiting
- Add GITHUB_TOKEN env var to pr-check.yaml lint-format and typecheck jobs
- Add GITHUB_TOKEN env var to npmjs-publish.yaml workflow
- Ensures token is available during post-install script execution in CI

Fixes #2592

### DIFF
--- a/.github/workflows/npmjs-publish.yaml
+++ b/.github/workflows/npmjs-publish.yaml
@@ -97,6 +97,8 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Show publish parameters
         run: |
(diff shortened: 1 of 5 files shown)


benoitf (commented):
the token should be used only for the required steps

benoitf (approved):

bmahabirbu (approved):
Looks good! Thanks for the fix!

fbricon:
@coderabbitai resolve

PR #2602, opened 2026-07-29
fix(#2578): remove misleading Unrestricted network option from workspaces

### LINKED ISSUE #2578: Remove the Unrestricted network option from agent workspaces
### Task content

Unrestricted is misleading.

It sounds like the sandbox can reach the host.
It only turns off the outbound allowlist. The sandbox stays isolated.

serbangeorge-m:
Drop Unrestricted from create/wizard UI. Keep a disabled legacy label for existing allow-mode configs; sandbox isolation was never removed.

Closes https://github.com/openkaiden/kaiden/issues/2578

Screenshot 2026-07-29 at 12 22 02 PM


fbricon (changes requested):
Looks good but you should fix the issue reported by coderabbit

fbricon (approved):

PR #2603, opened 2026-07-29
ci: skip workspace smoke e2e playwright tests on macos/windows gha runners

### LINKED ISSUE #2563: Update workspace playwright e2e tests for multi-gateway support
### Task content

Gateways are being introduced as a core concept across workspace and secret management. Current e2e tests have no gateway awareness and will break once these changes land.

serbangeorge-m:
Requires an OpenShell gateway with the Podman driver, which is not available on macOS/Windows GitHub Actions runners.

Closes https://github.com/openkaiden/kaiden/issues/2563

### DIFF
--- a/tests/playwright/src/specs/guided-setup-smoke.spec.ts
+++ b/tests/playwright/src/specs/guided-setup-smoke.spec.ts
@@ -208,7 +208,7 @@ test.describe('Guided setup smoke', { tag: '@smoke' }, () => {
 
     test.describe('Onboarding - guided setup persistence (Claude)', () => {
       test.skip(
-        process.platform !== 'linux' && !process.env.PODMAN_ENABLED,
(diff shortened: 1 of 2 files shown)


benoitf (commented):
question: shouldn't it check env variable for CI or Github actions ?

if I want to run on a Linux, I would not expect to see error messages about GitHub actions runners

benoitf:
something like GITHUB_ACTIONS https://docs.github.com/en/actions/reference/workflows-and-actions/variables

serbangeorge-m:
> question: shouldn't it check env variable for CI or Github actions ?
> 
> if I want to run on a Linux, I would not expect to see error messages about GitHub actions runners

yes, that's correct, I updated.

benoitf (commented):
I would not pickup CI, you could run that on CI environment where nested environment is available.
The message is specific for GHA, so it needs to check if it's running on GHA, not on any CI

benoitf (approved):

PR #2604, opened 2026-07-29
fix(agent-workspace): persist and display workspace description

### LINKED ISSUE #2383: Workspace description is lost after creation
### Bug description

If you enter a description fo a workspace, it doesn't show up on the workspace list, workspace details page, and doesn't appear to be serialized at all. I believe it should be stored in the .kaiden/workspace.json file

Image

### Operating system

macOS

### Installation Method

None

### Version

next (development version)

### Steps to reproduce

- Create a new workspace
- Type a description
- Complete the new workspace creation workflow

### Relevant log output

```shell

```

### Additional context

_No response_

fbricon:
Persist workspace description at creation time and display it in the
overview page. Make the description editable in Settings > General with
change tracking, save/discard support, and a character counter.

Extracts a shared WorkspaceDescriptionField component used by both the
creation wizard and the settings page. Distinguishes the project path
from the description in the overview by rendering it as a smaller
monospace path with a folder icon.

https://github.com/user-attachments/assets/16683a8a-b1ff-4334-83c0-1058be556d55

fixes #2383


bmahabirbu (approved):
pending review

bmahabirbu (dismissed):
See above didnt mean to approve yet

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetailsSettings.svelte
```diff
@@ -53,6 +54,9 @@ const activeLabel = $derived(sections.find(s => s.id === activeSection)?.label ?
 
 const workspaceName = $derived(workspaceSummary?.name ?? '');
 
+const originalDescription = $derived(configuration?.description ?? '');
+let pendingDescription: string = $state(originalDescription);
```
bmahabirbu:
```suggestion
let pendingDescription = $derived(originalDescription);
```

the originalDescrtion gets loaded async so if og is blank before the update then pendingDescription would be blank.

https://github.com/user-attachments/assets/06f2d2a4-94ff-4276-9525-07c5e125f4b5

 You can reproduce this effect if the render is reloaded on the setting page. It wipes it

bmahabirbu: Otherwise LGTM

bmahabirbu (approved):
pls see above comment but its non blocking imo

packages/renderer/src/lib/agent-workspaces/AgentWorkspaceDetailsSettings.svelte
fbricon: Nice catch! fixed

fbricon:
@coderabbitai resolve

PR #2606, opened 2026-07-29
fix(openshell): remove stale gateway entry before re-registering on startup

### LINKED ISSUE #2605: openshell gateway fails to re-register on restart: 'kaiden-local' already exists
## Bug

When Kaiden restarts, the openshell gateway from the previous session is no longer running, but its `kaiden-local` entry persists in the openshell CLI config. The `init()` flow correctly detects the gateway is unreachable and auto-starts a new one, but `registerWithCli()` then fails with:

```
Gateway 'kaiden-local' already exists.
Remove it first with: openshell gateway remove kaiden-local
Or choose a different name with: --name <name>
```

This was introduced by #2085, which changed `init()` to health-check existing gateways and fall through to auto-start when unreachable. Before that commit, `init()` returned early as soon as `listGateways()` returned any results, so `registerWithCli()` was never called when `kaiden-local` already existed.

## Expected behavior

On restart, the stale `kaiden-local` gateway entry should be cleaned up automatically before re-registering the newly started gateway.

## Actual behavior

Console shows noisy errors and the gateway fails to register with the CLI, even though the gateway process itself starts successfully.

benoitf:
## Summary

- Remove stale `kaiden-local` CLI entry before re-adding in `registerWithCli()`, fixing "Gateway already exists" errors on restart

Fixes #2605

## Test plan

- [x] Existing tests pass (49/49)
- [x] New test: `removes stale gateway entry before re-registering` — verifies `removeGateway` is called before `addGateway`
- [x] New test: `registers successfully even when removeGateway fails` — verifies graceful handling when no prior entry exists

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
@@ -107,6 +107,7 @@ beforeEach(() => {
   vi.mocked(createWriteStream).mockReturnValue(gatewayLogStream);
   vi.mocked(exec.exec).mockResolvedValue({ command: '', stdout: '', stderr: '' });
   vi.mocked(openshellCli.removeGateway).mockResolvedValue();
+  vi.mocked(openshellCli.listGateways).mockResolvedValue([]);
   gateway = new OpenshellGateway(cliToolRegistry, openshellCli, directories, exec, notificationRegistry);
 });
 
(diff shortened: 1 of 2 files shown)


fbricon (approved):
No more errors about existing gateways

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -354,6 +354,7 @@ export class OpenshellGateway implements Disposable {
   private async registerWithCli(): Promise<void> {
     const endpoint = `http://${this.#bindAddress}:${this.#port}`;
     try {
+      await this.openshellCli.removeGateway('kaiden-local').catch(() => {});
       await this.openshellCli.addGateway({ endpoint, local: true, name: 'kaiden-local' });
       console.log(`[openshell-gateway] registered with CLI as kaiden-local at ${endpoint}`);
```
MarsKubeX: Is it not better to, instead of removing and adding, check if the gateway is there? Something like:
```try {
    const gateways = await this.openshellCli.listGateways();
    const existing = gateways.find(gw => gw.name === 'kaiden-local');
    if (existing) {
      if (!existing.active) {
        await this.openshellCli.selectGateway('kaiden-local');
      }
      console.log(`[openshell-gateway] kaiden-local already registered, selected at ${existing.endpoint}`);
      return;
    }
    await this.openshellCli.addGateway({ endpoint, local: true, name: 'kaiden-local' });
    console.log(`[openshell-gateway] registered with CLI as kaiden-local at ${endpoint}`);
```

benoitf: it's like the endpoint might be changed

benoitf: I dropped all my env, will see if can reproduce again my issue and try your suggestion

bmahabirbu (approved):
Overall I think its good on my side but I came across an edge case where openshell gateway remove isn't always reliable. (mac/linux)

I removed the openshell default gateway and started kaiden. It correctly tried to start the kaiden-local gateway but encounter a conflict error because the default gateway still somehow hanged onto the process on that port and kaiden-local uses the same port 17670.

Maybe we need a follow up so we can scan the port activity and fully kill the gateway at the process level

```
11:28:49 AM [main] [openshell-gateway] registered with CLI as kaiden-local at http://127.0.0.1:17670

11:28:49 AM [main] openshell failed: /Users/bmahabir/kaiden/extensions/openshell/assets/darwin-arm64/openshell provider list -o json — Command execution failed with exit code 1 (stderr: Error:   × code: 'Operation is not implemented or not supported', message: "grpc-
  │ status header missing, mapped from HTTP status code 404")
11:28:49 AM [main] openshell failed: /Users/bmahabir/kaiden/extensions/openshell/assets/darwin-arm64/openshell sandbox list -g kaiden-local -o json — Command execution failed with exit code 1 (stderr: Error:   × code: 'Operation is not implemented or not supported', message: "grpc-
```

Edit*

on further inspection the default gateway is created with mtls while local-kaiden uses plaintext. The mismatch from the hanged process caused the grpc error because openshell thinks local-kaiden has mtls

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
benoitf: Adopted this approach : `registerWithCli` now calls `listGateways()` first:
- If `kaiden-local` already exists with the **same endpoint**, skip re-registration entirely
- If it exists with a **different endpoint**, remove then re-add (handles endpoint changes across restarts)
- If it **doesn't exist**, just add directly (no unnecessary remove)

Also added an `invocationCallOrder` assertion per CodeRabbit's suggestion to verify remove-before-add ordering.

packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
```diff
@@ -395,6 +397,40 @@ describe('start', () => {
...
+
+    await gateway.start();
+
+    expect(openshellCli.removeGateway).toHaveBeenCalledWith('kaiden-local');
+    expect(openshellCli.addGateway).toHaveBeenCalledWith({
+      endpoint: 'http://127.0.0.1:17670',
+      local: true,
+      name: 'kaiden-local',
+    });
+  });
```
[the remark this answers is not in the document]
benoitf: Done — added `invocationCallOrder` assertion in the "removes stale gateway and re-registers when endpoint differs" test, following the existing pattern from `semantic-router-manager.spec.ts`.

benoitf:
@fbricon if you can check again as I implemented the proposal of @MarsKubeX 

on my end it's working as well

MarsKubeX (approved):
LGTM codewise. Just check the codeRabbit comment, seems valid to me.

fbricon:
FIx works for me. I am not convinced by coderabbit's explanation though. If openshell fails to list gateways, chances are you won't be able to add a new one

PR #2610, opened 2026-07-29
fix: configure npmjs access for fullsend

### LINKED ISSUE #2609: Configure fullsend access to npmjs
### Task content

#2599 updated the image used by fullsend code and fix agents to one that includes Node 24, but neither before or after image were able to access npmjs to install packages.

deboer-tim:
The Fullsend code and fix agent have not been able to access npmjs to install packages. Based on an analysis of the agent logs the issue appears to be that packages like @podman-desktop/ui-svelte cause encoded slashes to be required, and that's blocked by default in OpenShell's network policy.

The solution is to add `allow_encoded_slash: true` to the agent's policy. This should be done in the fullsend agent repo, but to confirm it is the problem we'll try copying the policy here temporarily. The two policy files are ugly and long, but the only change from the default is line 99/95 where allow_encoded_slash is added. If this works, then this PR would be undone once the fix is released upstream.

Fixes #2609.

### DIFF
--- a/.fullsend/code.yaml
+++ b/.fullsend/code.yaml
@@ -1,2 +1,3 @@
 base: https://raw.githubusercontent.com/fullsend-ai/agents/ca518d9353f5a8363c6ba499d7a2070a1b0e7c5d/harness/code.yaml#sha256=6ac2a3dc3e163e53137d6839b7f31f17a3942d9ca316741d21a041df6784fb1d
 image: ghcr.io/openkaiden/kaiden-fullsend-image@sha256:752de42626946c37a453a0b0ddd2effd7a2a4cecda86595586097d14b1c21635
+policy: .fullsend/policies/code.yaml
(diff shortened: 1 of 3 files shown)


benoitf (commented):
as there is a comment

```
# Identical to the code agent policy — the fix agent needs the same network
```

we might use a single file rather than duplicating it

also while the comment say it's the same there are few differences 

```
  Fewer binaries in fix.yaml:
  - code.yaml allows these extra binaries that fix.yaml omits:
    - **/npx
    - **/yarn
    - **/yarnpkg
    - **/pnpm
```

while I think it's important to have pnpm in both

fbricon (approved):
LGTM

benoitf (commented):
please use a single policy file

fix.yaml is missing `**/pnpm` while it's saying in comments it's the same policy file than code

so better to use a single shared policy file (can link fix to code policy yaml)

benoitf (approved):

PR #2630, opened 2026-07-30
ci: work around podman/crun OCI runtime-spec mismatch on ubuntu-24.04 runners

serbangeorge-m:
## Summary

Every e2e job on `ubuntu-24.04` that creates a Podman-backed provider (Docling,
Milvus) is currently failing with:

```
Error: (HTTP code 500) server error - crun: unknown version specified: OCI runtime error
```

This blocks `knowledge-env-smoke.spec.ts` (KDB-02, KDB-07, and anything
downstream) on every PR and nightly run right now, regardless of the PR's
actual changes.

## Root Cause

The `ubuntu-24.04` GitHub-hosted runner image was updated recently, bumping
Podman from 4.9.3 to 5.8.4 while leaving crun at 1.14.1. Podman ≥5.6.2 emits
OCI runtime-spec v1.2.0/1.2.1 by default, and crun <1.14.3 has a known
version-check bug that rejects that spec string — fixed upstream in crun
1.14.3 ([containers/podman#27272](https://github.com/containers/podman/issues/27272)).

This PR pins crun to 1.28 on Linux e2e runners as a temporary workaround,
removable once the runner image ships a compatible crun.

### DIFF
--- a/.github/workflows/e2e-test-job.yaml
+++ b/.github/workflows/e2e-test-job.yaml
@@ -133,7 +133,9 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           CRUN_VERSION=1.28
+          CRUN_SHA256=2aa6b7024a9c9f153895c0d11ae233d3758f54844011c3a039e3e89048d01d42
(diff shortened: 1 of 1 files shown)


bmahabirbu (approved):
LGTM hope tests pass

PR #2632, opened 2026-07-31
feat(openshell): track gateway lifecycle state

### LINKED ISSUE #2515: Handle state for gateways
### Task content

When detecting or starting gateways, we should monitor their state (available/unavailable, healthy/non healthy, started/stopped) and do not consider the ones that we cannot reach in part of the UI (create workspace)

bmahabirbu:
Fixes #2515

## Summary

- Add reachability and health state to `GatewayInfo`.
- Monitor OpenShell gateways in the main process and notify the renderer when gateway information changes.
- Exclude unreachable gateways from workspace creation.
- Make the monitoring interval configurable through `openshell.gateway.pollInterval` in seconds (default: 5, minimum: 1).

The implementation follows the state-monitor pattern used by Podman Desktop for Kubernetes: a dedicated backend manager owns polling and cached gateway information, emits an invalidation event when its snapshot changes, and the renderer refetches through the existing IPC and store path. Polling is required because OpenShell does not currently expose gateway lifecycle notifications; see NVIDIA/OpenShell#2569.

`GatewayInfo` remains the single gateway object returned through the existing `GatewayInfo[]` IPC contract. The monitor enriches each gateway with `gatewayState`:

```ts
gatewayState: {
  reachable: boolean;
  health: 'healthy' | 'degraded' | 'unhealthy' | 'unknown';
}
```

Reachability is determined by whether `openshell gateway info -g <name>` succeeds. When it succeeds, health comes from the command's `status` field. When it cannot connect, the gateway is reported as unreachable with unknown health.

Existing gateway startup checks continue to use `openshell status`.

## Manual testing

1. Add a secondary gateway whose endpoint is not running:

```sh
openshell gateway add \
  http://127.0.0.1:17671 \
  --local \
  --name test-secondary
```

2. Start Kaiden:

```sh
pnpm watch
```

3. In the renderer DevTools console, inspect the cached gateway information:

```js
await window.listOpenshellGateways()
```

The secondary gateway should contain:

```js
gatewayState: { reachable: false, health: 'unknown' }
```

A running gateway should report `reachable: true` and its OpenShell health status. This call reads Kaiden's cache; it does not invoke the OpenShell CLI directly.

4. Confirm that the unreachable secondary gateway is not offered during workspace creation.

5. Change the polling interval from the DevTools console:

```js
await window.updateConfigurationValue('openshell.gateway.pollInterval', 60)
await window.getConfigurationValue('openshell.gateway.pollInterval')
```

The second command should return `60`, and the monitor reschedules immediately without restarting Kaiden.

6. Remove or otherwise change a gateway with the OpenShell CLI. After the configured polling interval, run:

```js
await window.listOpenshellGateways()
```

The cached gateway information should reflect the CLI change.

### DIFF
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -58,6 +58,7 @@
           "scope": "DEFAULT",
           "default": 5,
           "minimum": 1,
+          "maximum": 3600,
           "description": "Interval in seconds between OpenShell gateway state checks."
         }
       }
--- a/extensions/openshell/package.json
+++ b/extensions/openshell/package.json
@@ -52,6 +52,13 @@
           "when": "!isWindows",
           "default": "",
           "description": "Custom path to openshell-gateway binary (Default is blank)"
+        },
+        "openshell.gateway.pollInterval": {
+          "type": "number",
+          "scope": "DEFAULT",
+          "default": 5,
+          "minimum": 1,
+          "description": "Interval in seconds between OpenShell gateway state checks."
         }
       }
     }
--- a/packages/api/src/openshell-gateway-info.ts
+++ b/packages/api/src/openshell-gateway-info.ts
@@ -18,6 +18,17 @@
 
 import z from 'zod';
 
+export const GatewayHealthSchema = z.enum(['healthy', 'degraded', 'unhealthy', 'unknown']);
+
+export type GatewayHealth = z.output<typeof GatewayHealthSchema>;
+
+export const GatewayStateSchema = z.object({
+  reachable: z.boolean(),
+  health: GatewayHealthSchema,
+});
+
+export type GatewayState = z.output<typeof GatewayStateSchema>;
+
 export const GatewayInfoSchema = z.object({
   name: z.string(),
   endpoint: z.string(),
@@ -28,6 +39,7 @@ export const GatewayInfoSchema = z.object({
   is_remote: z.boolean().optional(),
   remote_host: z.string().nullable().optional(),
   resolved_host: z.string().nullable().optional(),
+  gatewayState: GatewayStateSchema.optional(),
 });
 
 export type GatewayInfo = z.output<typeof GatewayInfoSchema>;
@@ -126,6 +138,7 @@ export interface OpenshellGatewayStartOptions {
 }
 
 export const GatewayRuntimeInfoSchema = z.looseObject({
+  status: GatewayHealthSchema,
   compute_drivers: z.array(
     z.looseObject({
       capabilities: z.looseObject({
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -140,6 +140,14 @@ export class AgentWorkspaceManager implements Disposable {
         throw new Error('model is required to create a workspace');
       }
 
+      await this.openshellGatewayStateManager.whenReady();
+      const gateway = this.openshellGatewayStateManager
+        .listGateways()
+        .find(candidate => candidate.name === options.gateway);
+      if (!gateway?.gatewayState?.reachable) {
+        throw new Error(`gateway "${options.gateway}" is unreachable`);
+      }
+
       if (options.replaceConfig) {
         if (options.sourcePath) {
           await rm(join(options.sourcePath, '.kaiden', 'workspace.json'), { force: true });
--- a/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
+++ b/packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
@@ -33,6 +33,7 @@ import { IPCHandle, WebContentsType } from '/@/plugin/api.js';
 import { Directories } from '/@/plugin/directories.js';
 import { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
 import { OpenshellGateway } from '/@/plugin/openshell-cli/openshell-gateway.js';
+import { OpenshellGatewayStateManager } from '/@/plugin/openshell-cli/openshell-gateway-state-manager.js';
 import {
   buildPolicyObject,
   collectBinaryFlags,
@@ -113,6 +114,8 @@ export class AgentWorkspaceManager implements Disposable {
     private readonly agentRegistry: AgentRegistry,
     @inject(OpenshellGateway)
     private readonly openshellGateway: OpenshellGateway,
+    @inject(OpenshellGatewayStateManager)
+    private readonly openshellGatewayStateManager: OpenshellGatewayStateManager,
     @inject(Directories)
     private readonly directories: Directories,
   ) {}
@@ -559,7 +562,7 @@ export class AgentWorkspaceManager implements Disposable {
   }
 
   async listOpenshellGateways(): Promise<GatewayInfo[]> {
-    return this.openshellCli.listGateways();
+    return [...this.openshellGatewayStateManager.listGateways()];
   }
 
   async deleteOpenshellSandbox(name: string, gateway: string): Promise<void> {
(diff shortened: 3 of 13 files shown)


packages/main/src/plugin/openshell-cli/openshell-gateway-state-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+
+import type { Disposable } from '@openkaiden/api';
+import { inject, injectable, preDestroy } from 'inversify';
+
+import { Emitter } from '/@/plugin/events/emitter.js';
+import { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
+import type { Event } from '/@api/event.js';
+import type { GatewayInfo } from '/@api/openshell-gateway-info.js';
+
+const GATEWAY_POLL_INTERVAL_MS = 5000;
```
benoitf: should be configurable I guess. Also, can we get automatic/push notifications ?

packages/api/src/openshell-gateway-info.ts
```diff
@@ -30,6 +30,23 @@ export const GatewayInfoSchema = z.object({
...
+  reachable: z.boolean(),
+  health: GatewayHealthSchema,
+});
+
+export type GatewayState = z.output<typeof GatewayStateSchema>;
+
+export const GatewayInfoSchema = GatewayRegistrationSchema.extend({
+  gatewayState: GatewayStateSchema,
+});
+
```
benoitf: I don't expect InfoSchema extends GatewayRegistrationSchema

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -279,12 +279,9 @@ export class OpenshellCli {
     const results: GatewaySandboxes[] = [];
     for (const gateway of gateways) {
       try {
-        const sandboxes = await this.listSandboxes(gateway.name);
+        const sandboxes = await this.listSandboxes(gateway.name, { quiet: true });
```
benoitf: why quiet = true ?

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -350,24 +347,29 @@ export class OpenshellCli {
...
+    if (gatewayName) {
+      args.push('-g', gatewayName);
+    }
+    const data = await this.execCLI<unknown>(args, undefined, options?.quiet);
     return GatewayRuntimeInfoSchema.parse(data);
   }
 
   async checkEndpointStatus(endpoint: string): Promise<boolean> {
-    const args = ['status', '--gateway-endpoint', endpoint];
+    const args = ['gateway', 'info', '--gateway-endpoint', endpoint];
```
benoitf: why for status it does now gateway info

packages/main/src/plugin/openshell-cli/openshell-cli.ts
```diff
@@ -489,15 +491,17 @@ export class OpenshellCli {
...
+  private async execCLI<T>(args: string[], options?: RunOptions, quiet = false): Promise<T> {
     const cliPath = this.getCliPath();
     const fullArgs = [...args, '-o', 'json'];
     try {
       const result = await this.exec.exec(cliPath, fullArgs, options);
       return JSON.parse(result.stdout) as T;
     } catch (err: unknown) {
       const detail = this.extractCliError(err);
-      console.error(`openshell failed: ${cliPath} ${fullArgs.join(' ')} — ${detail}`);
+      if (!quiet) {
```
benoitf: should be outside of this PR

packages/main/src/plugin/openshell-cli/openshell-cli.ts
bmahabirbu: I noticed when a gateway isnt reachable it will output an error that gets fed to the console consistently but this can be a logging followup

packages/main/src/plugin/openshell-cli/openshell-cli.ts
bmahabirbu: Originally If i tested an ordinary http server

`python3 -m http.server 18888 --bind 127.0.0.1'

and did openshell status the way kaiden checks it would pass as the endpoint is reachable

but its a edge case and confuses the relationship between status and gateway info will revert. that bug is an upstream one that we shouldn't handle. And we can also guard against checking a random server

extensions/openshell/package.json
```diff
@@ -52,6 +52,13 @@
           "when": "!isWindows",
           "default": "",
           "description": "Custom path to openshell-gateway binary (Default is blank)"
+        },
+        "openshell.gateway.pollInterval": {
+          "type": "number",
+          "scope": "DEFAULT",
+          "default": 5000,
+          "minimum": 1000,
+          "description": "Interval in milliseconds between OpenShell gateway state checks."
```
benoitf: might be in seconds ? (I mean milliseconds are for the technical side but I'm not sure I want to enter milliseconds as a user ?)

packages/api/src/openshell-gateway-info.ts
```diff
@@ -32,6 +32,23 @@ export const GatewayInfoSchema = z.object({
...
+  health: GatewayHealthSchema,
+});
+
+export type GatewayState = z.output<typeof GatewayStateSchema>;
+
+export const GatewayInfoWithStateSchema = GatewayInfoSchema.extend({
+  gatewayState: GatewayStateSchema,
+});
+
+export type GatewayInfoWithState = z.output<typeof GatewayInfoWithStateSchema>;
```
benoitf: I don't see why we have both GatewayInfoWithState and GatewayInfo

I would expect I have only one GatewayInfo object

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -548,8 +551,8 @@ export class AgentWorkspaceManager implements Disposable {
     return results;
   }
 
-  async listOpenshellGateways(): Promise<GatewayInfo[]> {
-    return this.openshellCli.listGateways();
+  async listOpenshellGateways(): Promise<GatewayInfoWithState[]> {
+    return [...this.openshellGatewayStateManager.getGateways()];
```
benoitf: why do we change listGateways into getGateways ?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -699,7 +702,7 @@ export class AgentWorkspaceManager implements Disposable {
       return this.listOpenshellSandboxes();
     });
 
-    this.ipcHandle('agent-workspace:listOpenshellGateways', async (): Promise<GatewayInfo[]> => {
+    this.ipcHandle('agent-workspace:listOpenshellGateways', async (): Promise<GatewayInfoWithState[]> => {
```
benoitf: AFAIK it should not change the return type

just that the state is now part of the info

extensions/openshell/package.json
bmahabirbu: Sounds good will change to seconds

packages/api/src/openshell-gateway-info.ts
bmahabirbu: Ok I will extend GatewayInfo object with state

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: We need state info so we dont pick a gateway that is unheathy or unreachable

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: Right! This will be fixed as just having the state part of the Gatewayinfo object

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: But now since the state is part of info no need for this

packages/main/src/plugin/openshell-cli/openshell-gateway-state-manager.ts
```diff
@@ -0,0 +1,150 @@
...
+          };
+        }
+      }),
+    );
+    const nextGateways = new Map(gateways.map(gateway => [gateway.name, gateway]));
+    if (this.hasChanged(nextGateways)) {
+      this.#gateways = nextGateways;
+      this.#onDidUpdateGateways.fire(this.listGateways());
+    }
+  }
```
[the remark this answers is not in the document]
bmahabirbu: This is a bigger fix than it seems. Per gateway polling complicates things greatly and the information needed will come fairly quickly

MarsKubeX (approved):
LGTM. Tested and checked that the second gateway shows the correct state

PR #2635, opened 2026-07-31
fix(acp): use absolute path for ACP session cwd

benoitf:
## Summary
- Change `cwd: '.'` to `cwd: '/sandbox'` in both `newSession` calls in the ACP session manager (initial connection and reconnect)
- The Claude ACP agent (`claude-agent-acp`) requires an absolute path for `cwd`; `.` causes `Invalid params: cwd must be an absolute path`
- `/sandbox` is the standard sources directory inside all sandboxes

## Test plan
- [ ] Run `pnpm vitest run packages/main/src/plugin/acp/acp-session-manager.spec.ts` — all 15 tests pass


bmahabirbu (approved):
LGTM

PR #2644, opened 2026-08-04
feat(openshell): warn when no gateways are available

### LINKED ISSUE #2528: Handle no available gateways
### Task content

If no gateways are available, we should refect it on the various pages (workspaces, secrets,...)

bmahabirbu:
## Summary

- show a global warning if there are no usable gateways.

Fixes #2528

## Testing

Use openshell gateway remove *name so you dont have any gateways
See the header appear

https://github.com/user-attachments/assets/fa43c09c-6588-4523-8aed-880a1c1354db

### DIFF
--- a/packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
+++ b/packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
@@ -1,0 +1,10 @@
+<script lang="ts">
+import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
+import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+</script>
+
+{#if $openshellGatewaysReady && $openshellGateways.length === 0}
+  <WarningMessage
+    class="shrink-0 px-3 py-2 bg-(--pd-content-card-bg)"
+    error="No Openshell Gateways Found. Please Create One." />
+{/if}
--- a/packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
+++ b/packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
@@ -3,8 +3,8 @@ import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
 import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
 </script>
 
-{#if $openshellGatewaysReady && $openshellGateways.length === 0}
+{#if $openshellGatewaysReady && $openshellGateways.every(gateway => gateway.gatewayState?.reachable !== true)}
   <WarningMessage
     class="shrink-0 px-3 py-2 bg-(--pd-content-card-bg)"
-    error="No Openshell Gateways Found. Please Create One." />
+    error="No usable OpenShell gateways available." />
 {/if}
--- a/packages/renderer/src/lib/ui/NoUsableGatewayWarning.svelte
+++ b/packages/renderer/src/lib/ui/NoUsableGatewayWarning.svelte
@@ -4,7 +4,7 @@ import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-g
 import type { GatewayInfo } from '/@api/openshell-gateway-info';
 
 function noUsableGateways(gateways: GatewayInfo[]): boolean {
-  return gateways.every(gateway => gateway.gatewayState?.reachable !== true);
+  return !gateways.some(gateway => gateway.gatewayState?.reachable === true);
 }
 </script>
 
--- a/packages/renderer/src/lib/ui/NoUsableGatewayWarning.svelte
+++ b/packages/renderer/src/lib/ui/NoUsableGatewayWarning.svelte
@@ -1,9 +1,14 @@
 <script lang="ts">
 import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
 import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+import type { GatewayInfo } from '/@api/openshell-gateway-info';
+
+function noUsableGateways(gateways: GatewayInfo[]): boolean {
+  return gateways.every(gateway => gateway.gatewayState?.reachable !== true);
+}
 </script>
 
-{#if $openshellGatewaysReady && $openshellGateways.every(gateway => gateway.gatewayState?.reachable !== true)}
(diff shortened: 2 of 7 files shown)


packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
+import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+</script>
+
+{#if $openshellGatewaysReady && $openshellGateways.length === 0}
+  <WarningMessage
+    class="shrink-0 px-3 py-2 bg-(--pd-content-card-bg)"
+    error="No Openshell Gateways Found. Please Create One." />
```
fbricon: Should we provide instructions to create a gateway, or better yet, add a button that will run the gateway creation command?

MarsKubeX: I think the correct product name is OpenShell instead of Openshell

packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
+import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+</script>
+
+{#if $openshellGatewaysReady && $openshellGateways.length === 0}
```
MarsKubeX: Question: here you are checking ``$openshellGateways.length === 0``but what if there is just 1 unreachable gateway? Should the warning be displayed too?

bmahabirbu: Thanks I was thinking along similar lines. The issue details "no-available" gateways so really that means no gateways that are usable

I'll update!

packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
bmahabirbu: https://github.com/openkaiden/kaiden/issues/2522 I have this issue which is to create the ui for starting restarting a gateway etc

So maybe I can add a route button to this as a follow up after I implement that issue? Maybe I can create a sub issue for this additional feature

bmahabirbu: https://github.com/openkaiden/kaiden/issues/2647 created a sub issue

bmahabirbu:
@fbricon @MarsKubeX updated!

packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
+import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+</script>
+
+{#if $openshellGatewaysReady && $openshellGateways.every(gateway => gateway.gatewayState?.reachable !== true)}
```
MarsKubeX: Suggestion: maybe you can moved this logic to a function in the script so the template is more readable.

MarsKubeX (approved):
LGTM, thanks for the changes. Just look my suggestion if you think it's valid.

packages/renderer/src/lib/ui/NoOpenshellGatewayWarning.svelte
bmahabirbu: Good suggestions I updated it and renamed it too since the naming was a little too long

packages/renderer/src/stores/openshell-gateways.ts
```diff
@@ -36,9 +36,12 @@ export async function checkForUpdate(eventName: string): Promise<boolean> {
 }
 
 export const openshellGateways: Writable<GatewayInfo[]> = writable([]);
+export const openshellGatewaysReady: Writable<boolean> = writable(false);
 
-const listOpenshellGateways = (): Promise<GatewayInfo[]> => {
-  return window.listOpenshellGateways();
+const listOpenshellGateways = async (): Promise<GatewayInfo[]> => {
+  const gateways = await window.listOpenshellGateways();
+  openshellGatewaysReady.set(true);
```
fbricon: Redundant store notifications on every fetch (openshell-gateways.ts:42)

  openshellGatewaysReady.set(true) runs on every successful listOpenshellGateways call, not just the first. Svelte writables notify subscribers on every .set() even when the value hasn't changed. This
  causes the {#if} in the warning component to re-evaluate on each gateway refresh event.

  Suggested fix — guard the set:
```typescript
  const listOpenshellGateways = async (): Promise<GatewayInfo[]> => {
    const gateways = await window.listOpenshellGateways();
    if (!get(openshellGatewaysReady)) {
      openshellGatewaysReady.set(true);
    }
    return gateways;
  };
```
  This is low-severity — the DOM won't actually change — but it's a cheap fix.

fbricon (changes requested):
Just minor nits

PR description references stale filename — body mentions NoOpenshellGatewayWarning.spec.ts but the actual file is NoUsableGatewayWarning.spec.ts. Cosmetic only.

packages/renderer/src/lib/ui/NoUsableGatewayWarning.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+import WarningMessage from '/@/lib/ui/WarningMessage.svelte';
+import { openshellGateways, openshellGatewaysReady } from '/@/stores/openshell-gateways';
+import type { GatewayInfo } from '/@api/openshell-gateway-info';
+
+function noUsableGateways(gateways: GatewayInfo[]): boolean {
+  return gateways.every(gateway => gateway.gatewayState?.reachable !== true);
```
fbricon: it'll make 0 visible difference in terms of perf here, but finding there's at least 1 reachable gateway is faster

bmahabirbu: good point for readability!

bmahabirbu: Cool just amended my prev commit and updated

fbricon (approved):

PR #2645, opened 2026-08-04
fix(openshell): pin secret operations to gateways

### LINKED ISSUE #2525: Bind secret/providers operations to the proper gateway

Handle multiple gateways for secrets/providers

List: get providers from all gateway and allow filtering to a single gateway
Delete: execute operation on the proper gateway

_No response_

### Additional context

_No response_

bmahabirbu:
## Summary

OpenShell providers are scoped to gateways, but Kaiden previously treated provider names as globally unique. This caused secret-vault entries from different gateways to collide and could send deletion to the wrong gateway.

This change makes the owning gateway part of each secret's identity across the main process, preload bridge, renderer store, and Secret Vault UI. Kaiden now:

- lists providers from every registered OpenShell gateway while retaining gateway ownership
- keeps same-named providers from different gateways as distinct Secret Vault entries
- routes provider lookup and deletion to the gateway that owns the selected secret
- continues listing secrets from reachable gateways when another gateway is unavailable

Fixes #2525

## Try it locally

Use a fake credential so no real token is required. Replace `<gateway>` with a name from `openshell gateway list`.

```bash
openshell provider create \
  -g <gateway> \
  --name kaiden-gateway-smoke \
  --type github \
  --credential api_token=not-a-real-token

openshell provider list -g <gateway> --names
pnpm watch
```

In Kaiden:

1. Open **Secret Vault** and confirm `kaiden-gateway-smoke` is listed.
2. Open or remove that entry.
3. Confirm it was removed from the owning gateway:

```bash
openshell provider list -g <gateway> --names
```

If you have two gateways, create `kaiden-gateway-smoke` on both. Secret Vault should show two distinct entries, and removing either entry should leave the provider on the other gateway intact.

Clean up any remaining test provider with:

```bash
openshell provider delete kaiden-gateway-smoke -g <gateway>
```

### DIFF
--- a/packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
+++ b/packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
@@ -62,8 +62,12 @@ export class OpenshellSecretAdapter implements SecretCliBackend {
     return await this.openshellCli.listProviders(gateway);
   }
 
-  async removeSecret(name: string): Promise<SecretName> {
-    await this.openshellCli.deleteProvider(name);
+  async removeSecret(name: string, gateway?: string): Promise<SecretName> {
+    if (gateway) {
+      await this.openshellCli.deleteProvider(name, gateway);
+    } else {
+      await this.openshellCli.deleteProvider(name);
+    }
     return { name };
   }
 
--- a/packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
+++ b/packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
@@ -63,11 +63,7 @@ export class OpenshellSecretAdapter implements SecretCliBackend {
   }
 
   async removeSecret(name: string, gateway?: string): Promise<SecretName> {
-    if (gateway) {
-      await this.openshellCli.deleteProvider(name, gateway);
-    } else {
-      await this.openshellCli.deleteProvider(name);
-    }
+    await this.openshellCli.deleteProvider(name, gateway);
     return { name };
   }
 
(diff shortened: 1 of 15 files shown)


packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
```diff
@@ -62,8 +62,12 @@ export class OpenshellSecretAdapter implements SecretCliBackend {
     return await this.openshellCli.listProviders(gateway);
   }
 
-  async removeSecret(name: string): Promise<SecretName> {
-    await this.openshellCli.deleteProvider(name);
+  async removeSecret(name: string, gateway?: string): Promise<SecretName> {
+    if (gateway) {
```
MarsKubeX: I think the conditional is not needed here since openshellCli.deleteProvider can handle gateway=undefined. Can be simplified to
```async removeSecret(name: string, gateway?: string): Promise<SecretName> {
    await this.openshellCli.deleteProvider(name, gateway);
    return { name };
}
```

packages/main/src/plugin/secret-manager/secret-manager.ts
```diff
@@ -84,7 +111,18 @@ export class SecretManager {
 
     const expectedName = `${info.providerId}-${info.connection.id}`;
     const secrets = await this.list(gateway);
-    return secrets.find(s => s.name === expectedName);
+    const secret = secrets.find(s => s.name === expectedName);
+    if (!secret) return undefined;
+    return {
```
MarsKubeX: Maybe this can use a destructure pattern like ``const { gateway: _, ...rest } = secret; return rest;``. That would also helps if SecreInfo gains new fields in the future.

MarsKubeX (changes requested):

bmahabirbu:
@coderabbit review

MarsKubeX (approved):
LGTM. Thanks for the changes.

PR #2649, opened 2026-08-05
chore: skip Husky DCO when running workflow

deboer-tim:
The Fullsend code and fix agents are failing due to DCO:
- their instructions say not to use DCO, because agents are not human (the correct behaviour)
- the agent commits, which triggers Husky - which lints, checks commitlint ... and applies DCO
- agent postscript fails because DCO was applied

This change avoids Husky DCO when running in GitHub actions workflows.

Code agent: https://github.com/openkaiden/kaiden/actions/runs/30927076530/job/92052139133
Fix agent: https://github.com/openkaiden/kaiden/actions/runs/30576976374/job/90987636583

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -30,7 +30,7 @@ if test ! -f "$1"; then
   exit 1
 fi
 
-pnpm commitlint --edit "$1"
+pnpm commitlint --edit "$1" || exit $?
 
 # skip DCO for agents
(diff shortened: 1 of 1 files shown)


fbricon (changes requested):
coderabbit's review sounds legit. Should be fine to merged once addressed

fbricon (approved):

PR #2652, opened 2026-08-06
feat(openshell): create local gateways from settings

### LINKED ISSUE #2522: Allow to add a new local gateway

According to the mockup:

https://github.com/openkaiden/mockups/blob/milestone-0.4/settings/index.html

Image

Allow to start a local gateway::

Options are not clear yet but should be:
- driver podman vs vm
- mount enabled or not

_No response_

### Additional context

_No response_

bmahabirbu:
Fixes #2522

## Summary

- add a Settings action for creating an independent local OpenShell gateway with a name and port
- allow users to choose a bind address and port while enabling bind mounts automatically
- keep user-created gateway processes independent from Kaiden's managed `kaiden-local` lifecycle

UPDATED VIDEO

https://github.com/user-attachments/assets/22f56661-9884-49c9-9df3-5e58f7d2b870

## Why this implementation

Openshell has a bug with --local that overrites certs on existing gateways causing errors. So implementation gives each new gateway its own directory with own certs and db!

More specifically: Each gateway also receives isolated PKI. OpenShell's `gateway add --local` copies certificates from the default local TLS directory into the new registration. Reusing that directory would overwrite certificates belonging to another gateway, causing `BadSignature` errors. 

This implementation generates certificates with an isolated `XDG_CONFIG_HOME`, starts the server with explicit TLS paths, and installs the matching client certificates into only the new gateway's registration after `--local` completes.

## Local testing

1. Ensure OpenShell and a local compute driver such as Podman are installed.

2. Check for existing gateway processes and registrations:

   ```sh
   openshell gateway list
   ps ax -o pid=,command= | grep '[o]penshell-gateway'
   ```

3. Stop any gateway process using the port you plan to test. For example, for port 17675:

   ```sh
   lsof -nP -iTCP:17675 -sTCP:LISTEN
   kill -TERM <PID>
   ```

4. Remove a stale registration with the same test name, if present:

   ```sh
   openshell gateway remove local-gateway
   ```

5. Start Kaiden:

   ```sh
   pnpm watch
   ```

6. Open **Settings → Gateways** and click **Create local gateway** beneath the gateway cards.

7. Enter a unique name, bind address, and available port. The defaults are `127.0.0.1` and `17675`. Confirm that an occupied port is rejected before registration.

8. Create the gateway and verify it is healthy over mTLS:

   ```sh
   openshell gateway list
   openshell gateway info -g local-gateway
   ```

9. Confirm its files are isolated and bind mounts are enabled in the generated config:

    ```sh
    ls -la ~/.local/share/kaiden/openshell-gateways/local-gateway
    ls -la ~/.config/openshell/gateways/local-gateway/mtls
    grep enable_bind_mounts ~/.local/share/kaiden/openshell-gateways/local-gateway/gateway.toml
    ```

10. Quit Kaiden and verify the user-created gateway process remains running:

    ```sh
    openshell gateway info -g local-gateway
    ps ax -o pid=,command= | grep '[o]penshell-gateway'
    ```

11. Until UI removal is implemented, stop the server process and remove its CLI registration manually:

    ```sh
    lsof -nP -iTCP:17675 -sTCP:LISTEN
    kill -TERM <PID>
    openshell gateway remove local-gateway
    ```

## Verification

- 182 main-process tests passed
- 17 renderer tests passed
- main and renderer typechecks passed
- Svelte check completed with 0 errors


packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -225,7 +317,14 @@ export class OpenshellGateway implements Disposable {
...
-      await this.registerWithCli();
+      try {
+        await this.registerWithCli();
+      } catch (err: unknown) {
+        await this.stop();
+        this.#port = previousPort;
+        this.#bindAddress = previousBindAddress;
+        throw err;
+      }
     }
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: managed registration failures now propagate so start() performs its rollback; the orphan-gateway init path explicitly retains best-effort handling.

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
```diff
@@ -0,0 +1,181 @@
...
+
+        <label for="gateway-port" class="block mt-5 mb-2 text-sm font-semibold">Port</label>
+        <NumberInput
+          name="gateway-port"
+          aria-label="Gateway port"
+          bind:value={port}
+          minimum={1}
+          maximum={65535}
+          type="integer"
+          showError={false} />
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: the NumberInput is now nested inside its Port label, associating the rendered input without changing the shared component API.

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
```diff
@@ -0,0 +1,181 @@
...
+  name = value;
+  if (configName !== value.trim()) {
+    config = '';
+    configName = '';
+  }
+  clearTimeout(configLoadTimeout);
+  configLoadTimeout = setTimeout(() => {
+    loadConfig().catch(() => {});
+  }, 250);
+}
```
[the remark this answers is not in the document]
bmahabirbu: Superseded in 8f723b68c: config preloading/debouncing was removed with the Config tab. Config is generated only after Create is clicked.

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
```diff
@@ -0,0 +1,181 @@
...
+    return 'Use lowercase letters, numbers, dots, dashes, or underscores';
+  }
+  if (existingNames.includes(name.trim())) return 'A gateway with this name already exists';
+  return '';
+});
+let portError = $derived(
+  !Number.isInteger(port) || port < 1 || port > 65_535 ? 'Enter a port between 1 and 65535' : '',
+);
+let bindMountsEnabled = $derived(getBindMountsEnabled(config));
+let canCreate = $derived(!nameError && !portError && config.trim() !== '' && !loadingConfig && !creating);
```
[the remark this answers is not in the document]
bmahabirbu: Partially superseded by the scope reduction; the driver regex was removed with the Config tab. The still-relevant gateway-name pattern is now shared from the API package in 8f723b68c.

packages/renderer/src/lib/preferences/openshell-gateway-config.ts
```diff
@@ -0,0 +1,20 @@
...
+export function setBindMountsEnabled(config: string, enabled: boolean): string {
+  const value = `enable_bind_mounts = ${enabled}`;
+  if (BIND_MOUNTS_SETTING.test(config)) {
+    return config.replace(BIND_MOUNTS_SETTING, value);
+  }
+
+  const section = DRIVER_SECTION.exec(config);
+  if (!section?.index && section?.index !== 0) {
+    return config;
+  }
```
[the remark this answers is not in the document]
bmahabirbu: Superseded in 8f723b68c: the Config tab and openshell-gateway-config helper were removed to reduce this PR's scope.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -425,4 +534,64 @@ export class OpenshellGateway implements Disposable {
...
+      );
+    }
+  }
+
+  private getGatewayStorageDirectory(name: string): string {
+    if (name === DEFAULT_GATEWAY_NAME) {
+      return join(this.directories.getDataDirectory(), 'openshell-gateway');
+    }
+    return join(this.directories.getDataDirectory(), 'openshell-gateways', name);
+  }
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: kaiden-local is reserved for the Kaiden-managed gateway and rejected by user-created gateway creation.

packages/preload/src/index.ts
```diff
@@ -398,6 +403,17 @@ export function initExposure(): void {
...
+  contextBridge.exposeInMainWorld('getLocalGatewayConfig', async (name: string): Promise<string> => {
+    return ipcInvoke('agent-workspace:getLocalGatewayConfig', name);
+  });
+
+  contextBridge.exposeInMainWorld(
+    'createLocalGateway',
+    async (options: CreateLocalGatewayOptions): Promise<GatewayInfo[]> => {
+      return ipcInvoke('agent-workspace:createLocalGateway', options);
+    },
+  );
```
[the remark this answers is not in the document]
bmahabirbu: Verified as already satisfied: build:preload:types generates both methods in packages/preload/exposedInMainWorld.d.ts, and renderer typecheck passes.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -425,4 +534,64 @@ export class OpenshellGateway implements Disposable {
...
+    await Promise.all([
+      copyFile(join(storageDirectory, 'ca.crt'), join(mtlsDirectory, 'ca.crt')),
+      copyFile(join(storageDirectory, 'client', 'tls.crt'), join(mtlsDirectory, 'tls.crt')),
+      copyFile(join(storageDirectory, 'client', 'tls.key'), join(mtlsDirectory, 'tls.key')),
+    ]);
+  }
+
+  private async waitForIndependentGateway(process: ChildProcess, name: string): Promise<void> {
+    let spawnError: Error | undefined;
+    process.once('error', err => (spawnError = err));
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: renamed the parameter to gatewayProcess.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -425,4 +534,64 @@ export class OpenshellGateway implements Disposable {
...
+    storageDirectory: string,
+    config: string,
+    isolatePki = false,
+  ): Promise<string> {
+    const configPath = join(storageDirectory, 'gateway.toml');
+    await this.generateCerts(binaryPath, storageDirectory, isolatePki);
+    await mkdir(storageDirectory, { recursive: true });
+    await writeFile(configPath, config, 'utf-8');
+    return configPath;
+  }
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: the storage directory is created before certificate generation and the redundant creation call was removed.

packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
```diff
@@ -270,6 +277,113 @@ describe('init', () => {
...
+    );
+    expect(openshellCli.addGateway).toHaveBeenCalledWith({
+      endpoint: 'https://127.0.0.1:17675',
+      local: true,
+      name: 'local-dev',
+    });
+    const mtlsDirectory = join(homedir(), '.config', 'openshell', 'gateways', 'local-dev', 'mtls');
+    expect(copyFile).toHaveBeenCalledWith(join(storageDirectory, 'ca.crt'), join(mtlsDirectory, 'ca.crt'));
+    expect(copyFile).toHaveBeenCalledWith(join(storageDirectory, 'client', 'tls.crt'), join(mtlsDirectory, 'tls.crt'));
+    expect(copyFile).toHaveBeenCalledWith(join(storageDirectory, 'client', 'tls.key'), join(mtlsDirectory, 'tls.key'));
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: the test pins XDG_CONFIG_HOME and builds the expected mTLS path from that deterministic value.

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -724,6 +724,18 @@ export class AgentWorkspaceManager implements Disposable {
...
+      return this.openshellGateway.getLocalGatewayConfig(name);
+    });
+
+    this.ipcHandle(
+      'agent-workspace:createLocalGateway',
+      async (_listener, options: CreateLocalGatewayOptions): Promise<GatewayInfo[]> => {
+        await this.openshellGateway.createLocalGateway(options);
+        return this.listOpenshellGateways();
+      },
+    );
```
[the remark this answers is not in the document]
bmahabirbu: Fixed in 8f723b68c: the creation IPC now awaits a gateway-state refresh before returning the list.

jeffmaury (commented):
While testing, I restarted Kaiden so the new gateway keeps listed but we cannot not start/stop it. Do we have a follow up issue ?

bmahabirbu:
No, I can create a follow up issue for kaiden to handle created gateway lifecycle 

It was originally part of this pr but I thought the scope was too large

bmahabirbu:
See https://github.com/openkaiden/kaiden/issues/2664

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
```diff
@@ -0,0 +1,127 @@
...
+let portAvailabilityError = $state('');
+
+let nameError = $derived.by((): string => {
+  if (!name.trim()) return 'Enter a gateway name';
+  if (!GATEWAY_NAME_PATTERN.test(name.trim())) {
+    return 'Use lowercase letters, numbers, dots, dashes, or underscores';
+  }
+  if (existingNames.includes(name.trim())) return 'A gateway with this name already exists';
+  return '';
+});
```
MarsKubeX, line 36: Question: I was wondering if it need to have a conditional to protect against creating a "kaiden-local" gateway. Normally if "kaiden-local" is registered it should go through the conditional of existingNames, but I wonder what if it's not registered or if it not exist, you create manually a "kaiden-local" gateway, and then maybe the app try to create also a "kaiden-local" gateway. It's really a edge case so it's not blocking.

MarsKubeX (approved):
LGTM. I have tested it, and I was able to create a new gateway.

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
bmahabirbu: Good point I will investigate the potential conflict between our kaiden-local vs if a user created that gateway with the same name

bmahabirbu:
Given the new epic to switch to the vm driver i will add a new no op selector to select the vm driver for gateway creation

packages/renderer/src/lib/preferences/PreferencesOpenshellGatewayCreate.svelte
bmahabirbu: right i think the easiest approach is to just remove the ability for the user to create a gateway with "kaiden-local"

bmahabirbu: Done! @MarsKubeX please take another look also updated the video

bmahabirbu:
mtls for gateway creation is not specifically needed here as well will remove and keep it plaintext like kaiden-local. Can file another issue to bring back the implementation at another time

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -155,6 +166,76 @@ export class OpenshellGateway implements Disposable {
...
+  async createLocalGateway(options: CreateLocalGatewayOptions): Promise<void> {
+    if (options.driver && !['vm', 'podman', 'docker'].includes(options.driver)) {
+      throw new Error(`Unsupported local gateway driver: ${String(options.driver)}`);
+    }
+    const driver = options.driver ?? (await this.detectLocalComputeDriver()) ?? 'podman';
+    if (driver === 'vm') {
+      throw new Error('VM support soon');
+    }
+    await this.createContainerGateway(options, driver);
+  }
```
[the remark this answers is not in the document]
bmahabirbu: I wanted the vm to be a no op by design its by no means a fully functioning feature yet so i will dismiss this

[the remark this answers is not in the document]
bmahabirbu: The problem with the error is the create button will still be active we dont want that so thats why i coded it that way

bmahabirbu:
@coderabbit resolve

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -155,6 +166,76 @@ export class OpenshellGateway implements Disposable {
     return tool?.path;
   }
 
+  async createLocalGateway(options: CreateLocalGatewayOptions): Promise<void> {
+    if (options.driver && !['vm', 'podman', 'docker'].includes(options.driver)) {
```
jeffmaury: **question:** why do we have this test as the type is an enumeration ?

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -155,6 +166,76 @@ export class OpenshellGateway implements Disposable {
...
+    }
+    await isFreePort(options.port);
+
+    const binaryPath = this.getGatewayBinaryPath();
+    if (!binaryPath) {
+      throw new Error('openshell-gateway binary not registered in CLI tool registry');
+    }
+    const storageDirectory = this.getGatewayStorageDirectory(name);
+    const configPath = await this.createNamedGatewayConfig(binaryPath, storageDirectory, driver);
+    const logFd = openSync(join(storageDirectory, GATEWAY_LOG_FILENAME), 'w');
```
jeffmaury: **suggestion:** should use the promise version of open

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -425,4 +517,73 @@ export class OpenshellGateway implements Disposable {
...
+        'Gateway name must start with a lowercase letter or number and contain only lowercase letters, numbers, dots, dashes, or underscores',
+      );
+    }
+    if (!allowDefault && name === DEFAULT_GATEWAY_NAME) {
+      throw new Error(`"${DEFAULT_GATEWAY_NAME}" is reserved for the gateway managed by Kaiden`);
+    }
+  }
+
+  private getGatewayStorageDirectory(name: string): string {
+    if (name === DEFAULT_GATEWAY_NAME) {
```
jeffmaury: **issue:** we should not have a special case for the default gateway

jeffmaury (changes requested):

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: This check just made sure an unsupported driver didn't reach the gateway config. But since the driver is constrained by localgatewaydriver than this is less of a valid check will remove

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Makes sense all gateway configs should live in the same place for creation

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: thanks updated using node:fs/promises

MarsKubeX (commented):
LGTM in general. I've tested and it works. I'm only not sure if the pr should add the vm option ion the form since it's not supported yet, it feels like a UX dead end.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -155,6 +166,73 @@ export class OpenshellGateway implements Disposable {
...
+        this.buildArgs(true, configPath, storageDirectory, options.port, bindAddress),
+        {
+          stdio: ['ignore', logFile.fd, logFile.fd],
+          detached: true,
+        },
+      );
+      gatewayProcess.once('error', err => (processState.spawnError = err));
+    } finally {
+      await logFile.close();
+    }
```
MarsKubeX, line 214: nit: if the spawn throws (that is rarely to occur), gatewayProcess is uninitialized, so it can led to errors where is used after.

bmahabirbu: Good catch will update

bmahabirbu:
Thanks! That's a fair take I will remove the vm selector till we have full support and can be added back later. I can create a follow-up issue once the initial vm groundwork is done

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: k updated!

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -323,22 +425,18 @@ export class OpenshellGateway implements Disposable {
...
+      const storageDirectory = this.getGatewayStorageDirectory(DEFAULT_GATEWAY_NAME);
       const configPath = join(storageDirectory, 'gateway.toml');
+      await mkdir(storageDirectory, { recursive: true });
       await this.generateCerts(binaryPath, storageDirectory);
       const config = Mustache.render(gatewayConfigTemplate, {
         supervisorImage: image,
         gatewayDir: storageDirectory,
         q: '"',
-        driver,
+        driver: driver === 'podman' || driver === 'docker' ? driver : undefined,
```
jeffmaury, line 436: **question:** why this check where this is an enumeration ?

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -45,8 +52,9 @@ const MAX_HEALTH_CHECK_ATTEMPTS = 30;
 const STOP_TIMEOUT_MS = 5000;
 const SUPERVISOR_IMAGE_BASE = 'ghcr.io/nvidia/openshell/supervisor';
 const GATEWAY_LOG_FILENAME = 'gateway.log';
+const DEFAULT_GATEWAY_NAME = KAIDEN_LOCAL_GATEWAY_NAME;
 
-type LocalComputeDriver = 'docker' | 'podman';
+type LocalComputeDriver = Exclude<LocalGatewayDriver, 'vm'>;
```
jeffmaury, line 57: **question:** why are we excluding vm driver ?

jeffmaury (commented):
Also seems that every gateway that is started or not by Kaiden is displayed as Managed

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: Right that is too aggressive I added it because the vm driver isnt implemented yet but we simply can just remove this.

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
bmahabirbu: removed

bmahabirbu:
Changed so gateway created says referenced instead of managed

jeffmaury (approved):

PR #2656, opened 2026-08-06
fix(build): kill all spawned vite processes when pnpm watch exits

fbricon:
- Track all child processes and install SIGINT/SIGTERM/SIGQUIT handlers
  to propagate termination to every spawned process tree
- Spawn children with detached: true (non-Windows) so process-group
  kill via negative pid cleans up entire subtrees (pnpm → vite)
- Re-raise the original signal instead of process.exit(0) so the
  parent shell sees the correct exit status
- Use taskkill /T on Windows for proper process-tree teardown

I was experiencing frequent OOM issues because of dangling vite processes, accumulating after each `pnpm watch`. Now all those are cleaned up on exit.

### DIFF
--- a/scripts/watch.mjs
+++ b/scripts/watch.mjs
@@ -20,7 +20,7 @@
 
 import { createServer, build, createLogger } from 'vite';
 import electronPath from 'electron';
-import { spawn } from 'child_process';
+import { spawn, spawnSync } from 'node:child_process';
 import { generateAsync } from 'dts-for-context-bridge';
 import { dirname, resolve } from 'path';
 import { fileURLToPath } from 'url';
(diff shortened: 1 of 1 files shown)


benoitf (commented):
FYI: needs to be committed first in upstream

scripts/watch.mjs
```diff
@@ -29,6 +29,70 @@ import { delimiter, join } from 'node:path';
...
+ */
+const childrenProcesses = new Set();
+
+/**
+ * Track a child process so it can be killed together with the watch script.
+ * @param {import('child_process').ChildProcess} child
+ */
+function trackChildProcess(child) {
+  childrenProcesses.add(child);
+  child.once('exit', () => childrenProcesses.delete(child));
```
bmahabirbu, line 44: Something I noticed is that exited child’s remaining process tree is skipped Keep the PID/process group tracked until cleanup completes or else some processes might stick. 

2 places I will comment to adjust 

Something like 

// Register cleanup first so the process group remains tracked while
  // cleanupAndExit terminates any surviving descendants.
```
  spawnProcess.on('exit', cleanupAndExit);
  trackChildProcess(spawnProcess);
```

So switch the ordering

scripts/watch.mjs
```diff
@@ -114,8 +187,11 @@ const setupMainPackageWatcher = ({ config: { server, extensions } }) => {
         logger.error(data, { timestamp: true });
       });
 
+      // Track the Electron process so it's cleaned up when the watch script exits
+      trackChildProcess(spawnProcess);
```
bmahabirbu: See above switch ordering to 
```
  spawnProcess.on('exit', cleanupAndExit);
  trackChildProcess(spawnProcess);
```

scripts/watch.mjs
```diff
@@ -179,8 +255,13 @@ const setupExtensionApiWatcher = name => {
     console.error(data, { timestamp: true });
   });
 
+  // Track the child so it's killed together with the watch script, and make
+  // sure stdout/stderr keep flowing while we're detached.
+  trackChildProcess(spawnProcess);
```
bmahabirbu, line 264: See above switch ordering to 
```
  spawnProcess.on('exit', cleanupAndExit);
  trackChildProcess(spawnProcess);
```

bmahabirbu (commented):
Reviewing this pr as if it were upstream so overall LGTM 

See comments below

You can address the changes in the podman desktop pr

scripts/watch.mjs
```diff
@@ -91,8 +155,16 @@ const setupMainPackageWatcher = ({ config: { server, extensions } }) => {
...
+        spawnProcess.off('exit', cleanupAndExit);
+        if (spawnProcess.pid !== undefined && process.platform !== 'win32') {
+          try {
+            process.kill(-spawnProcess.pid, 'SIGINT');
+          } catch {
+            spawnProcess.kill('SIGINT');
+          }
+        } else {
+          spawnProcess.kill('SIGINT');
+        }
```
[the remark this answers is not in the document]
fbricon: Extracted killChild() helper, used it in the Electron restart path instead of duplicating logic (didn't use killChildren() directly since that would kill extension watchers too)

fbricon:
I'll let @bmahabirbu bring it into podman desktop, as I'm going on PTO

fbricon:
Now the commit is the one backported from Podman Desktop, minus the "setupUiPackageWatcher" bits that don't exist in Kaiden

fbricon:
@coderabbitai resolve
no more changes in this PR, everything must be done upstream first

benoitf (approved):

PR #2668, opened 2026-08-11
fix(milvus): preserve class names through minification

### LINKED ISSUE #2667: Milvus indexing fails in production builds: "Unable to load service: milvus.proto.milvus.MilvusService"
### Bug description

Uploading a file to a Milvus-backed knowledge base always fails, but only in packaged/production builds (installer builds, or anything built via `pnpm compile:current`). Dev mode (`pnpm watch`) works fine.

The file gets stuck in `error` status with this toast:

```
Error: Unable to load service: milvus.proto.milvus.MilvusService
```

**Cause:** production builds minify the milvus extension's bundled code, and the minifier renames a class that the Milvus SDK checks by name at runtime. That check then always fails, so every indexing attempt errors out immediately, regardless of file type or LLM provider.

### Operating system

All platforms (confirmed on Linux CI; cause is a JS build setting, not OS-specific)

### Installation Method

### Version

next (development version)

### Steps to reproduce

1. Build or install a production Kaiden build.
2. Create a knowledge base using Milvus.
3. Upload any file.
4. File status goes to `error` with the message above.

### Relevant log output

```shell
Error Indexing /path/to/file on <knowledge-base-name>
Error: Unable to load service: milvus.proto.milvus.MilvusService
```

serbangeorge-m:
Fixes #2667

## Summary
- Production builds minify the milvus extension's bundled code, and esbuild renames a class (`ServiceClientImpl`, dynamically declared by `@grpc/grpc-js`) that `@zilliz/milvus2-sdk-node` checks by name (`Function.prototype.name`) at runtime
- That check always fails after minification, so every Milvus indexing attempt errors immediately in packaged builds, regardless of file type or LLM provider
- Adds `esbuild: { keepNames: true }` to `extensions/milvus/vite.config.js`, which makes esbuild emit a `__name()` call restoring the original class name after minification

## Test plan
- [x] Verified the built bundle contains the preserved class name (`static { __name(this, "ServiceClientImpl") }`) instead of a mangled one
- [x] `pnpm --filter milvus test` passes (34/34)

### DIFF
--- a/extensions/milvus/vite.config.js
+++ b/extensions/milvus/vite.config.js
@@ -34,6 +34,15 @@ const config = {
       '/@/': join(PACKAGE_ROOT, 'src') + '/',
     },
   },
+  // @grpc/grpc-js (bundled here, like every other extension dependency, since the
+  // packaged app only ships extensions/**/dist and never node_modules) dynamically
+  // declares `class ServiceClientImpl` and relies on its Function.prototype.name
+  // matching that identifier at runtime. esbuild's production-only minification
+  // renames the class and breaks that check inside milvus2-sdk-node, so names must
+  // be preserved through minification instead.
+  esbuild: {
+    keepNames: true,
+  },
   build: {
     sourcemap: 'inline',
     target: 'esnext',


bmahabirbu (approved):
LGMT codewise

PR #2699, opened 2026-08-17
fix(agent-workspaces): reset stale wizard draft on onboarding restart

### LINKED ISSUE #2691: Create Workspace wizard shows stale default model after "Onboard again" if a draft was already open
**Summary**
After using "Onboard again" (Settings → Preferences → Onboarding) to re-run guided setup and pick a different default model, the Create Workspace wizard can keep showing the old model instead of the new one.

**Steps to reproduce**
1. Complete guided setup, picking a specific model for an agent (e.g. OpenCode).
2. Open Workspaces → Create workspace, go to the Agent & Model step. Note the model shown (matches step 1).
3. Navigate away without cancelling or creating the workspace.
4. Go to Settings → Preferences → Onboarding → click "Onboard again."
5. Complete guided setup again, picking a *different* model.
6. Open Workspaces → Create workspace → Agent & Model step again.

**Expected:** the newly-selected model from step 5 is shown as the default.
**Actual:** the old model from step 1 is still shown.

**Workaround:** clicking Cancel on the create-workspace wizard (or successfully creating a workspace) before re-visiting fixes it.

**Root cause**
`wizard.draft` in `packages/renderer/src/stores/agent-workspace-create-draft.svelte.ts` is a module-level singleton with an `initialized` flag. `AgentWorkspaceCreate.svelte` (`onMount`, lines ~179-211) only applies the freshly-fetched `onboarding.defaultWorkspaceSettings` into the draft when `initialized` is still `false`. Once set on a first visit, it stays `true` for the rest of the app session and is only reset by `resetDraft()` — called on Cancel or successful creation, never on onboarding reset. So a live, unfinished draft keeps its stale agent/model even though the underlying persisted config was correctly updated.

**Suggested fix direction**
Either invalidate `wizard.draft.initialized` when `onboarding:restart` fires, or have `AgentWorkspaceCreate.svelte` re-apply `defaultWorkspaceSettings` whenever it changes, not just on first mount.

**Files involved**
- `packages/renderer/src/lib/agent-workspaces/AgentWorkspaceCreate.svelte:179-211`
- `packages/renderer/src/stores/agent-workspace-create-draft.svelte.ts:85-96`

serbangeorge-m:
## Summary
- The Create Workspace wizard's draft only re-applies onboarding defaults on first mount (`wizard.draft.initialized`), which is reset by `resetDraft()` on Cancel/create — but never on "Onboard again". A live, unfinished draft kept its stale default agent/model after re-running guided setup with different picks.
- Reset `wizard.draft.initialized` on the existing `onboarding:restart` event so the next mount re-applies the fresh defaults.

Fixes #2691

## Test plan
- [x] Added `agent-workspace-create-draft.onboarding-restart.spec.ts` covering the reset behavior
- [x] `pnpm run test:unit`, `pnpm run typecheck`, `pnpm run lint:check` pass

## Manual test steps

1. Launch Kaiden fresh (or use "Onboard again" if already set up) and complete guided setup, picking a specific model for an agent (e.g. OpenCode).
2. Go to Workspaces → Create workspace, navigate to the Agent & Model step. Confirm it shows the model from step 1.
3. Navigate away (e.g. click Chat or any other nav item) — do not click Cancel and do not create the workspace.
4. Go to Settings → Preferences → Onboarding → Onboard again.
5. Complete guided setup again, picking a different model for the same agent.
6. Go back to Workspaces → Create workspace → Agent & Model step again.

Expected (with the fix): the wizard shows the model you picked in step 5.
Before the fix: it would still show the stale model from step 1.

### DIFF
--- a/packages/renderer/src/stores/agent-workspace-create-draft.onboarding-restart.spec.ts
+++ b/packages/renderer/src/stores/agent-workspace-create-draft.onboarding-restart.spec.ts
@@ -1,0 +1,58 @@
+/**********************************************************************
+ * Copyright (C) 2026 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ***********************************************************************/
+
+import { beforeEach, expect, test, vi } from 'vitest';
+
+const callbacks = new Map<string, () => Promise<void> | void>();
+
+beforeEach(() => {
+  callbacks.clear();
+  vi.resetAllMocks();
+  vi.resetModules();
+  Object.defineProperty(window, 'events', {
+    value: {
+      receive: vi.fn((channel: string, callback: () => Promise<void> | void) => {
+        callbacks.set(channel, callback);
+      }),
+    },
+    configurable: true,
+  });
+});
+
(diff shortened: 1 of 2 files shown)


bmahabirbu (commented):
Overall LGTM

One super small edge case is if re-running onboarding clears the default or picks an unavailable model, the old model sticks

bmahabirbu (approved):

PR #2711, opened 2026-08-20
feat(vllm-semantic-router): add extension for vLLM semantic router containers

### LINKED ISSUE #2420: Create a new vllm-semantic-router extension

We need an implementation of the semantic router factory api

Implemented the factory based on:
- use the container extension api to start the vllm-semantic router
- use the ghcr.io/vllm-project/semantic-router/vllm-sr:v0.3.0 image when starting
- bind the config file to /app/config.yaml
- bind the port from the confguration file to a random local port
- once the container is started, register an inference connection factory (using AI SKD openai) with local URL localhost:allocated_port
- the container is labelled with ai.openkaiden.vllm-semantic-router.name and  ai.openkaiden.vllm-semantic-router.id (uuid of the connection) so we can restore the connection when starting the extension

_No response_

### Additional context

_No response_

jeffmaury:
Introduce a new extension that manages dual-container deployments (vLLM semantic router + Envoy sidecar) with container lifecycle, config generation, Zod schemas for both Envoy and SR configs, and existing container discovery on create.

To test:
- create a connection to a remote OpenAI endpoint and model (I used LiteMaas)
- create a connection to a local model (I used ollama with Qwen3.6)
- create a semantic router with a keyword for java and route it to the remote model, use the local one as the default model
- Go to the settings page and start the Semantic Router connection
- With Podman Desktop verify the 2 containers (envoy and semantic router have been started)
- Create a new workspace using the virtual model (Semantic Router name) and ask a question containing java

https://github.com/user-attachments/assets/ef27b37e-8fab-4bd8-8e0c-35e15981fc44

Fixes #2420


extensions/vllm-semantic-router/src/manager/semantic-router-container-manager.ts
```diff
@@ -0,0 +1,591 @@
...
+                      socket_address: { address: host.split('/')[0], port_value: port },
+                    },
+                    hostname: host.split('/')[0],
+                  },
+                },
+              ],
+            },
+          ],
+        },
+        transport_socket: {
```
bmahabirbu, line 520: Here I think the TLS configurations should be conditional 

 - https backend such as LiteMaaS → include transport_socket
  - http backend such as RamaLama/Ollama omit transport_socket

Since the Envoy schema already makes transport_socket optional

bmahabirbu (commented):
Also i think when creating a semantic router it should automatically start the vllm semantic router service u shouldnt need to go into resources to start it

extensions/vllm-semantic-router/src/manager/semantic-router-container-manager.ts
```diff
@@ -0,0 +1,591 @@
...
+      },
+      Env: [
+        'SR_LOG_LEVEL=debug',
+        'SR_LOG_ENCODING=console',
+        'OTEL_TRACES_EXPORTER=console',
+        'OTEL_SERVICE_NAME=semantic-router',
+      ],
+      HostConfig: {
+        PortBindings: {
+          [`${ROUTER_GRPC_PORT}/tcp`]: [{ HostPort: String(routerGrpcHostPort) }],
```
bmahabirbu, line 139: Ned to add ro,Z for rootless podman

Binds: [`${configFilePath}:/app/config.yaml:ro,Z`],

for it to work or i get permission denied

extensions/vllm-semantic-router/src/manager/semantic-router-container-manager.ts
```diff
@@ -0,0 +1,591 @@
...
+      User: '101',
+      Cmd: ['envoy', '-c', '/etc/envoy/envoy.yaml', '-l', 'debug'],
+      Labels: {
+        [LABEL_NAME]: params.name,
+        [LABEL_ID]: connectionId,
+        [LABEL_ROLE]: 'envoy',
+      },
+      ExposedPorts: {
+        [`${ENVOY_LISTEN_PORT}/tcp`]: {},
+      },
```
bmahabirbu, line 172: Need to add ro, Z 

 Binds: [`${envoyConfigFilePath}:/etc/envoy/envoy.yaml:ro,Z`],

for rootless podman to work or i get permission denied

bmahabirbu (commented):
I also could not create a semantic router on linux

  - Semantic Router: open /app/config.yaml: permission denied
  - Envoy: unable to read file: /etc/envoy/envoy.yaml

See comments

bmahabirbu (commented):
I was able to test this feature end to end on mac and linux using two cloud models ie gemini light for normal and the pro for keyword and it worked

The problems comes from trying to test ollama and by extension ramalama. I think whats happening is the vllm router container gets the normal url and not the openshell internal url so it cant actually route to ollama or ramalama and it doesnt work for me

@jeffmaury how did u startup ollama did u enable loopback?

bmahabirbu (commented):
@jeffmaury I noticed the vllm semantic router restart button in resources only stops the container and doesnt start it back

it has to do with the code in PreferenceResourceRendering in provider.inferenceConnections on line 171 onward because there is a bug because it doesn't check the status of an inference connection. Its a small fix

Since the bug is outside this pr that can be solved via a follow up but wanted to note it

bmahabirbu:
I also still wasnt able to test via ollama on mac

bmahabirbu (approved):
Image

It works!! Ramalama qwen3.5:9b as the default with ollama qwen2.5 as the keyword model

PR #2712, opened 2026-08-20
feat(openshell): add @nvidia/openshell-sdk and cached mTLS-aware client factory

### LINKED ISSUE #2708: Add @nvidia/openshell-sdk and a cached mTLS-aware OpenShell client
### Task content

Add `@nvidia/openshell-sdk` (published on GitHub Packages with OpenShell v0.0.106) and a small factory that Kaiden can use for later RPC migrations. **Do not change any `OpenshellCli` method behavior in this PR.** Gateway add/list/select stay on the CLI.

MarsKubeX:
Adds the SDK dependency from GitHub Packages and a small factory (OpenshellSdkClient) that later migration slices will use for gateway RPCs. No existing OpenshellCli behavior is changed.

## Test plan

- [ ] Run `export GITHUB_TOKEN=$(gh auth token) && pnpm install` on a clean checkout — 
      `@nvidia/openshell-sdk` resolves from GitHub Packages without errors
- [ ] Run `pnpm build:main` — the main process builds successfully with the new dependency
- [ ] Run `pnpm test:unit` — all existing tests pass, plus 14 new tests for the SDK client factory
- [ ] CI checks pass — confirms `GITHUB_TOKEN` + `packages: read` is sufficient 
      for cross-org GitHub Packages auth in workflows

Closes #2708

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -27,6 +27,7 @@ concurrency:
 
 permissions:
   contents: read
+  packages: read
 
 jobs:
   windows:
--- a/README.md
+++ b/README.md
@@ -65,11 +65,14 @@ git clone https://github.com/openkaiden/kaiden && cd kaiden
 
 ### Step 2. Install dependencies
 
-Fetch all dependencies using the command `pnpm`:
+Kaiden depends on `@nvidia/openshell-sdk` which is published to GitHub Packages. A GitHub token with `read:packages` scope is required for `pnpm install` to succeed:
 
 ```sh
+export GITHUB_TOKEN=$(gh auth token)
 pnpm install
 ```
+
+If `gh auth token` does not include `read:packages`, run `gh auth refresh -s read:packages` first.
 
 ### Step 3. Start in watch mode
 
--- a/packages/main/src/plugin/openshell-cli/openshell-sdk-client-manager.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-sdk-client-manager.ts
@@ -31,7 +31,7 @@ import type { GatewayInfo } from '/@api/openshell-gateway-info.js';
  * Clients are lazy — no network request is made until the first RPC.
  */
 @injectable()
-export class OpenshellSdkClient {
+export class OpenshellSdkClientManager {
   readonly #cache = new Map<string, OpenShellClient>();
 
   constructor(
--- a/packages/main/src/plugin/openshell-cli/openshell-sdk-client.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-sdk-client.ts
@@ -16,75 +16,17 @@
  * SPDX-License-Identifier: Apache-2.0
  ***********************************************************************/
 
-import { readFile } from 'node:fs/promises';
-import { homedir } from 'node:os';
-import { join } from 'node:path';
-
-import type { ConnectOptions, OpenShellClient } from '@nvidia/openshell-sdk';
+import type { OpenShellClient } from '@nvidia/openshell-sdk';
 import { inject, injectable, preDestroy } from 'inversify';
 
 import { OpenshellCli } from '/@/plugin/openshell-cli/openshell-cli.js';
+import { OpenshellGatewayConfig } from '/@/plugin/openshell-cli/openshell-gateway-config.js';
 import type { GatewayInfo } from '/@api/openshell-gateway-info.js';
 
 /**
- * Resolve the OpenShell config root following the same logic as the Rust CLI:
- *   1. $XDG_CONFIG_HOME (all platforms, including Windows)
- *   2. %APPDATA% on win32
- *   3. $HOME/.config
- */
-function openshellConfigRoot(): string {
-  const xdg = process.env['XDG_CONFIG_HOME'];
-  if (xdg) return xdg;
-
-  if (process.platform === 'win32') {
-    const appdata = process.env['APPDATA'];
-    if (appdata) return appdata;
-  }
-
-  return join(homedir(), '.config');
-}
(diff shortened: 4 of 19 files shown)


.github/workflows/pr-check.yaml
```diff
@@ -27,6 +27,7 @@ concurrency:
 
 permissions:
   contents: read
+  packages: read
```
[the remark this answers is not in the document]
MarsKubeX: the pre-existing contents: read was never documented either. Out of scope.

packages/main/src/plugin/openshell-cli/openshell-sdk-client-manager.ts
```diff
@@ -0,0 +1,146 @@
...
+    return client;
+  }
+
+  invalidate(gatewayName?: string): void {
+    if (gatewayName) {
+      this.#cache.delete(gatewayName);
+    } else {
+      this.#cache.clear();
+    }
+  }
```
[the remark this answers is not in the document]
MarsKubeX: Over-engineering for slice 1. No caller uses this factory yet; the concurrent-getClient scenario is hypothetical.

README.md
```diff
@@ -65,12 +65,15 @@ git clone https://github.com/openkaiden/kaiden && cd kaiden
...
-Fetch all dependencies using the command `pnpm`:
+Kaiden depends on `@nvidia/openshell-sdk` which is published to GitHub Packages. A GitHub token with `read:packages` scope is required for `pnpm install` to succeed:
 
 ```sh
+export GITHUB_TOKEN=$(gh auth token)
 pnpm install
 ```
 
+If `gh auth token` does not include `read:packages`, run `gh auth refresh -s read:packages` first.
+
```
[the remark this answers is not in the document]
MarsKubeX: The existing README already uses POSIX-only syntax throughout. Adding PowerShell for just this step is inconsistent.

packages/main/src/plugin/openshell-cli/openshell-sdk-client.ts
```diff
@@ -0,0 +1,149 @@
...
+    return await readFile(filePath);
+  } catch (error: unknown) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
+      return undefined;
+    }
+    throw error;
+  }
+}
+
+async function buildConnectOptions(gateway: GatewayInfo): Promise<ConnectOptions> {
```
benoitf: hello, I'm wondering why we have functions on this class

It can't be private methods or another singleton class/object ?

MarsKubeX: Hi @benoitf, thanks for the feedback. You're right, I've created another singleton for those functions.

bmahabirbu: Just a question but wouldn't a singleton here not be useful since OpenshellGatewayConfig is stateless? 

I think u can just remove the .inSingletonScope() and it will still work fine since u already injected it into the sdk?

MarsKubeX: @bmahabirbu You're right. Removed the `.inSingletonScope()`. Thanks for the feedback.

packages/main/src/plugin/openshell-cli/openshell-sdk-client-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(
+    @inject(OpenshellCli)
+    private readonly openshellCli: OpenshellCli,
+    @inject(OpenshellGatewayConfig)
+    private readonly gatewayConfig: OpenshellGatewayConfig,
+  ) {}
+
+  async getClient(gatewayName?: string): Promise<OpenShellClient> {
+    const gateway = await this.#resolveGateway(gatewayName);
```
bmahabirbu, line 45: Super small nit:  there's an await between checking the cache and saving to it. So two calls for the same gateway at once both see an empty cache and both build a client

Since mostly everything is lazy this wouldnt matter but maybe for the future?

bmahabirbu (approved):
Overall LGTM see my one super small comment

packages/main/src/plugin/openshell-cli/openshell-sdk-client-manager.ts
MarsKubeX: Good catch, agreed this is a latent race. I'll address it when the first consumer lands in https://github.com/openkaiden/kaiden/issues/2713

packages/main/src/plugin/openshell-cli/openshell-sdk-client.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Cached factory for OpenShell SDK clients. Resolves gateway metadata from
+ * the CLI (`openshell gateway list`) and delegates connect-option assembly
+ * to {@link OpenshellGatewayConfig}.
+ *
+ * Clients are lazy — no network request is made until the first RPC.
+ */
+@injectable()
+export class OpenshellSdkClient {
```
jeffmaury: **praise:** name is confusing as it manages OpenShellClient so I would perhaps rename it to OpenShellSdkClientManager or OpenShellSdkClientCache

jeffmaury (approved):

packages/main/src/plugin/openshell-cli/openshell-sdk-client.ts
MarsKubeX: Thanks for the feedback. Renamed to `OpenshellSdkClientManager`.

benoitf (approved):

PR #2734, opened 2026-08-26
fix: use OpenShell SDK for sandbox create and delete

### LINKED ISSUE #2731: Switch openshell CLI create and delete sandboxes to openshell sdk client.sandbox.xxx()
### Task content

Continue the integration of the OpenShell Typescript SDK

bmahabirbu:
## Summary

- replace OpenShell CLI sandbox create and delete calls with the TypeScript SDK
- wait for sandbox readiness before running CLI-backed uploads and policy updates
- wait for deletion completion and show an immediate red Deleting state after confirmation
- remove obsolete CLI lifecycle wrappers, polling workarounds, option types, and tests
- keep upload, policy, inference, and listing operations on the CLI where the current SDK has no equivalent RPC

## Validation

try creating and deleting a workspace

Fixes #2731


MarsKubeX (approved):
LGTM. Tested creating and deleting a workspace and working good.

benoitf (commented):
I would update first to a recent version of OpenShell before doing this switch

jeffmaury (commented):
Why did we remove the polling of the gateways ?

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -473,22 +492,29 @@ export class AgentWorkspaceManager implements Disposable {
...
     task.state = 'running';
     task.status = 'in-progress';
     try {
-      await this.openshellCli.deleteSandbox(workspaceName, gateway);
-      this.closeWorkspaceTerminal(id);
-      await rm(this.getGlobalConfigDir(gateway, workspaceName), { recursive: true, force: true });
+      const sdkClient = await this.openshellSdkClientManager.getClient(gateway);
+      await sdkClient.sandbox.delete(name);
       this.apiSender.send('agent-workspace-update');
+      await sdkClient.sandbox.waitDeleted(name, SANDBOX_DELETE_TIMEOUT_SECONDS);
```
jeffmaury: **praise:** I don't get why you are adding this as it is managed by the polling mecanism that will update the state.

bmahabirbu: Are you referencing the waitDeleted?

That makes sure the sandbox is fully deleted before we do the final cleanup 

(which is basically removing the unique id and config)

 but your right there isnt a reason to wait for that since the cleanup is on the kaiden side

and the polling for state change is separate

jeffmaury (approved):

bmahabirbu:
@coderabbit resolve

packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
bmahabirbu: hmm on further inspection waiting on the polling makes the delete feel less responsive im still ok with removing as long as thats ok

PR #2743, opened 2026-08-27
chore: add agent to keep the website in sync

### LINKED ISSUE #2509: Evaluate custom fullsend agent to sync code and documentation
### Task content

Evaluate if we can use fullsend with a custom agent so that the documentation is updated with the latest changes in the codebase

benoitf:
1. it scans the website for extensions, and major providers and generate a file (pre script)

2. then we run the agent (if there are changes on this file) and the agent generate updates for a website

3. if there are changes, post-script create/update a pull request to update the website no changes: do nothing

fixes https://github.com/openkaiden/kaiden/issues/2509


.fullsend/scripts/pre-webupd.sh
```diff
@@ -0,0 +1,169 @@
...
+
+    # Capture sidebar structure
+    if [[ -f "${WEBSITE_DIR}/docs/sidebars.ts" ]]; then
+      website_sidebar=$(cat "${WEBSITE_DIR}/docs/sidebars.ts")
+    fi
+
+    echo "Captured $(echo "$website_docs" | jq 'length') website doc files"
+    rm -rf "${WEBSITE_DIR}"
+  else
+    echo "WARNING: Failed to clone website repo. Agent will run without website content."
```
bmahabirbu: Question: Lets say if the clone fails wouldnt the agent still launch and treat every page as not having docs? since above website_docs [] is emtpy?

bmahabirbu (commented):
LGTM but I had one question about what happens if the clone fails

.github/workflows/fullsend-update-website.yaml
```diff
@@ -0,0 +1,109 @@
...
+  cancel-in-progress: true
+
+jobs:
+  run:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout kdn repository
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+
+      - name: Checkout website repository
```
jeffmaury, line 41: **question:** It seems to me that the scripts are cloning the website repo so do we need this ?

benoitf: I'm wondering the best case, running locally it's easier if it's done in the script but probably on github action side it's better if we run it through an action

.fullsend/scripts/pre-webupd.sh
benoitf: I think it's related to the question of Jeff also. If I move cloning to the action or keep only in the script

we can abort earlier if the clone fails if it's the action

.github/workflows/fullsend-update-website.yaml
benoitf: will use GH action only

.fullsend/scripts/pre-webupd.sh
benoitf: will use GH action only

jeffmaury (approved):

PR #2749, opened 2026-08-27
chore: fix the creation of sandboxes when using OpenShell 0.0.113

benoitf:
note: this is a PR against a branch (OpenShell update)

we were mostly creating the sandboxes as a background task but were using a "true" command

now, use detach mode. Also ensure we give --tty mode on the creation else it won't be possible to connect later

note: I noticed that now, when we logout from connecting to the sandbox, the sandbox is dropped as a container using podman (but it's not related to this issue)


jeffmaury (approved):
LGTM

fbricon (approved):
works for me without having to bind the gateway to the IPv6 loopback. +1

PR #2756, opened 2026-08-31
feat(openshell): bundle mkfs.ext4 and debugfs on macOS

### LINKED ISSUE #2670: Package mkfs.ext4 on MacOS

In order to enable vm-driver, mkfs.ext4 is required to we need to bundle it on MacOS

Binaries are available from an OCI images which are located here:

https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fe2fsprogs/versions

References are seen here: https://github.com/Homebrew/homebrew-core/blob/main/Formula/e/e2fsprogs.rb

The manifest for the package should be downloaded first to list the proper bottle/layer 

_No response_

### Additional context

_No response_

MarsKubeX:
Download pinned e2fsprogs 1.47.4 and gettext 1.0 Homebrew bottles from GHCR, verify their digests, and stage mke2fs, debugfs, and the required dylibs (including libintl from gettext) into the darwin-arm64 asset directory. Shell wrappers set DYLD_LIBRARY_PATH and MKE2FS_CONFIG so the bundled tools run self-contained without a Homebrew installation.

### Unit tests

```bash
pnpm exec vitest --run extensions/openshell/src/e2fsprogs-download.spec.ts
```

7 tests covering:
- [x] Downloads and stages both e2fsprogs and gettext bottles for macOS
- [x] Bundles all required dylibs (libblkid, libcom_err, libe2p, libext2fs, libss, libuuid, libintl)
- [x] Creates shell wrappers with DYLD_LIBRARY_PATH and MKE2FS_CONFIG
- [x] Creates mke2fs → mkfs.ext4 symlink for correct argv[0] behavior
- [x] Reuses a complete cached installation (skips re-download)
- [x] Re-downloads when installation is incomplete (missing files or wrong version)
- [x] Reports registry failures with the failing endpoint
- [x] Rejects unsupported architecture
- [x] Cleans up partial installation on checksum mismatch
- [x] Selects correct OCI platform for x64 architecture

### Local integration test

Downloaded and checksum-verified the live e2fsprogs 1.47.4 and gettext 1.0 macOS arm64 bottles. With `PATH` restricted to Kaiden's packaged asset directory, the bundled `mkfs.ext4` created and populated an ext4 image and the bundled `debugfs` read the proof file back:

```bash
e2fs_bin="$PWD/extensions/openshell/assets/darwin-arm64"
test_dir=$(mktemp -d /tmp/kaiden-e2fsprogs-test.XXXXXX)

mkdir "$test_dir/source"
printf 'bundled mkfs.ext4 works\n' > "$test_dir/source/proof.txt"
truncate -s 32M "$test_dir/rootfs.ext4"

env -i PATH="$e2fs_bin" mkfs.ext4 -q -F -t ext4 \
  -d "$test_dir/source" "$test_dir/rootfs.ext4"
env -i PATH="$e2fs_bin" debugfs -R 'cat /proof.txt' "$test_dir/rootfs.ext4"

rm -rf "$test_dir"
```

Observed output:

```
debugfs 1.47.4 (6-Mar-2025)
bundled mkfs.ext4 works
```

### Dynamic library verification

Confirmed via `otool -L` that all non-system dependencies of the bundled `mke2fs` binary are present in `.mkfs-ext4/lib/`:

| Install name (leaf) | Bundled file | Source |
|---------------------|-------------|--------|
| `libext2fs.2.1.dylib` | ✅ present | e2fsprogs bottle |
| `libcom_err.1.1.dylib` | ✅ present | e2fsprogs bottle |
| `libe2p.2.1.dylib` | ✅ present | e2fsprogs bottle |
| `libblkid.2.0.dylib` | ✅ present | e2fsprogs bottle |
| `libuuid.1.1.dylib` | ✅ present | e2fsprogs bottle |
| `libintl.8.dylib` | ✅ present | gettext bottle |
| `libss.1.0.dylib` | ✅ present | e2fsprogs bottle |

Closes #2670

### DIFF
--- a/extensions/openshell/scripts/download.ts
+++ b/extensions/openshell/scripts/download.ts
@@ -137,8 +137,10 @@ if (targets.length === 0) {
   }
 
   if (selectedDownloads.some(entry => entry.id === 'openshell')) {
-    for (const { platform, arch } of targets.filter(target => target.platform === 'linux')) {
-      await downloadMkfsExt4(pkg.e2fsprogsVersion, arch, resolve(ASSETS_DIR, `${platform}-${arch}`));
+    for (const { platform, arch } of targets.filter(
+      target => target.platform === 'linux' || target.platform === 'darwin',
+    )) {
+      await downloadMkfsExt4(pkg.e2fsprogsVersion, arch, platform, resolve(ASSETS_DIR, `${platform}-${arch}`));
     }
   }
 })();
--- a/extensions/openshell/src/e2fsprogs-download.ts
+++ b/extensions/openshell/src/e2fsprogs-download.ts
@@ -14,10 +14,10 @@
  * limitations under the License.
  *
  * SPDX-License-Identifier: Apache-2.0
- ***********************************************************************/
+ **********************************************************************/
 
 import { createWriteStream, existsSync } from 'node:fs';
-import { chmod, copyFile, mkdir, mkdtemp, readdir, readFile, rm, writeFile } from 'node:fs/promises';
+import { chmod, copyFile, mkdir, mkdtemp, readdir, readFile, rm, symlink, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { pipeline } from 'node:stream/promises';
@@ -27,7 +27,9 @@ import * as tar from 'tar';
 import { sha256 } from './sha256';
 
 const REGISTRY = 'https://ghcr.io';
-const REPOSITORY = 'homebrew/core/e2fsprogs';
+const E2FSPROGS_REPOSITORY = 'homebrew/core/e2fsprogs';
+const GETTEXT_REPOSITORY = 'homebrew/core/gettext';
+const GETTEXT_VERSION = '1.0';
 const MANIFEST_ACCEPT = [
   'application/vnd.oci.image.index.v1+json',
   'application/vnd.oci.image.manifest.v1+json',
@@ -37,8 +39,8 @@ const MANIFEST_ACCEPT = [
 const METADATA_TIMEOUT_MS = 30_000;
 const BOTTLE_DOWNLOAD_TIMEOUT_MS = 5 * 60_000;
 
-const LIBRARIES = ['libcom_err.so.2', 'libe2p.so.2', 'libext2fs.so.2', 'libss.so.2'];
-const RUNTIME_DIRECTORY = 'e2fsprogs';
+const LINUX_LIBRARIES = ['libcom_err.so.2', 'libe2p.so.2', 'libext2fs.so.2', 'libss.so.2'];
+const DARWIN_LIBRARY_STEMS = ['libblkid', 'libcom_err', 'libe2p', 'libext2fs', 'libss', 'libuuid'];
 const BINARIES = [
   { name: 'mkfs.ext4', payload: 'mke2fs' },
   { name: 'debugfs', payload: 'debugfs' },
@@ -67,13 +69,21 @@ interface RegistryToken {
   access_token?: string;
 }
 
+function runtimeDirectoryName(platform: string): string {
+  return platform === 'linux' ? 'e2fsprogs' : '.mkfs-ext4';
+}
+
 function architectureForOci(arch: string): string {
   if (arch === 'x64') return 'amd64';
   if (arch === 'arm64') return 'arm64';
   throw new Error(`unsupported e2fsprogs architecture: ${arch}`);
 }
 
-function isRequiredBottleEntry(path: string, type: string, version: string): boolean {
+function isDarwinLibrary(path: string, bottleRoot: string): boolean {
+  return DARWIN_LIBRARY_STEMS.some(stem => path.startsWith(`${bottleRoot}/lib/${stem}`) && path.endsWith('.dylib'));
+}
+
+function isRequiredBottleEntry(path: string, type: string, version: string, platform: string): boolean {
   if (type === 'Link' || type === 'SymbolicLink') {
     return false;
   }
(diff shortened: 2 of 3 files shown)


extensions/openshell/scripts/download.ts
```diff
@@ -29,6 +29,7 @@ import {
   OPENSHELL_IMAGE_BUILDER_DOWNLOAD,
   type GitHubArtifactDownload,
 } from '../src/openshell-download';
+import { downloadMkfsExt4 } from '../src/e2fsprogs-download';
```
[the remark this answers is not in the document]
MarsKubeX: The existing import on the line above (from '../src/openshell-download') uses the same relative pattern. Additionally, scripts/download.ts runs via tsx at build time, not through Vite

extensions/openshell/src/e2fsprogs-download.ts
```diff
@@ -0,0 +1,332 @@
...
+}
+
+async function downloadAndVerifyBottle(
+  repository: string,
+  version: string,
+  arch: string,
+  outputDir: string,
+): Promise<string> {
+  const { digest, token } = await resolveBottle(repository, version, arch);
+  const expectedDigest = digest.replace(/^sha256:/, '');
```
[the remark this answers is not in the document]
MarsKubeX: Homebrew rebuilds bottles independently of the upstream e2fsprogs version — the same 1.47.4 tag gets new digests whenever Homebrew recompiles for a new macOS release or dependency update. Pinning per-bottle-rebuild digests would require maintaining a matrix of SHA-256 values per architecture per macOS version, and would break silently whenever Homebrew publishes a rebuild. The current approach pins the e2fsprogs version in package.json and verifies the downloaded blob matches the digest the registry advertises for that version, which is the same trust model Homebrew itself uses.

bmahabirbu (commented):
A lot of the same work is being redone here https://github.com/openkaiden/kaiden/pull/2692 

im wondering if that should go in first then for this pr to build on top adding the necessarily libraries for mac?

jeffmaury (approved):
Approving but should be synced with #2692

extensions/openshell/src/e2fsprogs-download.ts
```diff
@@ -195,78 +280,131 @@ async function stageBottle(archive: string, version: string, arch: string, outpu
...
-    join(outputDir, RUNTIME_DIRECTORY, 'mke2fs.conf'),
-    join(outputDir, RUNTIME_DIRECTORY, 'NOTICE'),
-  ].every(path => existsSync(path));
+    ...BINARIES.map(binary => join(runtimeDir, binary.payload)),
+    join(runtimeDir, 'mke2fs.conf'),
+  ];
+  if (platform === 'linux') {
+    paths.push(...LINUX_LIBRARIES.map(library => join(runtimeDir, 'lib', library)), join(runtimeDir, 'NOTICE'));
+  } else {
+    paths.push(join(runtimeDir, 'mkfs.ext4'), join(runtimeDir, 'lib'));
```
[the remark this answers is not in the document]
MarsKubeX: The version marker file (.mkfs-ext4-version) is only written after all staging completes — including the gettext download and dylib copy for Darwin. A cached installation with a valid marker already implies every dylib was staged successfully.

On Linux the check validates each library by exact SONAME (e.g. libcom_err.so.2) because those names are stable. On Darwin the versioned dylib filenames vary across Homebrew bottle rebuilds (e.g. libcom_err.1.1.dylib vs libcom_err.1.2.dylib), so validating exact names would require either persisting a staged-file manifest or hardcoding version suffixes — both add complexity disproportionate to the risk. The lib/ directory existence check catches the practical failure mode (partial deletion or interrupted staging). If stricter per-file validation is needed it can be tracked as a separate issue.

extensions/openshell/src/e2fsprogs-download.ts
```diff
@@ -67,26 +69,39 @@ interface RegistryToken {
...
   if (type === 'Link' || type === 'SymbolicLink') {
     return false;
   }
   const bottleRoot = `e2fsprogs/${version}`;
   const requiredPaths = [
     ...BINARIES.map(binary => `${bottleRoot}/sbin/${binary.payload}`),
     `${bottleRoot}/.bottle/etc/mke2fs.conf`,
-    `${bottleRoot}/NOTICE`,
   ];
+  if (platform === 'linux') {
```
bmahabirbu: We should not exclude the notice for macos here! I'm pretty sure its needed for gpl compliance when distributing this package

extensions/openshell/src/e2fsprogs-download.ts
```diff
@@ -195,78 +285,143 @@ async function stageBottle(archive: string, version: string, arch: string, outpu
...
+    paths.push(...LINUX_LIBRARIES.map(library => join(runtimeDir, 'lib', library)), join(runtimeDir, 'NOTICE'));
+  } else {
+    paths.push(join(runtimeDir, 'mkfs.ext4'));
+    const libDir = join(runtimeDir, 'lib');
+    if (!existsSync(libDir)) return false;
+    try {
+      const libFiles = readdirSync(libDir);
+      const hasDarwinLibs = DARWIN_LIBRARY_STEMS.every(stem =>
+        libFiles.some(f => f.startsWith(stem) && f.endsWith('.dylib')),
```
bmahabirbu, line 372: Here is the validation 

`libFiles.some(file => file.startsWith(stem) && file.endsWith('.dylib'))`

We should handle versioning too for example ```
libext2fs.1.1.dylib
  libext2fs.dylib
```

bmahabirbu (changes requested):
Looks good thanks! just two small fixes and we will be all set

1. Blocker: macOS drops third-party notices
  Linux extracts and preserves the e2fsprogs NOTICE, but Darwin explicitly excludes it

2. Non blocker: Validation only requires any filename matching each stem but we have files like this

  libext2fs.1.1.dylib
  libext2fs.dylib

so if we have libext2fs.1.1.dylib and that exact file disappears, the remaining unversioned one might not work

extensions/openshell/src/e2fsprogs-download.ts
MarsKubeX: Agreed it's non-blocker. The tar extraction already filters out Link/SymbolicLink entries, so in practice only the versioned dylibs (the real files) get staged and the unversioned ones in Homebrew bottles are symlinks and are never materialized. The "at least one per stem" check covers the files that actually exist after staging. If we want stricter validation in the future we can track it separately.

extensions/openshell/src/e2fsprogs-download.ts
MarsKubeX: Thanks for the feedback. I added again the ```NOTICE```.

bmahabirbu (approved):
Thanks for the changes! LGTM

PR #2765, opened 2026-09-01
fix(openshell): support renamed sandbox release assets

bmahabirbu:
### What does this PR do?

Supports both the legacy `unknown-linux-gnu` and current `unknown-linux-musl` OpenShell sandbox archive names. The downloader selects the name present in the release digest map, preserving the current 0.0.113 build while allowing upgrades to 0.0.114 and newer.

This unblocks #2752.

### How was this tested?

- `pnpm exec vitest --run --project=openshell extensions/openshell/src/openshell-download.spec.ts`
- `pnpm run typecheck:extensions:openshell`


jeffmaury (commented):
As I am a little bit cautious about musl, why do we need the sandbox binary ?

MarsKubeX (approved):
LGTM codewise

benoitf (approved):
@jeffmaury this is bc OpenShell do not ship anymore the gnu binary

it's using musl to be more universal and run on a wider range of Linux

jeffmaury (approved):

PR #2769, opened 2026-09-01
Allow running packaged and dev Kaiden instances simultaneously

fbricon:
Separate dev instance identity so it doesn't conflict with a packaged
install: redirect userData to kaiden-dev, set app name to "Kaiden Dev",
use a dev-branded dock icon, and stamp a "DEV" pixel badge on the
macOS tray icon.

Screenshot 2026-09-01 at 16 51 31


benoitf (changes requested):
I'm not sure dev instance should use a different folder for settings
note: we already have an ENV to use custom settings folder (for tests)

then, some files should be handled first upstream

bmahabirbu (commented):
Following the PD upstream pr but it seems pd folks are handling the review

PR #2776, opened 2026-09-03
fix(codex): route OpenAI through OpenShell inference

### LINKED ISSUE #2775: Codex Agent doesnt work with codex api
### Bug description

Create an openai connection using:

Base URL: https://api.openai.com/v1
apikey

and enable codex it wont work

### Operating system

all

### Installation Method

None

### Version

0.3.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

bmahabirbu:
## Summary

- configure Codex to use an HTTP-only OpenShell model provider
- select OpenAI connections as the active OpenShell inference route
- provide a non-secret API key placeholder while keeping credentials in OpenShell
- trust the Kaiden-managed `/sandbox` workspace to avoid the startup prompt

## Testing

- `pnpm run test:extensions:codex`

Fixes #2775

### DIFF
--- a/extensions/codex/src/extension.spec.ts
+++ b/extensions/codex/src/extension.spec.ts
@@ -126,11 +126,30 @@ describe('activate', () => {
       const agent = getRegisteredAgent();
 
       const configFile = createConfigFile();
-      await agent.preWorkspaceStart(createContext([configFile]));
+      const context = createContext([configFile]);
+      await agent.preWorkspaceStart(context);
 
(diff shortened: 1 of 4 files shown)


jeffmaury (commented):
I would delay this PR as latest OpenShell does not use inference.local anymore so we need to check what would be the solution

jeffmaury (commented):
I would delay after the latest version of OpenShell as it does not use inference.local anymore

jeffmaury:
See #2928

PR #2785, opened 2026-09-03
fix(openshell): configure sandbox policy through SDK

### LINKED ISSUE #2746: Set sandbox network policy and providers-v2 flags through the SDK
### Task content

Replace:

- `updatePolicy(sandbox, endpoints, binaries)` → `client.sandbox.setPolicy()` (sandbox-scoped; network policies only)
- `isV2ProviderEnabled()` / `enableV2Provider()` → `client.raw.getGatewayConfig` / `client.raw.updateConfig({ global: true, ... })`

Keep the same `OpenshellCli` signatures so `agent-workspace-manager` does not need a redesign.

## Problem

Workspace create enables providers v2 globally, then patches network policy with CLI `policy update --add-endpoint`. Curated `getConfig` / `setSetting` are **sandbox-scoped** and are the wrong APIs for the v2 flag.

## Scope

- Translate `--add-endpoint` / `--binary` strings into `SandboxPolicy.networkPolicies` (reuse `openshell-network-policy` helpers if they already build a policy object)
- `setPolicy` cannot change filesystem/landlock/process after create
- Read/write `providers_v2_enabled` via `raw` + global scope
- On `updatePolicy` failure, callers already delete the sandbox; keep that contract

bmahabirbu:
## Summary

Switch policy updates from CLI merges to SDK `setPolicy()`. SDK reads include gateway-generated `_provider_*` rules, so exclude those before submitting the base policy; OpenShell adds them back automatically. This applies to any provider that generates rules, including Vertex AI.

Wait for the submitted revision to report `LOADED` before continuing workspace setup.

## Validation

- Live CLI/SDK comparisons confirmed the behavior with Vertex AI and a dummy NVIDIA provider.
- 2,936 main-process tests passed; type checking, Svelte checking, formatting, and lint passed.

Closes #2746


jeffmaury (changes requested):
- I have an error while creating a sandbox (using Vertex AI)
- Don't get why gateway management has been updated and part of this PR

packages/main/src/plugin/openshell-cli/openshell-policy-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+      process: currentPolicy?.process,
+      networkPolicies: {
+        ...baseNetworkPolicies,
+        ...policy.networkPolicies,
+      },
+      networkMiddlewares: currentPolicy?.networkMiddlewares,
+    });
+    // The SDK hash-based wait compares a base hash with the effective policy
+    // hash when providers are attached. Wait for the revision to load instead.
+    const { PolicyStatus } = await import('@nvidia/openshell-sdk/raw');
```
jeffmaury: **praise:** should be moved at the beginning of the file

packages/main/src/plugin/openshell-cli/openshell-policy-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+    const deadline = Date.now() + 60_000;
+    while (Date.now() < deadline) {
```
jeffmaury, line 53: **praise:** why don't you simply run a single call with the required timeout ?

bmahabirbu: Yeah agreed thats a easy fix ill adjust

packages/main/src/plugin/openshell-cli/openshell-policy-manager.ts
bmahabirbu: Will do

packages/main/src/plugin/openshell-cli/openshell-policy-manager.ts
bmahabirbu: seems the api has a bug

```
  await client.sandbox.setPolicy(name, policy, {
    wait: true,
    waitTimeoutSecs: 60,
  });
```

Basically we want this but according to claude and checking some experiments The SDK wait returned in 5–6 ms while the revision was still PENDING; actual loading took 3–10 seconds. So it would return before the policy is fully loaded

jeffmaury (approved):

PR #2786, opened 2026-09-03
feat(workspaces): use mounts for supported gateways

### LINKED ISSUE #2768: Update create workspace with mount
### Task content

Once #2767 is implemented, we should update the create workspace so that uploads are replaced by mounts if the gateway supports it

bmahabirbu:
Switch to bind mounts on supported gateways: project folders are read-write and skills are read-only. Agent config still uses uploads because agents need to modify and atomically replace those files at runtime. Gateways without mount support keep using uploads.

Fixes #2768


bmahabirbu:
Addressed both CodeRabbit findings in `fc3545826`:

- normalized relative/home mount targets and require them to remain strict descendants of `/sandbox`, with upload fallback for traversal or workspace-root targets
- require the selected gateway registration to be local and locally addressed before enabling bind mounts, preventing stale managed storage from enabling mounts for a remote gateway
- added regression coverage for `$SOURCES/foo/..`, `$SOURCES/../etc`, and remote registrations with stale local configuration

Validated with 201 focused tests, `typecheck:main`, ESLint, formatting, and `git diff --check`.

MarsKubeX (approved):
LGTM. The comment from codeRabbit regarding "Reject the normalized sandbox root" seems valid to me.

benoitf (approved):

jeffmaury (changes requested):
I tried and entered the agent but using df I did not see the workspace folder being mounted. Seems this PR is only considering mounts from workspace.json (which is deprecated or only used by agents) and should rather consider uploads as mounts

jeffmaury (changes requested):
Broken for as Claude on Vertex is launched but the UI does not show up.

Also why create a file for the driver config as the CLI expect as string

packages/api/src/openshell-gateway-info.ts
```diff
@@ -117,7 +129,8 @@ export interface CreateSandboxOptions {
   providers?: string[];
   env?: Record<string, string>;
   labels?: Record<string, string>;
-  uploads?: Array<{ local: string; remote: string }>;
+  uploads?: OpenshellUpload[];
+  driverConfig?: Partial<Record<'podman' | 'docker', { mounts: OpenshellBindMount[] }>>;
```
jeffmaury, line 133: **question:** why not using LocalGatewayDriver

jeffmaury (approved):
LGTM whith the simple remark on code

packages/api/src/openshell-gateway-info.ts
bmahabirbu: clean thanks

bmahabirbu:
@coderabbit resolve

PR #2800, opened 2026-09-07
fix: escape paths in gateway configuration file

### LINKED ISSUE #2797: Gateway configuration file causes issues on Windows
### Bug description

As paths are stored in this file and on Windows in paths \ should be escaped

### Operating system

Windows

### Installation Method

Other

### Version

next (development version)

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

jeffmaury:
Fixes #2797

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
@@ -23,7 +23,7 @@ supervisor_image = "{{{supervisorImage}}}"
 allow_unauthenticated_users = true
 
 [openshell.gateway.gateway_jwt]
-signing_key_path = '{{{gatewayDir}}}/jwt/signing.pem'
-public_key_path = '{{{gatewayDir}}}/jwt/public.pem'
-kid_path = '{{{gatewayDir}}}/jwt/kid'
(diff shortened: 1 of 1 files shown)


packages/main/src/plugin/openshell-cli/openshell-gateway.toml.template
```diff
@@ -23,7 +23,7 @@ supervisor_image = "{{{supervisorImage}}}"
 allow_unauthenticated_users = true
 
 [openshell.gateway.gateway_jwt]
-signing_key_path = "{{{gatewayDir}}}/jwt/signing.pem"
-public_key_path = "{{{gatewayDir}}}/jwt/public.pem"
-kid_path = "{{{gatewayDir}}}/jwt/kid"
+signing_key_path = '{{{gatewayDir}}}/jwt/signing.pem'
```
benoitf: it won't work if path is for example `C:\Users\O'Brien\`

you need to escape path before it's being added there

also if you don't want to mix slash and backslash, probably need to provide the publicKey Path as a template value

or you can also replace backslash by slash when providing gatewayDir

benoitf (changes requested):

benoitf (approved):

fbricon (approved):

PR #2822, opened 2026-09-08
fix(gateway): use detected gateway instead of kaiden-local on same port

### LINKED ISSUE #2751: Default gateway is used when sandbox is created
### Bug description

When a sandbox is create, the default selected gateway is used instead of the one selected by the user in the create sandbox wizard

### Operating system

All

### Installation Method

None

### Version

0.3.3

### Steps to reproduce

_No response_

### Relevant log output

```shell

```

### Additional context

_No response_

fbricon:
Previously, a gateway at 0.0.0.0 was not recognized as local, so when
openshell was running there, init() would fall through, find the default
port healthy (because 0.0.0.0 accepts connections on 127.0.0.1), and
register kaiden-local — creating a duplicate pointing to the same
process. All subsequent CLI calls then targeted kaiden-local instead of
the user's gateway.

Changes:
- Add 0.0.0.0 to isLocalEndpoint() so gateways bound to all interfaces
  are recognized as local. This lets init() find openshell in the
  for loop and return early, so kaiden-local is never created.
- When kaiden-local already exists from a previous session and another
  gateway is healthy on the same port, try the other gateway first and
  remove the redundant kaiden-local registration.
- Fix the workspace creation dropdown to only show reachable gateways
  (reachable === true instead of !== false) and use untrack() to
  preserve the user's manual gateway selection across store updates.

kaiden-local is now only created when no other local gateway is running
on that port — i.e., Kaiden genuinely needs to start its own gateway.

fixes #2751

### DIFF
--- a/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
+++ b/packages/main/src/plugin/openshell-cli/openshell-gateway.spec.ts
@@ -185,6 +185,66 @@ describe('init', () => {
     expect(spawn).not.toHaveBeenCalled();
   });
 
+  test('treats gateway with 0.0.0.0 endpoint as local', async () => {
+    vi.spyOn(console, 'log').mockImplementation(() => undefined);
+    vi.mocked(openshellCli.listGateways).mockResolvedValue([
+      { name: 'openshell', endpoint: 'http://0.0.0.0:17670', active: false } as GatewayInfo,
+    ]);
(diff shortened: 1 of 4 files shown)


benoitf (commented):
coderabbit comment is valid ?

jeffmaury (approved):
LGTM but not sure it cover the intent of the issue.

PR #2868, opened 2026-09-11
fix: use latest dev OpenShell version

jeffmaury:
Pushing as draft as Claude on Vertex is still broken

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


packages/main/src/plugin/agent-workspace/agent-workspace-manager.ts
```diff
@@ -248,11 +233,6 @@ export class AgentWorkspaceManager implements Disposable {
 
     const t0 = performance.now();
 
-    const v2Globally = await this.openshellCli.isV2ProviderEnabled();
-    if (!v2Globally) {
-      await this.openshellCli.enableV2Provider();
-    }
-
     const tV2 = performance.now();
     console.log(`[workspace-timing] enableV2Provider: ${(tV2 - t0).toFixed(0)}ms`);
```
fbricon: needs to be deleted

PR #2885, opened 2026-09-14
refactor(secret-manager): migrate provider ops from CLI to SDK raw RPCs

### LINKED ISSUE #2747: Manage OpenShell providers through SDK raw RPCs
### Task content

Point Secret Vault / `openshell-secret-adapter` at gateway-level RPCs:

- `listProviders` → `client.raw.listProviders`
- `createProvider` → `client.raw.createProvider`
- `deleteProvider` → `client.raw.deleteProvider`
- `listProfiles` → `client.raw.listProviderProfiles`

Do **not** use `client.sandbox.listProviders(sandboxName)` — that lists providers attached to one sandbox.

## Problem

Secrets in Kaiden are OpenShell providers. The curated SDK only wraps sandbox attach/detach. Gateway CRUD is already on `client.raw`.

## Scope

- Keep `CreateProviderOptions` / `OpenshellProviderInfo` / `OpenshellProfile` as public Kaiden types
- Map credentials the same way as today; do not log them
- Gateway name still selects which client to use
- `setInference()` stays CLI (or a later issue)

jeffmaury:
## Summary

- Migrate `OpenshellSecretAdapter` from shelling out to `openshell` CLI to using `OpenshellSdkClientManager` with `client.raw` gRPC RPCs for all provider CRUD operations (listProviders, createProvider, deleteProvider, listProviderProfiles)
- Implement `--from-gcloud-adc` 3-step flow (create provider → configure OAuth2 refresh → rotate credential) with rollback on failure, replicating the Rust CLI behavior
- Remove now-unused provider methods (`listProviders`, `listProfiles`, `deleteProvider`, `createProvider`) from `OpenshellCli`

## Test plan

- [x] All 142 existing unit tests pass (92 openshell-cli, 24 secret-adapter, 26 secret-manager)
- [x] New tests cover gcloud ADC 3-step flow, rollback on failure, unsupported flags rejection, `readGcloudAdc()` validation
- [ ] Manual verification with a running gateway to confirm SDK RPCs work end-to-end

Fixes #2747


benoitf (changes requested):
question: the design pattern is invalid for flags

we pass an array of flags and then you're checking if flags are supported or not from a given constant. It looks very fragile

AFAIK, it should be like a boolean rather than a flag entry

but thinking it more, I suppose the createProviderWithGcloudAdc should be outside of this class of openshell-secret-adapter and handle only GcloudAdc

then rather doing 

if isAdc
 ...
else
 ...


it should do: getProviderFactory(....).createProvider()

and you've one factory for adc, another one for default, etc

bmahabirbu (commented):
i saw the gcloud configuration that was in the cli is now reimplemented in typescript in kaiden.

 Do u think we need this or should we keep delegating the functionality to the cli cause it might not be good for maintainability lognterm if we have this version then the cli one

jeffmaury:
> i saw the gcloud configuration that was in the cli is now reimplemented in typescript in kaiden.
> 
> Do u think we need this or should we keep delegating the functionality to the cli cause it might not be good for maintainability lognterm if we have this version then the cli one

Plan is to remove the use of the CLI thus this PR

bmahabirbu:
alright looking closer the cli part for --from-gcloud isnt that substantial so reproducing it in kaiden is ok

packages/main/src/plugin/secret-manager/gcloud-adc-provider-factory.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ProviderCredentialRefreshStrategy } from '@nvidia/openshell-sdk/raw';
+
+import type { SecretCreateOptions } from '/@api/secret-info.js';
+
+import type { ProviderFactory } from './provider-factory.js';
+
+export class GcloudAdcProviderFactory implements ProviderFactory {
+  async createProvider(client: OpenShellClient, options: SecretCreateOptions): Promise<void> {
+    const env = typeof options.value !== 'string' ? options.value.env : undefined;
+    const { clientId, clientSecret, refreshToken } = await readGcloudAdc(env);
```
bmahabirbu: Small-nit: Here I noticed we are reading the adc before we actually check if there is support for it? maybe we can move this to the after block?

bmahabirbu (approved):
Overall LGTM then! one nit

packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
```diff
@@ -18,56 +18,73 @@
...
+  #resolveFactory(options: SecretCreateOptions): ProviderFactory {
+    const flags = typeof options.value !== 'string' ? options.value.flags : undefined;
+    if (flags?.some(f => f !== FROM_GCLOUD_ADC)) {
+      const unsupported = flags.filter(f => f !== FROM_GCLOUD_ADC);
+      throw new Error(`Unsupported provider factory flag: ${unsupported.join(', ')}`);
+    }
+    if (flags?.includes(FROM_GCLOUD_ADC)) {
+      return new GcloudAdcProviderFactory();
+    }
+    return new DefaultProviderFactory();
```
benoitf: I think here it's not following dependency injection pattern

we should have factories being injected through dependency injection

then you don't instantiate them, it's singletons

about the choice of the factory, it can either be a Named injection metadata or you have an accept method on the factory where you provide the options

benoitf (commented):
as a side question why do we have flags in SecretOptions vs using a boolean like `isAdc` ?

jeffmaury:
@coderabbitai review

benoitf (commented):
question: both factories support the same type, isn't a design problem ?

jeffmaury:
> question: both factories support the same type, isn't a design problem ?

They are sorted using the priority field

benoitf:
but why is it a priority ? 'from-gcloud-adc' does not seem to be a priority, it seems to be an explicit option

jeffmaury:
> but why is it a priority ? 'from-gcloud-adc' does not seem to be a priority, it seems to be an explicit option

The factory is selected based on the connection type the flags have been removed as they were related to CLI

benoitf (approved):

packages/main/src/plugin/secret-manager/openshell-secret-adapter.ts
```diff
@@ -16,58 +16,74 @@
...
+      })),
+    }));
+  }
+
+  #resolveFactory(options: SecretCreateOptions): ProviderFactory {
+    const factory = this.providerFactories.find(f => f.supports(options.type));
+    if (!factory) {
+      return this.defaultProviderFactory;
+    }
+    return factory;
```
benoitf: return this.providerFactories.find(f => f.supports(options.type)) ?? this.defaultProviderFactory;

packages/main/src/plugin/secret-manager/gcloud-adc-provider-factory.ts
```diff
@@ -0,0 +1,139 @@
...
+    await client.raw.createProvider({
+      provider: {
+        metadata: { name: options.name },
+        type: options.type,
+        config: typeof value !== 'string' ? (value.config ?? {}) : {},
+      },
+      workspace: '',
+    });
+
+    const { clientId, clientSecret, refreshToken } = await readGcloudAdc(credentials);
```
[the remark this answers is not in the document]
jeffmaury: the provider is deleted in case of errors

jeffmaury:
@coderabbitai review

PR #2905, opened 2026-09-14
fix: sandboxes stuck in Error after app restart

serbangeorge-m:
## Summary

- Sandboxes go to Error state every time the app restarts because the gateway process is killed alongside Electron — they share the same process group (`detached: false`)
- The new gateway instance cannot re-adopt the orphaned sandbox containers and marks them Error
- This happens with or without active sessions, on every restart (including `pnpm watch` rebuilds)

## Fix

- Spawn gateway with `detached: true` + `unref()` so it survives app restarts
- Switch `start()` from piped stdio to file-descriptor stdio (matching `spawnCreatedGateway()`) to avoid SIGPIPE when Electron exits
- Don't kill the gateway on app quit — `init()` already discovers and reuses a healthy running gateway on next launch

## Test plan

- [x] Create a workspace (sandbox shows Ready)
- [x] Close the app
- [x] Reopen the app
- [x] Verify sandbox is still Ready (not Error)
- [x] Repeat with `pnpm watch` rebuild (edit a file, let watch restart)
- [x] Verify explicit gateway stop still works


packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -53,6 +52,16 @@ const STOP_TIMEOUT_MS = 5000;
 const SUPERVISOR_IMAGE_BASE = 'ghcr.io/nvidia/openshell/supervisor';
 const GATEWAY_LOG_FILENAME = 'gateway.log';
 const DEFAULT_GATEWAY_NAME = KAIDEN_LOCAL_GATEWAY_NAME;
+// Node spawn only configures listed stdio slots; unspecified parent fds are inherited.
```
bmahabirbu: I was trying to see what the smoke test failed and I saw it hanges on electron closing because of the openshell gateway hasnt stopped after the timeout. 

You can potentially use a different method for checking this

bmahabirbu (commented):
LGMT but seems test are failing check the comment i had for a suggestions that i gathered with the help of claude

serbangeorge-m:
@coderabbitai review

bmahabirbu (approved):
Looks good thanks

serbangeorge-m:
@coderabbitai resolve

jeffmaury (approved):

benoitf (changes requested):
please do not export functions

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -57,6 +56,16 @@ const SUPERVISOR_IMAGE_BASE = 'ghcr.io/nvidia/openshell/supervisor';
 const GATEWAY_LOG_FILENAME = 'gateway.log';
 const DEFAULT_GATEWAY_NAME = KAIDEN_LOCAL_GATEWAY_NAME;
 
+function spawnGatewayDetached(binaryPath: string, args: string[], logFd: number): ChildProcess {
```
benoitf: it should be in its proper utility class, injected by inversify

tests/playwright/src/fixtures/electron-app.ts
```diff
@@ -479,4 +482,18 @@ export async function closeAllWindows(electronApp: ElectronApplication): Promise
   await Promise.allSettled(windows.map(window => window.close()));
 }
 
+// On Linux the detached openshell-gateway inherits Playwright's CDP pipe FDs
+// via fork(). Kill it before app.close() so those pipes close and Playwright
+// can detect the Electron exit without hitting the 180s teardown timeout.
+export function killDetachedGateway(electronApp: ElectronApplication): void {
```
benoitf: should be a Promise

tests/playwright/src/fixtures/electron-app.ts
```diff
@@ -479,4 +482,18 @@ export async function closeAllWindows(electronApp: ElectronApplication): Promise
...
+  if (process.platform !== 'linux') return;
+  const pid = electronApp.process().pid;
+  if (!pid) return;
+  try {
+    execFileSync('/usr/bin/pkill', ['-TERM', '-f', '-P', String(pid), 'openshell-gateway'], { stdio: 'ignore' });
```
benoitf: issue: should not use a synchronous method

packages/main/src/plugin/openshell-cli/openshell-gateway.ts
```diff
@@ -421,7 +423,12 @@ export class OpenshellGateway implements Disposable {
       });
       this.#port = previousPort;
       this.#bindAddress = previousBindAddress;
-      const stderrOutput = stderrChunks.join('\n').trim();
+      let stderrOutput = '';
+      try {
+        stderrOutput = (await readFile(logPath, 'utf-8')).trim();
+      } catch {
+        // log file may not exist
```
benoitf: issue: do not silently ignore errors

benoitf:
side question: we're just pushing later the problem of restarting sandboxes.
If I restart my computer, is it expected that all my sandboxes are in error state ?

major issue: how is tracked this detached process ? if we create processes, we need a way to be be able to control the lifecycle if it's no longer the application lifecycle. Here depending on how you stop/do development, you can end up in multiple processes. What you're trying to achieve in the e2e tests is what users will face 
(it's like for a podman machine, you might want to control and let the user decide if stopping the app is stopping everything or not).

serbangeorge-m:
> side question: we're just pushing later the problem of restarting sandboxes. If I restart my computer, is it expected that all my sandboxes are in error state ?
> 
> major issue: how is tracked this detached process ? if we create processes, we need a way to be be able to control the lifecycle if it's no longer the application lifecycle. Here depending on how you stop/do development, you can end up in multiple processes. What you're trying to achieve in the e2e tests is what users will face (it's like for a podman machine, you might want to control and let the user decide if stopping the app is stopping everything or not).

Yes, I think after a full restart the gateway process is gone but this PR only fixed the app restart case, not a machine restart case. I can create a ticket related to full restart

PR #2935, opened 2026-09-16
fix(#2917): add hypervisor entitlement to mac vm driver

### LINKED ISSUE #2917: VM driver on macOS requires Hypervisor entitlement signing
## Context

This blocks switching the gateway from the podman driver to the vm driver (issue 2761, part of the vm-driver support epic 2669) on macOS.

## Problem

On macOS, the vm driver needs to be signed with the Hypervisor entitlement (com.apple.security.hypervisor), otherwise macOS will not allow the driver to create or manage VMs. Without this entitlement, the vm driver cannot function on macOS.

## Expected behavior

The vm driver binary (or the Kaiden application bundle) should be signed with the com.apple.security.hypervisor entitlement so that the vm driver works on macOS.

## Source

Identified by the assignee in issue 2761.

bmahabirbu:
this pr adds the hypervisor entitlement sign to the mac vm
driver because its needed to run it! Openshell cli does this
during installation and kaiden forgot to replicate it

Signed-off-by: Brian <bmahabir@bu.edu>

Fixes https://github.com/openkaiden/kaiden/issues/2917

### Manual verification (macOS)

1. On an Apple Silicon Mac with the project dependencies and OpenShell assets installed, build a local app without a signing certificate:

   ```sh
   CSC_IDENTITY_AUTO_DISCOVERY=false pnpm compile
   ```

   Packaging should complete without a `No identity found for signing` error. This tests the packaged app; `pnpm watch` does not run the packaging signing hook.

2. Inspect the packaged VM driver's entitlements:

   ```sh
   codesign --display --entitlements - "dist/mac-arm64/Kaiden.app/Contents/Resources/openshell/openshell-driver-vm"
   ```

   Confirm `com.apple.security.hypervisor` is present and set to `true`.

3. Verify the driver signature:

   ```sh
   codesign --verify --strict --verbose=2 "dist/mac-arm64/Kaiden.app/Contents/Resources/openshell/openshell-driver-vm"
   ```

   Verification should succeed. The local fallback signs only the VM driver; builds with an Apple signing certificate retain the normal app signing flow.


jeffmaury (commented):
Why is this needed ? What about the OpenShell CLI and gateway binaries ?

feloy:
> Why is this needed ? What about the OpenShell CLI and gateway binaries ?

This entitlement is necessary for using libkrun.

Only VM driver binary needs this entitlement, as it is the only one using libkrun.

What I am not completely sure is if it is necessary to do it for a binary released by OpenShell, I'll check

MarsKubeX (approved):
Tested and LGTM

feloy (approved):
LGTM
I confirm that the driver only works after signing the binary with the entitlment (tested with binary of v0.0.116)

PR #2937, opened 2026-09-16
fix(openshell): make bundled mkfs discoverable by the VM driver

### LINKED ISSUE #2916: VM driver cannot discover mkfs packages on PATH
## Context

This blocks switching the gateway from the podman driver to the vm driver (issue 2761, part of the vm-driver support epic 2669).

## Problem

When using the vm driver, the mkfs packages cannot be found. The driver needs the mkfs binaries to be discoverable on PATH or through another mechanism.

## Expected behavior

The vm driver should be able to locate and use the mkfs packages, either by discovering them on PATH or through an alternative resolution mechanism.

## Source

Identified by the assignee in issue 2761.

bmahabirbu:
This is needed since the OpenShell VM driver does not look for the mkfs package in its folder, downloaded or otherwise. It makes sense since mkfs is an OS-level package usually. Could be interesting if we have a fix upstream for this, but so far this fix is needed so the VM driver can work.

Fixes #2916.

Summary: On macOS and Linux, prepend the registered gateway binary's directory to the gateway process's PATH so the VM driver can discover the bundled mkfs.ext4 wrapper

### Manual test (macOS and Linux)

1. Run a Kaiden build containing this fix and start/restart a local gateway through Kaiden. Existing gateway wouldn't work
2. Copy/paste the block for your OS into a terminal on the host running Kaiden. It checks the most recently started Kaiden-managed gateway.

**macOS** (run on the Mac):

```sh
gateway_pid=$(pgrep -n -f '[/]openshell-gateway .*[/]openshell-gateways/')
gateway_path=$(ps eww -p "$gateway_pid" -o command= | perl -ne 'print "$1\n" if /(?:^| )PATH=(.*?)(?= [A-Za-z_][A-Za-z0-9_]*=|$)/')
env PATH="$gateway_path" /bin/sh -c 'command -v mkfs.ext4 || { echo "FAIL: mkfs.ext4 is missing from the gateway PATH"; exit 1; }'
```

**Linux** (run on the Linux host):

```sh
gateway_pid=$(pgrep -n -f '[/]openshell-gateway .*[/]openshell-gateways/')
gateway_path=$(tr '\0' '\n' < "/proc/$gateway_pid/environ" | sed -n 's/^PATH=//p')
env PATH="$gateway_path" /bin/sh -c 'command -v mkfs.ext4 || { echo "FAIL: mkfs.ext4 is missing from the gateway PATH"; exit 1; }'
```

**Expected:** a path ending in `/mkfs.ext4` in the same bundled directory as `openshell-gateway` (for example, `.../assets/darwin-arm64/mkfs.ext4` on macOS). A system mkfs path does not verify the bundle.


jeffmaury (approved):
LGTM

MarsKubeX (approved):
LGTM

Issue #2853, opened 2026-09-10
Remove the Chat view


fbricon:
### Task content

The Chat view is no longer useful, now that ACP sessions pretty much do the same thing, but in a sandbox.
We need to remove all the components, navigation, settings, tests, dependencies, documentation related to the Chat view

fbricon:
done

PR #2940, opened 2026-09-17
fix: redirect root and unknown routes to /agent-workspaces

fbricon:
After removing the Chat view, the `/` route had no matching component,
causing a blank page on startup. Stale `/chat/*` routes saved in
sessionStorage also led to blank pages.

Redirect `/` to a DEFAULT_PAGE (`/agent-workspaces`) in the router
subscriber and sanitize stale `/chat` saved routes on startup. The
default page is not persisted to sessionStorage since it is the
implicit landing page.

becomes part of #2853 


packages/renderer/src/Route.svelte
```diff
@@ -9,7 +9,7 @@ import { TelemetryService } from './TelemetryService';
 
 export let path = '/*';
 export let fallback = false;
-export let redirect = false;
+export let redirect: boolean | string = false;
```
benoitf: we can't change this file (it's coming from upstream)

fbricon: reverted

fbricon: faking a string as a boolean redirect now

fbricon: dropped the route redirect approach

benoitf (approved):

packages/renderer/src/App.svelte
```diff
@@ -142,9 +147,14 @@ router.subscribe(function (navigation) {
...
     sessionStorage.removeItem(SETTINGS_PAGE_KEY);
-    nonSettingsPage = '/';
+    nonSettingsPage = DEFAULT_PAGE;
+    router.goto(DEFAULT_PAGE);
   } else if (navigation.url.startsWith('/preferences')) {
     sessionStorage.setItem(SETTINGS_PAGE_KEY, navigation.url);
+  } else if (navigation.url === DEFAULT_PAGE) {
+    sessionStorage.removeItem(LAST_ROUTE_KEY);
+    sessionStorage.removeItem(SETTINGS_PAGE_KEY);
+    nonSettingsPage = DEFAULT_PAGE;
```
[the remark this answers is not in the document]
fbricon: this is a false positive issue:

Stale /chat route in sessionStorage: Already sanitized at module load time before the subscriber even runs:
if (savedRoute?.startsWith('/chat')) {
  savedRoute = undefined;
  sessionStorage.removeItem(LAST_ROUTE_KEY);
}

Valid saved route (e.g. /images): The restoration branch correctly restores it — that's the intended route persistence behavior.

No saved route: savedRoute is undefined, so the restoration branch is skipped, and url === '/' redirects to DEFAULT_PAGE.

User last visited DEFAULT_PAGE: We now clear LAST_ROUTE_KEY when navigating there (finding 2 fix), so there's nothing stale to restore on next load.

The race can't happen — the sanitization runs at import time, before subscribeReady is set.

fbricon:
@coderabbitai resolve

PR #2953, opened 2026-09-18
feat(vm): switched default to vm driver

### LINKED ISSUE #2761: Start gateway with the vm driver
### Task content

When Kaiden starts a gateway (either automaticaly or from user request), use the vm driver option and not the default computed one.

bmahabirbu:
Changed kaiden-local to use the vm driver and any created gateways as well. Kept support for podman docker too!

Fixes #2761.

### Manual testing

1. If an existing `kaiden-local` uses Podman, stop it and remove its gateway registration, then restart Kaiden.
2. Confirm `kaiden-local` is healthy and uses the VM driver.
3. Create a sandbox and confirm it reaches Ready and can execute commands. Check stop/start and deletion.
4. Open **Create local gateway** and confirm VM is selected by default.

Validated an isolated OpenShell 0.0.116 VM on macOS: sandbox creation, exec, stop/start, persisted file after `sync`, and deletion. This used a signed temporary binary and a cached image; packaged-app testing remains unverified.

**Linux verification is needed**


benoitf:
note: I have failure to start vms with OpenShell as the socket file using a path being "too long"

jeffmaury (approved):
LGTM

Tested on MacOS once binary is signed.

Maybe we should document the code signing command line in README.md or somewhere else

bmahabirbu:
@jeffmaury good idea i can add to the readme. Maybe as a seperate small issue and make sure binaries are signed for other cases

